PR #29, opened 2016-08-12
Feature/addict

asterite:
- Adds a User model: email, encrypted_password
- Adds [addict](https://github.com/trenpixster/addict)
- Login is required on index page, after login the email is shown in the welcome message


config/config.exs
```diff
@@ -25,3 +25,11 @@ config :logger, :console,
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{Mix.env}.exs"
+
+config :addict,
+  secret_key: "243262243132244543362f4573334a324257346976794e2e7559764675",
```
nekron: We should change this to a `System.get_env("SECRET_KEY")` call so we can override it in production.

Also consider moving it into the proper environment file (only moving the `from_email` and `secret_key`, the rest can stay here).

PR #153, opened 2016-10-19
Survey respondent steps refactor

juanboca:
Before this, the Survey Wizard retrieved all the respondents from the `respondents_controller`, only to show 5. This was limited on the UI. 
Now the `index` action from `respondents_controller` accepts a limit, and also returns the total count of respondents, to be able to display it on the UI.


asterite:
Can this 5 be hardcoded in the controller? What's the use case for choosing a limit in the action? A user could potentially hit this route with a big number and make the server slow.

juanboca:
The same action in the controller is used for the Respondents tab on the Survey detail, and in that place for now we need to show all the respondents. When we implement pagination for that tab, we can reuse the limit logic to retrieve only the respondents for the desired page.

Issue #145, opened 2016-10-17
Uploading CSV files with \r as newline doesn't work well


asterite:
Lines with `\r` are merged with the next line.

nditada:
Relevant: https://nicercode.github.io/blog/2013-04-30-excel-and-line-endings/

PR #163, opened 2016-10-20
(#145) Support "\r" as a line separator.

fgasperij:
The CSV library that was used to parse the file with the respondents did
not support "\r" as a line separator. The generation of the lines
from the file is performed by us now.


web/controllers/respondent_controller.ex
```diff
@@ -65,11 +86,13 @@ defmodule Ask.RespondentController do
     {:ok, local_time } = Ecto.DateTime.cast :calendar.local_time()
 
     if Path.extname(file.filename) == ".csv" do
-      entries = File.stream!(file.path) |>
-      CSV.decode(separator: ?\t) |>
-      Enum.map(fn row ->
-        %{phone_number: Enum.at(row, 0), survey_id: integer_survey_id, inserted_at: local_time, updated_at: local_time}
-      end)
+      csv_string = File.read!(file.path)
+      rows = csv_rows(csv_string)
```
juanboca: if the variable `rows` is not used anywhere else, I would replace it by the function call:
``` entries = 
   csv_rows(csv_string)
   |>...```
or maybe
```entries =
        File.read!(file.path)
        |> csv_rows
        |>...```

File: web/controllers/respondent_controller.ex [hunk 2]
```diff
@@ -56,6 +56,27 @@ defmodule Ask.RespondentController do
     render(conn, "stats.json", stats: stats)
   end
 
+  defp csv_rows(csv_string) do
+    delimiters = ["\r\n", "\r", "\n"]
```
juanboca: If the allowed delimiters are 3, it would be useful to have a test for each type of delimiter.

web/controllers/respondent_controller.ex
fgasperij: Yes, I prefer it that way too.

PR #164, opened 2016-10-20
Avoid multiple dispatches on add channel (fixes #124)

### LINKED ISSUE #124: Clicking multiple times on the "+" icon for creating a Channel brings up multiple popups
The button should be disabled until the channel creation dialog is closed

fgasperij:


web/static/js/actions/channels.js
```diff
@@ -26,31 +26,42 @@ export const receiveChannels = (response) => ({
...
+    if (creatingChannel) {
+      return Promise.resolve()
+    }
+    creatingChannel = true
+    return Promise.all([
+      authorizeWithGuisso(dispatch, 'nuntium', config.nuntium),
+      pigeon.loadPigeonScript(config.nuntium.baseUrl)
+    ])
+      .then(([token, _]) => {
+        window.something_beautiful = false
```
waj: `something_beautiful`? 🙂

web/static/js/actions/channels.js
```diff
@@ -26,31 +26,42 @@ export const receiveChannels = (response) => ({
...
-      dispatch(createChannel({
-        name: nuntiumChannel.name,
-        type: 'sms',
-        provider: 'nuntium',
-        settings: {
-          nuntiumChannel: nuntiumChannel.name
+export const createNuntiumChannel = (() => {
+  let creatingChannel = false
+  return () => dispatch => {
+    if (creatingChannel) {
```
waj: This has the problem that if the authorization popup is closed, there is no way to start over the process. I guess this check must be done right before `addChannel`.

Also, in `guisso.js` we could use a fixed name for the window (instead of `_blank`) to avoid multiple popups.

fgasperij: - If the check is performed right before `addChannel` multiple pop ups produced because the pop up is opened before that.
- Using a fixed name for the window would avoid opening multiple login windows but it wouldn't prevent the emergence of multiple Add channel pop ups with overlay.

PR #185, opened 2016-10-25
When switching a response to edit mode, put focus o…

### LINKED ISSUE #182: [Quiz] When switching a response to edit mode, put focus on the desired input
Now when the user adds a new choice, he has to click in 3 different components to be able to input text

juanboca:
…n the desired input

### DIFF
--- a/web/static/css/_ncd.scss
+++ b/web/static/css/_ncd.scss
@@ -175,6 +175,12 @@ footer {
         }
       }
     }
+    table.responses-table {
+      input[type=text] {
+        margin-bottom: 0;
+        height: auto;
+      }
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


juanboca:
Maybe the inline styles should be replaced by css classes. But in order for this component to work properly, it is needed that the sizes of the edit and the non-edit modes are the same. 
If the sizes are different, when the user clicks outside of this component (like on the ADD STEP link), the event `onMouseDown` is fired into the clicked element, but the `onMouseUp` is fired in another (let's say the div that contains the ADD STEP link), because the element moved between those two events. So, if there is an `onClick` handler for the clicked element, it is not fired.
Also, the movement of the components was a little confusing.

Issue #154, opened 2016-10-19
Data store refactor


waj:
**NOTE**: this is an aggregate task. Don't assign this issue to yourself.

Points collected during the data store design meeting:

0. All stores must have a boolean "fetching" property to identify data that is being loaded from the server.
This also means that the data must be moved to a sub-property of the store, using "data" for single objects and "items" for collections.
0. While fetching new data, don't blank the old data (so it can be displayed until new data comes)
0. Single items used for view/edit pages must have a separate store.
0. Each page must show the "loading" state if, and only if: `fetching == true && (data | items) == null`
0. Include a "filter" property to each store that describes the data contained or being loaded in the store

When a "fetch if needed" action is executed, the following logic applies:
* same `filter`, `fetching == true`: nothing (data is already coming)
* same `filter`, `fetching == false`: fetch the data from the server, set `fetching = true`
* different `filter`: same as previous plus: set new `filter`, blank old data and page information

When the data arrives from the server, if the `filter` is different (it has changed after the request was sent), the response is ignore. Otherwise, set the new `data` or `items`, and recalculate the current page.

Examples:

```js
{
  projects: {
    fetching: true | false,

    items: {
      [id]: {
      }
    },

    order: [1, 2, 3, 4, 5, 6],
    page: {
      index: 0,
      size: 3
    }
  },

  surveys: {
    fetching: false,
    filter: { projectId: 2 },
    items: {
      [id]: {

      }
    }
  }

  survey: {
    fetching: true | false,
    data: {
      id: 8,
    },
    validations: {

    }
  },

  respondentChart: {
    123: {

    }
  }

  respondentGroups: {
    fetching: true | false,
    items: {
      42: {
        id: 42,
        name: "foo.csv",
        respondentCount: 3150,
        primaryChannel: 44,
        fallbackChannel: 88,
        preview: ["123", "456", "789"]
      }
    },
    order: [42]
  }

  respondents: {
    filter: { surveyId: 8 },
    fetching: true | false,
    items: [
    ],
    sortBy: "",
    page: 8,
    pageSize: 20,
    totalCount: 800000
  }

  project: {
    fetching: true | false
    data: {
    }
  }
}
```

PR #191, opened 2016-10-26
Refactor/160 normalize questionnaire

### LINKED ISSUE #160: Apply data store refactor to questionnaire editor
See #154

nekron:
Fixes #160


web/static/js/actions/questionnaire.js
```diff
@@ -115,12 +115,11 @@ export const addStep = (stepType) => {
 
 export const save = () => {
   return (dispatch, getState) => {
-    const questionnaire = getState().questionnaire
-
-    if (questionnaire.id == null) {
-      api.createQuestionnaire(questionnaire.projectId, questionnaire)
-    } else {
+    const questionnaire = getState().questionnaire.data
+    if (questionnaire.id) {
```
mverzilli: Why this instead of `questionnaire.id == null`? I'm a bit weary of falsiness in JS

test/js/reducers/questionnaire.spec.js
```diff
@@ -285,7 +285,7 @@ describe('questionnaire reducer', () => {
   })
 })
 
-const questionnaire = {
+const questionnaire = () => ({
```
mverzilli: (Out of curiosity) Why this? The only reason I can think of is immutability, is that it?

web/static/js/components/questionnaires/QuestionnaireEditor.jsx
```diff
@@ -77,8 +71,8 @@ class QuestionnaireEditor extends Component {
     // I couldn't find a better way. Ideally this should be a sort of "callback"
     // to the addStep method, without involving additional component state handling
     // or explicit management via Redux reducers.
-    const questionnaireData = newProps.questionnaire.data
-    if (this.state.addingStep && questionnaireData != null && questionnaireData.steps != null && questionnaireData.steps.length > 0) {
+    const questionnaireData = newProps.questionnaire
+    if (this.state.addingStep && questionnaireData && questionnaireData.steps != null && questionnaireData.steps.length > 0) {
```
mverzilli: Ditto falsiness

test/js/reducers/questionnaire.spec.js
nekron: Yup. Eventually we solved that by using deepFreeze.

web/static/js/components/questionnaires/QuestionnaireEditor.jsx
nekron: I'll make a regression over this once it's merged and see if there is a breaking case.

web/static/js/actions/questionnaire.js
```diff
@@ -0,0 +1,125 @@
...
+export const ADD_STEP = 'QUESTIONNAIRE_ADD_STEP'
+export const DELETE_STEP = 'QUESTIONNAIRE_DELETE_STEP'
+export const CHANGE_STEP_TITLE = 'QUESTIONNAIRE_CHANGE_STEP_TITLE'
+export const CHANGE_STEP_PROMPT_SMS = 'QUESTIONNAIRE_CHANGE_STEP_PROMPT_SMS'
+export const CHANGE_STEP_STORE = 'QUESTIONNAIRE_CHANGE_STEP_STORE'
+export const ADD_CHOICE = 'QUESTIONNAIRE_EDITOR_ADD_CHOICE'
+export const DELETE_CHOICE = 'QUESTIONNAIRE_EDITOR_DELETE_CHOICE'
+export const CHANGE_CHOICE = 'QUESTIONNAIRE_EDITOR_CHANGE_CHOICE'
+
+export const fetchQuestionnaire = (projectId, questionnaireId) => (dispatch, getState) => {
```
asterite: I don't understand why there is this and `if needed` variants. The logic is covered by the `fetching` and `filter` properties, and #154 doesn't mention two ways of fetching a questionnaire. Why is this needed?

web/static/js/components/questionnaires/QuestionnaireEditor.jsx
```diff
@@ -39,7 +39,45 @@ class QuestionnaireEditor extends Component {
...
+
+  internalState(currentStep, addingStep = false) {
+    return {
+      currentStep,
+      addingStep
+    }
+  }
+
+  componentWillReceiveProps(newProps) {
+    // This feels a bit hacky, but it let's us expand the step we just created.
```
asterite: If the stat stores what step is expanded (like it was before) it's super simple to implement and there's no need for a hack. Why not implement it that way?

web/static/js/actions/questionnaire.js
nekron: There is no need to export this function. It's just to be able to force it to fetch if necessary.

web/static/js/components/surveys/SurveyEdit.jsx
```diff
@@ -29,9 +29,8 @@ class SurveyEdit extends Component {
     const { dispatch, projectId, surveyId, router } = this.props
     if (projectId && surveyId) {
       projectActions.fetchProject(projectId)
-      dispatch(actions.fetch(projectId, surveyId))
+      dispatch(actions.fetchSurveyIfNeeded(projectId, surveyId))
```
nekron: The first is just there to simplify the reading of the second.

It's being exported just in case that someone need to force the fetch for some reason. Maybe we can keep it private for now until there is a good reason for that.

docker-compose.yml
```diff
@@ -34,6 +34,7 @@ services:
   brunch:
     image: node:6.3.1
     command: node_modules/brunch/bin/brunch watch
+    restart: always
```
nekron: True.

It was useful as we developed, and there were already some conflicting changes in master that would've led to a hard to test scenario. That prevented us from making this change in master and then merging it.

I should've kept it stashed.

web/static/js/actions/questionnaire.js
asterite: Ah, I see. So maybe the public API should be `fetchQuestionnaire` and the private one be `forceFetchQuestionnaire`?

web/static/js/components/questionnaires/QuestionnaireEditor.jsx
nekron: We should talk this with @mverzilli

mverzilli: It felt wrong because we'd have to add ad-hoc questionnaire substate to the store for the editor. This way we keep the reducer pretty focused on data management, and the editor manages UI state. 

But maybe it's not worth the effort, since as you said the previous approach seemed simpler. 

The questionnaire editor will suffer a lot of changes in the short term, so maybe we can move this to it's own issue.

PR #200, opened 2016-10-28
Fix/112 restrict characters in phone numbers

### LINKED ISSUE #112: The respondent list upload accepts letters on the phone numbers.
The upload of respondents doesn't validate that the telephone numbers have only numbers.

lmatayoshi:
Fixes #112


web/controllers/respondent_controller.ex
```diff
@@ -110,41 +110,66 @@ defmodule Ask.RespondentController do
...
+    {:ok, local_time } = Ecto.DateTime.cast :calendar.local_time()
+    {integer_survey_id, _ } = Integer.parse survey_id
+
+    entries = rows
+      |> Enum.map(fn row ->
+        %{phone_number: row, survey_id: integer_survey_id, inserted_at: local_time, updated_at: local_time}
+      end)
+
+    respondents_count = entries
+      |> Enum.chunk(1_000, 1_000, []) |> Enum.reduce(0, fn(chunked_entries, total_count)  ->
```
fgasperij: Chains with more than one pipe have one pipe per line and aren't further indented.

web/controllers/respondent_controller.ex
```diff
@@ -110,41 +110,66 @@ defmodule Ask.RespondentController do
...
+      invalid_entries = Enum.filter(indexed_rows, fn entry -> !Regex.match?(~r/^([0-9]|\(|\)|\+|\-| )+$/, entry.phone_number) end)
 
-      conn
-        |> put_status(:created)
-        |> render("index.json", respondents: respondents |> Repo.preload(:responses), respondents_count: respondents_count)
+      case invalid_entries do
+        [] -> render_respondents(conn, survey_id, rows)
+        _ -> render_invalid(conn, file.filename, invalid_entries)
+      end
     else
```
fgasperij: Since `indexed_rows` is only used to perform the `Enum.filter/2`, it could be directly added to the previous chain of pipes.

Issue #316, opened 2016-11-17
Enable multilingual questionnaires


waj:
Current prompts and responses must be considered as English (en). Write migration to reformat the questionnaires JSON so the prompts are enclosed in another level specifying the language:

```json
{
  "id": "9616feb6-33c0-4feb-8aa8-84ba9a607103",
  "type": "multiple-choice",
  "prompt": {
    "en": {
      "sms": "Do you smoke? Reply YES or NO",
      "ivr": "Do you smoke? Press 1 for YES, 2 for NO"
    }
  },
  "store": "Smokes",
  "choices": [
    {
      "value": "Yes",
      "responses": {
        "en": {
          "sms": ["yes", "y"],
          "ivr": ["1"]
        }
      }
    }
  ]
}
```

Modify also all the occurrences in the editors and runtime to find the elements under the "en" language.

PR #343, opened 2016-11-23
JSON Schema for Questionnaire steps

mverzilli:
WIP Related to #316 (Enable multilingual questionnaires)

Next steps: use this infrastructure to validate against questionnaire tables to be migrated. Then evolve the schema to reflect the JSON structures introduced to support multilanguage. Finally, consider moving validation from questionnaire_controller to questionnaire model, so it can be taken advantage of from everywhere.

### DIFF
--- a/lib/ask/json_schema.ex
+++ b/lib/ask/json_schema.ex
@@ -10,8 +10,6 @@ defmodule Ask.JsonSchema do
 
   def start_link do
     GenServer.start_link(__MODULE__, [], name: @server_ref)
-    IO.puts "STARTING JSON SCHEMA WORKER"
-    IO.inspect @server_ref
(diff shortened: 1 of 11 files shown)


lib/ask/json_schema.ex
```diff
@@ -0,0 +1,69 @@
...
+  """
+  use GenServer
+
+  @server_ref {:global, __MODULE__}
+
+  def server_ref, do: @server_ref
+
+  def start_link do
+    GenServer.start_link(__MODULE__, [], name: @server_ref)
+    IO.puts "STARTING JSON SCHEMA WORKER"
```
waj: Please remove these `IO.puts`

Issue #322, opened 2016-11-18
[Quiz] Skip logic for numeric steps


waj:
https://projects.invisionapp.com/d/main#/console/8374235/193624465/preview

Store the configuration like this:
```json
{
  "id": "f7f13e7f-3aca-41c8-8d5f-238277a7e6cb",
  "type": "numeric",
  "prompt": {
    "en": {
      "sms": "How old are you?"
    }
  },
  "store": "Age",
  "min_value": 0,
  "max_value": null,
  "range_delimiters": "18,65",
  "ranges": [
    {
      "from": 0,
      "to": 17,
      "skip_logic": "end"
    },
    {
      "from": 18,
      "to": 64,
      "skip_logic": null
    },
    {
      "from": 65,
      "to": null,
      "skip_logic": null
    }
  ]
}
```

Change also the runtime Flow to validate responses against the configured values.

PR #363, opened 2016-12-01
Feature/322 skip logic numeric steps

fgasperij:
Connects to #322

### DIFF
--- a/lib/ask/runtime/flow.ex
+++ b/lib/ask/runtime/flow.ex
@@ -45,8 +45,8 @@ defmodule Ask.Runtime.Flow do
           value = String.to_integer(reply_value)
           step["ranges"]
           |> Enum.find_value(nil, fn (range) ->
-            if (range["from"] == nil || range["from"] <= value) && (range["to"] >=
-              value || range["to"] == nil), do: range["skip_logic"], else: false
+            if (range["from"] == nil || range["from"] <= value) && (range["to"]
+              == nil || range["to"] >= value), do: range["skip_logic"], else: false
           end)
         "multiple-choice" ->
           step
--- a/lib/ask/runtime/flow.ex
+++ b/lib/ask/runtime/flow.ex
@@ -40,10 +40,20 @@ defmodule Ask.Runtime.Flow do
 
   defp next_step_by_skip_logic(flow, step, reply_value) do
     skip_logic =
-      step
-      |> Map.get("choices")
-      |> Enum.find(fn choice -> choice["value"] == reply_value end)
-      |> Map.get("skip_logic")
+      case step["type"] do
+        "numeric" ->
+          value = String.to_integer(reply_value)
+          step["ranges"]
+          |> Enum.find_value(nil, fn (range) ->
+            if (range["from"] == nil || range["from"] <= value) && (range["to"] >=
+              value || range["to"] == nil), do: range["skip_logic"], else: false
+          end)
+        "multiple-choice" ->
+          step
+          |> Map.get("choices")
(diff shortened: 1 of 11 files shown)


lib/ask/runtime/flow.ex
```diff
@@ -40,10 +40,20 @@ defmodule Ask.Runtime.Flow do
...
-      |> Map.get("choices")
-      |> Enum.find(fn choice -> choice["value"] == reply_value end)
-      |> Map.get("skip_logic")
+      case step["type"] do
+        "numeric" ->
+          value = String.to_integer(reply_value)
+          step["ranges"]
+          |> Enum.find_value(nil, fn (range) ->
+            if (range["from"] == nil || range["from"] <= value) && (range["to"] >=
+              value || range["to"] == nil), do: range["skip_logic"], else: false
```
asterite: The condition for `range["to"]` should be inverted: `range["to"] == nil || range["to"] >= value`

lib/ask/runtime/flow.ex
```diff
@@ -40,10 +40,20 @@ defmodule Ask.Runtime.Flow do
...
   defp next_step_by_skip_logic(flow, step, reply_value) do
     skip_logic =
-      step
-      |> Map.get("choices")
-      |> Enum.find(fn choice -> choice["value"] == reply_value end)
-      |> Map.get("skip_logic")
+      case step["type"] do
+        "numeric" ->
+          value = String.to_integer(reply_value)
+          step["ranges"]
```
asterite: Is there a migration that adds this `ranges` property to existing numeric steps? If not, the runtime for existing steps will fail.

lib/ask/runtime/flow.ex
```diff
@@ -40,10 +40,20 @@ defmodule Ask.Runtime.Flow do
...
+          |> Enum.find_value(nil, fn (range) ->
+            if (range["from"] == nil || range["from"] <= value) && (range["to"] >=
+              value || range["to"] == nil), do: range["skip_logic"], else: false
+          end)
+        "multiple-choice" ->
+          step
+          |> Map.get("choices")
+          |> Enum.find(fn choice -> choice["value"] == reply_value end)
```
asterite: What happens if there's no choice with that value? I guess it returns `nil` and the next line fails?

lib/ask/runtime/flow.ex
fgasperij: There is a [test](https://github.com/instedd/ask/pull/363/files#diff-d5b2e2f1b6436a3d4a30d58988617b2bR126) for it. The problem is that comparing `nil` with an integer doesn't raise an exception, it compares bigger to all of them. I'll re order the formula clauses just for clarity since the check for `nil` on the `range["to"]` could be deleted all together due to this strange behavior.

lib/ask/runtime/flow.ex
fgasperij: This will never happen because [here](https://github.com/instedd/ask/blob/feature/322-skip-logic-numeric-steps/lib/ask/runtime/flow.ex#L106) we assign `nil` to the `reply_value` if there isn't any choice with the response of the respondent and then [here](https://github.com/instedd/ask/blob/feature/322-skip-logic-numeric-steps/lib/ask/runtime/flow.ex#L79) we jump to the next question and never call `next_step_by_skip_logic/3` if the reply value is `nil`.

lib/ask/runtime/flow.ex
fgasperij: I'll added, thanks!

PR #368, opened 2016-12-03
Download quiz contents for translation

mverzilli:
WIP: DO NOT MERGE THIS. We need a very thorough code review of this before proceeding.

### DIFF
--- a/test/js/reducers/questionnaire.spec.js
+++ b/test/js/reducers/questionnaire.spec.js
@@ -4,7 +4,7 @@ import assert from 'assert'
 import { playActionsFromState } from '../spec_helper'
 import find from 'lodash/find'
 import deepFreeze from '../../../web/static/vendor/js/deepFreeze'
-import reducer, { stepStoreValues } from '../../../web/static/js/reducers/questionnaire'
+import reducer, { stepStoreValues, csvForTranslation } from '../../../web/static/js/reducers/questionnaire'
 import * as actions from '../../../web/static/js/actions/questionnaire'
 
 describe('questionnaire reducer', () => {
@@ -349,21 +349,17 @@ describe('questionnaire reducer', () => {
 
     const step = find(resultState.data.steps, s => s.id === '17141bea-a81c-4227-bdda-f5f69188b0e7')
     expect(step.choices.length).toEqual(2)
-    expect(step.choices[1]).toEqual({
-      value: 'Maybe',
-      responses: {
-        'en': {
-          sms: [
-            'M',
-            'MB',
-            '3'
-          ],
-          ivr: [
-            'May'
-          ]
-        }
-      },
-      skipLogic: 'end'
+    expect(step.choices[1].value).toEqual('Maybe')
+    expect(step.choices[1].skipLogic).toEqual('end')
+    expect(step.choices[1].responses['en']).toEqual({
+      sms: [
+        'M',
+        'MB',
+        '3'
+      ],
+      ivr: [
+        'May'
+      ]
     })
   })
 
@@ -645,7 +641,7 @@ describe('questionnaire reducer', () => {
 
     const languageSelection = resultState.data.steps[0]
     expect(languageSelection.type).toEqual('language-selection')
-    expect(languageSelection.choices).toInclude('fr')
+    expect(languageSelection.languageChoices).toInclude('fr')
   })
 
   it('should add a new language last inside the choices of the language selection step', () => {
@@ -660,7 +656,7 @@ describe('questionnaire reducer', () => {
     ])
 
     const languageSelection = resultState.data.steps[0]
-    expect(languageSelection.choices[languageSelection.choices.length - 1]).toEqual('de')
+    expect(languageSelection.languageChoices[languageSelection.languageChoices.length - 1]).toEqual('de')
   })
 
   it('should remove a language inside the choices of the language selection step', () => {
@@ -673,15 +669,15 @@ describe('questionnaire reducer', () => {
       actions.addLanguage('fr')
     ])
     const preLanguageSelection = preState.data.steps[0]
-    expect(preLanguageSelection.choices[2]).toEqual('de')
+    expect(preLanguageSelection.languageChoices[2]).toEqual('de')
 
     const resultState = playActionsFromState(preState, reducer)([
       actions.removeLanguage('de')
     ])
 
     const languageSelection = resultState.data.steps[0]
-    expect(languageSelection.choices[2]).toEqual('es')
-    expect(languageSelection.choices[3]).toEqual('fr')
+    expect(languageSelection.languageChoices[2]).toEqual('es')
+    expect(languageSelection.languageChoices[3]).toEqual('fr')
   })
 
   it('should reorder correctly the languages inside the choices of the language selection step', () => {
@@ -699,10 +695,10 @@ describe('questionnaire reducer', () => {
     ])
 
     const languageSelection = resultState.data.steps[0]
-    expect(languageSelection.choices[1]).toEqual('es')
-    expect(languageSelection.choices[2]).toEqual('de')
-    expect(languageSelection.choices[3]).toEqual('fr')
-    expect(languageSelection.choices[4]).toEqual('en')
+    expect(languageSelection.languageChoices[1]).toEqual('es')
+    expect(languageSelection.languageChoices[2]).toEqual('de')
+    expect(languageSelection.languageChoices[3]).toEqual('fr')
+    expect(languageSelection.languageChoices[4]).toEqual('en')
   })
 
   it('should reorder correctly the languages inside the choices of the language selection step 2', () => {
@@ -720,10 +716,10 @@ describe('questionnaire reducer', () => {
     ])
 
     const languageSelection = resultState.data.steps[0]
-    expect(languageSelection.choices[1]).toEqual('fr')
-    expect(languageSelection.choices[2]).toEqual('en')
-    expect(languageSelection.choices[3]).toEqual('es')
-    expect(languageSelection.choices[4]).toEqual('de')
+    expect(languageSelection.languageChoices[1]).toEqual('fr')
+    expect(languageSelection.languageChoices[2]).toEqual('en')
+    expect(languageSelection.languageChoices[3]).toEqual('es')
+    expect(languageSelection.languageChoices[4]).toEqual('de')
   })
 
   it('should add language', () => {
@@ -803,6 +799,7 @@ describe('questionnaire reducer', () => {
     ])
 
     const resultState = playActionsFromState(preState, reducer)([
+      actions.addLanguage('fr'),
       actions.setDefaultLanguage('fr')
     ])
 
@@ -835,6 +832,32 @@ describe('questionnaire reducer', () => {
       Exercises: ['Yes', 'No']
     })
   })
+
+  describe('csv for translation', () => {
+    it('should work', () => {
+      const state = playActions([
+        actions.fetch(1, 1),
+        actions.receive(questionnaire),
+        actions.addLanguage('fr'),
+        actions.addLanguage('es')
+      ])
+
+      const csv = csvForTranslation(state.data)
+
+      const expected = [
+        ['"en"', '"fr"', '"es"'],
+        ['"Do you smoke?"', '""', '"Fumas?"'],
+        ['"Yes, Y, 1"', '""', '"Sí, S, 1"'],
+        ['"No, N, 2"', '""', '"No, N, 2"'],
+        ['"Do you exercise?"', '""', '"Ejercitas?"']
+      ]
+
+      console.log(csv)
+
+      expect(csv.length).toEqual(expected.length)
+      expected.forEach((row, index) => expect(csv[index]).toEqual(row))
+    })
+  })
 })
 
 const questionnaire = deepFreeze({
@@ -857,6 +880,16 @@ const questionnaire = deepFreeze({
               ivr: [
                 '1'
               ]
+            },
+            'es': {
+              sms: [
+                'Sí',
+                'S',
+                '1'
+              ],
+              ivr: [
+                '1'
+              ]
             }
           },
           skipLogic: null
@@ -873,6 +906,16 @@ const questionnaire = deepFreeze({
               ivr: [
                 '2'
               ]
+            },
+            'es': {
+              sms: [
+                'No',
+                'N',
+                '2'
+              ],
+              ivr: [
+                '2'
+              ]
             }
           },
           skipLogic: 'b6588daa-cd81-40b1-8cac-ff2e72a15c15'
@@ -881,6 +924,9 @@ const questionnaire = deepFreeze({
       prompt: {
         'en': {
           sms: 'Do you smoke?'
+        },
+        'es': {
+          sms: 'Fumas?'
         }
       }
     },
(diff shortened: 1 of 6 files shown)


test/js/reducers/questionnaire.spec.js
```diff
@@ -350,21 +350,17 @@ describe('questionnaire reducer', () => {
...
-            'MB',
-            '3'
-          ],
-          ivr: [
-            'May'
-          ]
-        }
-      },
-      skipLogic: 'end'
+    expect(step.choices[1].value).toEqual('Maybe')
```
mverzilli: I changed this so that changes in the dummy questionnaire at the bottom of this file don't affect it so easily. An example of a change that would break this test is adding language prompts to the dummy object.

test/js/reducers/questionnaire.spec.js
```diff
@@ -646,7 +642,7 @@ describe('questionnaire reducer', () => {
 
     const languageSelection = resultState.data.steps[0]
     expect(languageSelection.type).toEqual('language-selection')
-    expect(languageSelection.choices).toInclude('fr')
+    expect(languageSelection.languageChoices).toInclude('fr')
```
mverzilli: I was forced to rename `choices` to `languageChoices` for language selection steps because of Flow. I wanted Step to be union of MultipleChoiceStep and LanguageSelectionStep and that (AFAICT) will only let you tag the union based on values (eg: type === 'language-selection' vs. type === 'multiple-choice'), having two types in the union with a property with the same name and different type apparently messes up the type system.

Multiple Choice Steps have complex `choices`, with values that change depending on mode and language, whereas Language Selection Steps have string `choices`. I don't know if that was on purpose, but in order to merge everything in a sane way I had to properly type choices in both cases, which ultimately led me to rename `choices` in Language Selection Steps to `languageChoices`.

test/js/reducers/questionnaire.spec.js
```diff
@@ -784,6 +780,7 @@ describe('questionnaire reducer', () => {
     ])
 
     const resultState = playActionsFromState(preState, reducer)([
+      actions.addLanguage('fr'),
```
mverzilli: Notice that `setDefaultLanguage('fr)` does not result in adding to the list of languages AFAICT. There are a number of invariants broken each time `setDefaultLanguage` and `addLanguage` are dispatched (I'll comment below).

test/js/reducers/questionnaire.spec.js
```diff
@@ -816,6 +813,32 @@ describe('questionnaire reducer', () => {
       Exercises: ['Yes', 'No']
     })
   })
+
+  describe('csv for translation', () => {
```
mverzilli: This test covers almost all that is needed to check the CSV Download works. The only piece missing is to add rows for the language selection prompts to the `expected` array. 

Currently this test fails. That is because the `state.data` fed to the `csvForTranslation` function is missing prompts in Spanish, which are important to validate the CSV export procedure (otherwise you'd rows with only English values in the first column).

I had originally added prompts to the dummy questionnaire at the bottom of this file, but that broke all the tests related to language management once I tried to rebase from `master`.

Then I rollbacked the changes to the dummy questionnaire and tried to get to the same state by issuing `addLanguage` actions and corresponding "prompt change" actions. 

When I set out to do this I discovered that `addLanguage` and `setDefaultLanguage` only affect the questionnaire's `languages` and `defaultLanguage` properties while they should also fill the model with properly initialized data for modes, prompts and languages. 

That caused the validation code to break in some places, because it's expecting that when the questionnaire has a given mode enabled, the structures underneath prompts and responses are sound.

I guess all of these wasn't uncovered until now because all the validation code is pre-multilingual support and assumes that everything is under a hardcoded 'en' key.

web/static/js/components/questionnaires/QuestionnaireEditor.jsx
```diff
@@ -87,6 +88,21 @@ class QuestionnaireEditor extends Component {
     }
   }
 
+  downloadCsv(e) {
```
mverzilli: This is done. This code runs when the user clicks the Download link, applies the `csvForTranslation` function on the current questionnaire and writes the content to an URI encoded binary that the downloads. I intended to refactor this before wrapping up the issue, although the code is not terrible.

web/static/js/reducers/questionnaire.js
```diff
@@ -20,7 +21,18 @@ const dataReducer = (state: Questionnaire, action): Questionnaire => {
 }
 
 const steps = (state, action) => {
-  const newSteps = state.steps == null ? [] : stepsReducer(state.steps, action)
+  // Up to now we've been assuming that all content was under corresponding 'en' keys,
```
mverzilli: Please read this comment carefully.

web/static/js/reducers/questionnaire.js
```diff
@@ -79,11 +91,14 @@ const deleteChoice = (state, action) => {
   }))
 }
 
-const changeChoice = (state, action) => {
+// TODO: now we're assuming that the choice content being changed is
+// in the default language. Revisit this if we ever allow users to
+// directly edit content in all languages directly from the UI.
+const changeChoice = (state, action, quiz: Questionnaire) => {
```
mverzilli: Here: all the references to `defaultLanguage` in the reducer are assuming that we only allow users to edit the current defaultLanguage content. I'm almost sure that this is wrong but refactoring this was way out of reach for the amount of time I had.

web/static/js/reducers/questionnaire.js
```diff
@@ -108,19 +123,19 @@ const changeChoice = (state, action) => {
...
-const autoComplete = (state, value) => {
+const autoComplete = (state, value, quiz: Questionnaire) => {
   let setted = false
 
   let smsValues = ''
   let ivrValues = ''
 
   state.forEach((step) => {
-    if (!setted) {
+    if ((step.type === 'multiple-choice') && !setted) {
```
mverzilli: Watch this: again, for timing issues I limited autocompletes to regular Multiple Choice steps. I'm almost sure this should also be made to work with Language Selection steps but I had to cut corners.

web/static/js/reducers/questionnaire.js
```diff
@@ -108,19 +123,19 @@ const changeChoice = (state, action) => {
...
       step.choices.forEach((choice) => {
         if (choice.value == value && !setted) {
           setted = true
-          smsValues = choice.responses['en'].sms.join(',')
-          ivrValues = choice.responses['en'].ivr.join(',')
+          smsValues = choice.responses[quiz.defaultLanguage].sms.join(',')
```
mverzilli: This should probably be changed to check all prompts and responses in all languages in all modes.

web/static/js/reducers/questionnaire.js
```diff
@@ -136,8 +151,9 @@ const deleteStep = (state, action) => {
   return filter(state, s => s.id != action.stepId)
 }
 
-const changeStep = (state, stepId, func: (step: Step) => Step) => {
+function changeStep<T: Step>(state, stepId, func: (step: Object) => T) {
```
mverzilli: These type annotations make Flow happy with a polymorphic `func` parameter. I suggest you not to mess with it :P

web/static/js/reducers/questionnaire.js
```diff
@@ -150,28 +166,28 @@ type ActionChangeStepSmsPrompt = {
...
 
-const changeStepSmsPrompt = (state, action: ActionChangeStepSmsPrompt) => {
+const changeStepSmsPrompt = (state, action: ActionChangeStepSmsPrompt, quiz: Questionnaire): Step[] => {
   return changeStep(state, action.stepId, step => ({
     ...step,
     prompt: {
       ...step.prompt,
-      'en': {
-        ...step.prompt['en'],
+      [quiz.defaultLanguage]: {
```
mverzilli: Another place to check: here I changed the assumption of editing English to the slightly less static assumption that we're editing the default language. Neither is actually right. I guess we should augment all relevant actions with a "lang" prop to use here.

web/static/js/reducers/questionnaire.js
```diff
@@ -150,28 +166,28 @@ type ActionChangeStepSmsPrompt = {
...
 
-const changeStepIvrPrompt = (state, action) => {
+const changeStepIvrPrompt = (state, action, quiz: Questionnaire) => {
   return changeStep(state, action.stepId, step => ({
     ...step,
     prompt: {
       ...step.prompt,
-      'en': {
-        ...step.prompt['en'],
+      [quiz.defaultLanguage]: {
```
mverzilli: Again, this is just a patch. The right solution would be to remove the `quiz` param and add a `lang` prop to the action that can then be used to address the right content.

web/static/js/reducers/questionnaire.js
```diff
@@ -222,26 +238,47 @@ const changeStepStore = (state, action) => {
...
-  type: 'multiple-choice',
-  title: '',
-  store: '',
-  prompt: {
-    'en': {
-      sms: '',
-      ivr: {
-        text: '',
-        audioSource: 'tts'
+const newLanguageSelectionStep = (first: string, second: string): LanguageSelectionStep => {
```
mverzilli: I had to remove `newStep` and introduce some code duplication to these factory methods for each kind of step, because I couldn't find a DRY way that would please Flow. For some (I guess good) reason Flow can't infer what happens when you use `Object.assign` so any attempt of merging "common step properties" with specific ones utterly failed. I suggest not to touch this for the time being. It works ok, Flow likes it and it's easy to read and hard to mess up. There are much more urgent issues to address :).

web/static/js/reducers/questionnaire.js
```diff
@@ -222,26 +238,47 @@ const changeStepStore = (state, action) => {
...
+          text: '',
+          audioSource: 'tts'
+        }
       }
-    }
-  },
-  choices: []
-})
+    },
+    languageChoices: [null, first, second]
```
mverzilli: I kept this from the original implementation, but I still don't understand why we create this with a `null` first element. I guess it is because we want to ensure that the default language is first there and there's some code to set it up further on. My suggestion here would be instead of doing that pass the default language as a param to this factory method. That way we can specify to Flow that `languageChoices` is an array of strings instead of an array of nullable strings, which will make a lot of subsequent code a lot less awkward.

web/static/js/reducers/questionnaire.js
```diff
@@ -347,16 +396,10 @@ const addLanguageSelectionStep = (state, action) => {
...
-  var step = newStep()
-  step.type = 'language-selection'
-  step.choices = [null, first, second]
-  step.title = 'Language selection'
-
-  return step
-}
-
 const setDefaultLanguage = (state, action) => {
+  console.log(state.steps)
```
mverzilli: Ooops. Feel free to remove these :)

web/static/js/reducers/questionnaire.js
```diff
@@ -395,20 +440,27 @@ const validateSteps = (path, steps, context) => {
 }
 
 const validateStep = (path, step, context) => {
-  if (context.sms && isBlank(step.prompt['en'].sms)) {
+  if (context.sms &&
```
mverzilli: I had to make all this guards super defensive because of the broken invariants resulting from the way `addLanguage` and `setDefaultLanguage` are implemented. The way to get away from this IMO is to provide more robust implementations to those actions so that these guards can be relaxed.

web/static/js/reducers/questionnaire.js
```diff
@@ -395,20 +440,27 @@ const validateSteps = (path, steps, context) => {
...
   }
 
-  if (step.type == 'multiple-choice') {
+  if (step.type === 'multiple-choice') {
     validateChoices(`${path}.choices`, step.choices, context)
   }
 }
 
-const validateChoices = (path, choices, context) => {
+const validateChoices = (path, choices: Choice[], context) => {
```
mverzilli: This is assuming the `choices` structure of a regular Multiple Choice step. That's why I guard it from being called with Language Selection steps .

web/static/js/reducers/questionnaire.js
```diff
@@ -425,37 +477,51 @@ const validateChoices = (path, choices, context) => {
     if (values.includes(choice.value)) {
       addError(context, `${path}[${i}].value`, 'Value already used in a previous response')
     }
-    for (let choiceSms of choice.responses['en'].sms) {
-      if (sms.includes(choiceSms)) {
-        addError(context, `${path}[${i}].sms`, `Value "${choiceSms}" already used in a previous response`)
+
+    if (choice.responses[context.defaultLanguage] && choice.responses[context.defaultLanguage].sms) {
```
mverzilli: Defensive. Should be simpler with stronger invariants.

web/static/js/reducers/questionnaire.js
```diff
@@ -425,37 +477,51 @@ const validateChoices = (path, choices, context) => {
...
     }
+
     values.push(choice.value)
-    sms.push(...choice.responses['en'].sms)
-    ivr.push(...choice.responses['en'].ivr)
   }
 }
 
-const validateChoice = (path, choice, context) => {
+const validateChoice = (path, choice: Choice, context) => {
```
mverzilli: See my comment on `validateChoices`.

web/static/js/reducers/questionnaire.js
```diff
@@ -483,4 +549,70 @@ export const stepStoreValues = (questionnaire) => {
   }, {})
 }
 
+export const csvForTranslation = (questionnaire: Questionnaire) => {
```
mverzilli: This code is nasty and full of duplicate pieces. I was supposed to refactor this after rebasing from `master`, but that turned out to be easier said than done :). 

It mostly works, except for language selection steps. 

The important things to understand are these:

-We ensure that the CSV headers start with the default lang, that's what the first few lines is about.
-We need to enclose all cells in `""` so that we can support `,` inside each column in Excel.
-We ignore language selection steps for now, out of ignorance, not intent :).
-We keep a dictionary of `exported` terms. It is just a hash of `string => bool`. The keys are always in the default lang and the values specify whether we already exported that term. That is to avoid duplicates in the output (So "Yes, Y, 1" will appear only ones although it's likely to be present in many steps).
-We intentionally skip IVR responses, because they're digits and we need them to match in all languages.

web/static/js/reducers/questionnaire.js
```diff
@@ -256,26 +272,47 @@ const changeStepStore = (state, action) => {
...
+    },
+    languageChoices: [null, first, second]
+  }
+}
+
+const newMultipleChoiceStep = () => {
+  return {
+    id: uuid.v4(),
+    type: 'multiple-choice',
+    title: 'Language selection',
```
asterite: This seems to be a mistake...

asterite:
This PR should be ready for merge now.

However, I think this feature just by itself adds very little value, because:

* You can download a CSV that will contains texts for the default language, and an empty column for each extra language (because you can't edit those in the UI)
* You can't upload a CSV with translations
* You can't define translations in the UI

So the CSV will always have all English texts and be useless, until we either allow uploading CSVs **and** showing/editing translations in the UI (the second point is needed because otherwise you upload a CSV and can't see if it worked, or what effect it had)

mverzilli:
Yeah, next step is upload. The UI features are fundamental, of course, but just with upload at least you can prepare a demo or test questionnaires with multiple languages.

asterite:
@mverzilli Note that yesterday I just wrote that as a reminder to myself (and others) to maybe implement those 2 steps so it can be shown more easily in the demo, not that I consider this PR of little value (in fact with just a few changes I had the tests and the feature working)

PR #424, opened 2016-12-12
Fixes #383. Improve name of CSV translations

### LINKED ISSUE #383: [Questionnaire] Improve the name of the CSV of translations.
Now the name of the file is `download.csv`.
In order to provide more information it would be useful to replace that name with something like `questionnaire_1_translations.csv`.

lmatayoshi:
Fixes #383

### DIFF
--- a/test/js/reducers/questionnaire.spec.js
+++ b/test/js/reducers/questionnaire.spec.js
@@ -818,26 +818,129 @@ describe('questionnaire reducer', () => {
 
   describe('helpers', () => {
     it('should provide valid answers for multiple-choice steps', () => {
-      const questionnaire: Questionnaire = deepFreeze({
+      const bareQuestionnaire: Questionnaire = {
+        name: 'q1',
+        modes: ['sms'],
+        languages: [],
+        defaultLanguage: 'en',
+        quotaCompletedMsg: null,
         steps: [
           {
             type: 'multiple-choice',
+            title: 'Do you smoke?',
             store: 'Smokes',
-            choices: [{value: 'Yes'}, {value: 'No'}]
+            id: '17141bea-a81c-4227-bdda-f5f69188b0e7',
+            choices: [
+              {
+                value: 'Yes',
+                responses: {
+                  'en': {
+                    sms: [
+                      'Yes'
+                    ]
+                  }
+                },
+                skipLogic: null
+              },
+              {
+                value: 'No',
+                responses: {
+                  'en': {
+                    sms: [
+                      'No',
+                      'N',
+                      '2'
+                    ]
+                  }
+                },
+                skipLogic: 'b6588daa-cd81-40b1-8cac-ff2e72a15c15'
+              }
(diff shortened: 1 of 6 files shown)


test/js/reducers/questionnaire.spec.js
```diff
@@ -817,7 +818,7 @@ describe('questionnaire reducer', () => {
 
   describe('helpers', () => {
     it('should provide valid answers for multiple-choice steps', () => {
-      const questionnaire = deepFreeze({
+      const questionnaire: Questionnaire = deepFreeze({
```
mverzilli: Either remove this type annotation or declare an intermediate variable with the literal object and apply `deepFreeze` to that. Otherwise this gives a false impression of type safety (because `deepFreeze` will type check no matter what you pass to it).

test/js/reducers/questionnaire.spec.js
```diff
@@ -1043,10 +1044,19 @@ describe('questionnaire reducer', () => {
       expect(resultState.data.quotaCompletedMsg.es.sms).toEqual('Listo')
       expect(resultState.data.quotaCompletedMsg.es.ivr).toEqual('Listo!')
     })
+
+    it('should compute a valid alphanumeric filename', () => {
```
mverzilli: Nice test! 💯

test/js/reducers/questionnaire.spec.js
```diff
@@ -1176,4 +1188,11 @@ const questionnaire = deepFreeze({
...
   languages: ['en'],
   quotaCompletedMsg: null
-})
+}
+
+// TODO: investigate why Flow ignores the result of `deepFreeze`
+// It probably is defined as `any` somewhere.
+// As a workaround, we define `bareQuestionnaire` and explicitly annotate it as
+// Questionnaire. That will let us catch inconsistencies when we define the
+// Questionnaire fixture for testing here.
```
mverzilli: Add the following in a new comment line: "The limitations of `deepFreeze` are probably related to sealed objects being used under its hood. See: https://flowtype.org/docs/objects.html#sealed-object-types"

web/static/js/decls/jquery.js
```diff
@@ -1 +1,3 @@
-declare var $: Function
+declare var $: Function;
+declare var it: Function;
```
mverzilli: Create a new `mocha.js` file in this same directory and move those two declarations there.

web/static/js/reducers/questionnaire.js
```diff
@@ -481,7 +481,7 @@ type ValidationState = {
 };
 
 const validateReducer = (reducer) => {
-  return (state: ValidationState, action: any) => {
+  return (state: ?ValidationState, action: any) => {
```
mverzilli: Add a comment saying: "React will call this with an `undefined` the first time for initialization. We mimic that in the specs, so ValidationState needs to become optional here."

Issue #554, opened 2017-01-16
SMS prompts spanning multiple SMS messages (Runtime)


mverzilli:
When an SMS prompt contains line breaks (`\n`), interpret them as message separators and send each line as an SMS. If the prompt contains a trailing `\n`, don't send an empty message, just ignore it.

edjez:
This is potentially dangerous as new lines in SMS are used for visual
separation too!
Also please test with copy and paste from different tools and the resulting
occasional /r/n

edjez:
Suggest "the last /n before the required message cutoff" also remember
cutoff length is a function of content in each message as they get
constructed.

mverzilli:
Thanks, Ed! We'll review the design then. As it is right now, ENTER confirms the message content, and SHIFT+ENTER would introduce new steps (`\n`). So there's no way to introduce line breaks from the UI without pasting the content. If we want to allow for multiple lines in a single SMS, that should change. 

Anyway the bulk of this issue is to make the runtime aware of purposeful multi-SMS messages, we can later change which char or char sequence we use as a separator.

edjez:
Thanks @mverzilli agree whatever separator char is a good idea in the meantime and agree /n may not be the best one - 2 backslashes \\ seem a good alternative too that doesnt have large impact on other things eg character counting.

mmuller:
Working properly in version b706119

PR #560, opened 2017-01-16
SMS prompts spanning multiple SMS messages

### LINKED ISSUE #553: SMS prompts spanning multiple SMS messages (Frontend)
As a questionnaire designer, I want to be able to enter questions longer than 140 chars. That means there will be questions I can't fit in a single SMSs, so I want to be able to specify how to split my question prompt in multiple SMSs.

# Graphics design 
You can find the related mockups here: https://projects.invisionapp.com/d/main#/console/8374235/193624221/preview
https://projects.invisionapp.com/d/main#/console/8374235/194469528/preview

# Interaction design
Notice that you introduce a new message by hitting SHIFT+ENTER while writing the prompt. 

When you click the X to remove a step, join that step's text to the previous one (instead of erasing the text).

# Data representation
Don't introduce any changes to the way SMS prompts are represented in a step. The only difference is that now we'll interpret the line break char (`\n`). For example:

1. Let's say there's a single message SMS prompt that says: "Peter Piper picked a peck of pickled peppers; A peck of pickled peppers Peter Piper picked; If Peter Piper picked a peck of pickled peppers, Where's the peck of pickled peppers Peter Piper picked?"
1. The user then introduces a new step (by pressing SHIFT+ENTER). That leaves the SMS prompt as:
"Peter Piper picked a peck of pickled peppers; \n A peck of pickled peppers Peter Piper picked; If Peter Piper picked a peck of pickled peppers, Where's the peck of pickled peppers Peter Piper picked?"

# Validations
Each chunk has to be shorter than 140 chars. There must be at least one step and it must not be blank. Only the last step can be blank, provided it's not the only one (the widget should actually make it impossible to introduce blank steps in the middle, so it is not necessary to validate this).

# Translations
Given we're not going to split the prompt in lines, instead keeping all content in a single line, translations should work just as they are. Notice that it makes sense for translators to introduce line breaks at different places based on their language.

asterite:
Fixes #553 and #554 

When the user presses SHIFT+ENTER in the **middle** of an SMS prompt input, the text is split in that position and another input appears. Do implement this, a `\n` is inserted in that position in the text and the inputs are rendered by splitting the text by `\n`.

Because `\n` is used as a delimiter, and because we always trim inputs' text, it's impossible to split a prompt at the beginning or end of it (this might be something good). Another consequence of this choice is that when the first or last input are empty they get automatically removed, because if the text either starts or ends with `\n` it's gone after being trimmed. This doesn't happen with intermediate inputs, though...

The UI isn't exactly like the proposed design ( https://projects.invisionapp.com/d/main#/console/8374235/193624221/preview and https://projects.invisionapp.com/d/main#/console/8374235/194469528/preview ) but it looks kind of good and I'm leaving the rest to a designer:

screenshot 2017-01-16 14 07 17

In the backend prompts are split by `\n` at the `Flow` step and also for SMS messages.

Because we'd likely want to have `\n` inside individual SMS messages, I suggest leaving this PR on hold until we decide what other character or sequence of characters to use, and that will also solve all of the above problems.

Missing:
- [ ] Validations
- [ ] Explanation on the UI on how to split (I tried putting `data-success` on the label but it had no effect)

I didn't do validations because I'd like to wait a bit on the final resolution for this issue.


steesdale:
Question: Is it possible to support different languages having a different number of SMS’s?   e.g., 1 for english and 2 for arabic?

asterite:
@steesdale Yes, the SMS prompt of one language doesn't affect other languages. You can split one language in 2 pieces and another one in 3, for example.

steesdale:
+100. Thanks

waj:
Let's use a control character to split the messages. We agreed to use this one: https://unicode-table.com/en/001E/

steesdale:
Question: why do we need this if we have the multiple sms feature per step in Ask?

- Scott

waj:
This IS the multiple SMS feature. When a SMS message is too long it's automatically split by mobile phones and network operators. However not all the phones can fit the pieces together back to the original text. That's why we want to split the messages *before* sending, so they arrive to the respondent as individual messages.

steesdale:
Is the candidate design still in place though?  I thought in the past there
were multiple lines that respond to each SMS?  Or will the multiple SMS's
be contained within a single line -- and the ?

- Scott

PR #596, opened 2017-01-26
A more strict JSON schema that defines specific attributes for each step type

waj:
I added definitions for each step type so not every attribute is allowed on all steps.

The problem with this is that error messages becomes more cryptic (almost unusable). That's why I refrained from committing this into master directly and send a PR instead. I'd like to hear other's opinions.

Also, when I ran the validation against my dev database (`mix ask.validate_json_in_db`) I got a lot of errors I had to manually fix (mostly badly generated questionnaires from old ages).


mverzilli:
I agree with this change in particular, and with any other effort put on specifying types more thoroughly (be it with JSON Schem on the database or with Flow in the client side code base). 

👍 from me

PR #645, opened 2017-02-10
Feature/642 next option for questionnaire last step

### LINKED ISSUE #642: Skip logic dropdown for the last step must display "next step" option
The current behavior is misleading. When a new step is added the skip logic option shows "End Survey", but if another step is added the skip logic will change to "Next Step". However, if the dropdown clicked and the "End Survey" item is selected, now when adding another step the skip logic will stay as "End Survey". In other words, there is a change in the behavior of the questionnaire without any visible UI changes.

To solve this, let's show both "Next Step" and "End Survey" options in the last step. This allows the user to choose the right option even when the following steps were not created yet.

lmatayoshi:
Fixes #642

### DIFF
--- a/lib/ask/runtime/flow.ex
+++ b/lib/ask/runtime/flow.ex
@@ -126,6 +126,14 @@ defmodule Ask.Runtime.Flow do
     {flow, %Reply{}}
   end
 
+  defp accept_reply(flow, :no_reply) do
+    if flow.retries >=  @max_retries do
+      {%{flow | current_step: flow |> end_flow}, %Reply{}}
+    else
+      {%{flow | retries: flow.retries + 1}, %Reply{}}
+    end
+  end
+
   defp accept_reply(flow, {:reply, reply}) do
     reply = reply |> clean_string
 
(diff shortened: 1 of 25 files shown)


lib/ask/runtime/flow.ex
```diff
@@ -81,7 +81,11 @@ defmodule Ask.Runtime.Flow do
 
     case skip_logic do
       nil ->
-        flow.current_step + 1
+        if (flow.current_step == flow.questionnaire.steps) do
```
waj: `flow.current_step` is an integer. `flow.questionnaire.steps` is an array. So this comparison is always false and the code executed is the same that existed before the change. Please, revert this change.

test/lib/runtime/flow_test.exs
```diff
@@ -399,6 +399,15 @@ defmodule Ask.FlowTest do
     assert prompts == ["Do you smoke?", "Reply 1 for YES, 2 for NO"]
   end
 
+  describe "only step with skip logic null" do
```
waj: This test has never failed without introducing changes to `flow.exs`. By the rules of TDD this test must be deleted :)

PR #742, opened 2017-03-09
Fixes #740. Cutoff is not working properly

### LINKED ISSUE #740: Cutoff is not working properly.
Found in version:
02bcf64

Preconditions:
There is a Questionnaire `Foo`.
'Foo' has enabled the SMS mode.

Reproduce steps:
Create a Survey.
Select the Questionnaire `Foo`.
Select `SMS` as Channel.
Upload 20 respondents.
Define a Cutoff rule equal to 1 respondent.
Start the Survey.
Answer the Survey using 1 respondent.

Obtained:
The Survey is sent to 10 respondents, and after that the Survey is sent to the `pending` respondents (10 in this test).

Expected:
`pending` respondents should not receive the Survey when the cutoff was covered.

lmatayoshi:
Fixes #740

### DIFF
--- a/lib/ask/runtime/broker.ex
+++ b/lib/ask/runtime/broker.ex
@@ -86,10 +86,10 @@ defmodule Ask.Runtime.Broker do
       reached_quotas = reached_quotas?(survey)
 
       cond do
-        reached_quotas || (active == 0 && ((pending + stalled) == 0 || survey.cutoff <= completed)) ->
+        (active == 0 && ((pending + stalled) == 0 || survey.cutoff <= completed || reached_quotas)) ->
           complete(survey)
 
-        active < batch_size() && pending > 0 ->
+        active < batch_size() && pending > 0 && !(survey.cutoff <= completed || reached_quotas) ->
           start_some(survey, batch_size() - active)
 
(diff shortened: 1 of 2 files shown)


lib/ask/runtime/broker.ex
```diff
@@ -86,10 +86,10 @@ defmodule Ask.Runtime.Broker do
       reached_quotas = reached_quotas?(survey)
 
       cond do
-        reached_quotas || (active == 0 && ((pending + stalled) == 0 || survey.cutoff <= completed)) ->
+        (active == 0 && ((pending + stalled) == 0 || survey.cutoff <= completed || reached_quotas)) ->
           complete(survey)
 
-        active < batch_size() && pending > 0 ->
+        active < batch_size() && pending > 0 && !(survey.cutoff <= completed || reached_quotas) ->
```
asterite: @lmatayoshi @nekron Maybe extract this condition to a variable? 

```
survey.cutoff <= completed || reached_quotas
```

With so many `&&` and `||` it's hard to follow the logic.

(before this change it was also like that, but we should improve the readability of this chunk of code)

PR #860, opened 2017-04-17
Allow more respondents to be added to an ongoing survey. Fixes #845

### LINKED ISSUE #845: Allow more respondents to be added to an ongoing survey
After a survey has been started, support adding more respondents.

We should allow only to add respondents to existing groups, NOT allow new respondents groups.

- [ ] Turn the existing group into a dropzone, with a prompt for adding or replacing the current respondents if the survey has not started yet, or just adding them if the survey is running.
- [ ] Notify the user that "n respondents have been added"
- [ ] Add an ADD MORE RESPONDENTS blue button to the left of the footer.
- [ ] Make the DELETE button grey.

asterite:
This implements #845. All the functionality is there though some UI things are missing. I'll list the existing items and what I've done and what's missing for each.

> Turn the existing group into a dropzone, with a prompt for adding or replacing the current respondents if the survey has not started yet, or just adding them if the survey is running.

This is done. However, these things are missing:
- [ ] An indication that shows that a file is being dropped and accepted (blue) or rejected (red) like what happens in the "Drop your CSV file here" box. Design is missing.
- [ ] A progress indicator showing that the file is being processed. Design is missing.
- [ ] Some way to show that the CSV file was invalid (for example, only letters were present). This is already shown in the "Drop your CSV file here" box, turning it all red, but doing the same for an existing group might be confusing. Design is missing for this too.

> Notify the user that "n respondents have been added"

How should this be notified? Design is missing.

> Add an ADD MORE RESPONDENTS blue button to the left of the footer.

Done!

> Make the DELETE button grey.

There's no such button, though there is a "Remove respondents" orange link. I don't understand if the link should always turn grey (but hidden once a survey starts) or just shown as grey (so, disabled) when the survey starts.

In any case, we should wait to release 0.8 before considering merging this PR.


nekron:
Let's talk to @jkicillof for design guidelines on the missing points.

waj:
I'll merge this and open new issue for the UI details

PR #864, opened 2017-04-18
Show error page for non-first user that opens the link. Fixes #863

### LINKED ISSUE #863: Show error page for non-first user that opens the link
Store cookie client-side with a code on first open, store that code on the respondent, and only allow to open the link if the code is not yet generated for a respondent, or it is supplied and matches the existing one.

Confirm the code in the server before start requiring it via a round trip to the client (ack). This is to avoid having a code set that none received due to a temporary connection loss.

asterite:
Wait until we close 0.8 to merge this.


nekron:
> Confirm the code in the server before start requiring it via a round trip to the client (ack). This is to avoid having a code set that none received due to a temporary connection loss.

We are missing this part.

asterite:
@nekron I don't understand what that means

In any case, this PR is wrong because it always uses the same name for the cookie. That means that a single browser will only be able to open one survey. The cookie should have in its name the id of the respondent.

Issue #857, opened 2017-04-13
[MobileWeb] Ensure there is a button drawn in half for respondent view


lmatayoshi:
Either add a hint to ensure the user notes there are more options if she scrolls down, or modify via js the whitespace between the question prompt and the buttons to ensure there is a button drawn in half at the bottom of the screen to suggest scrolling

waj:
I checked with @jkicillof and decided not to implement this feature. Instead an animated arrow will be displayed at the bottom of the page when there is more content to display (https://projects.invisionapp.com/d/main#/console/8374235/227391086/preview). I'll create a separate issue for this

PR #901, opened 2017-04-19
Feature/857 ensure button drawn in half

lmatayoshi:
Connects to #857


waj:
We decided not to implement this feature. Please check #857 for details.

PR #912, opened 2017-04-20
Fix/889 fix fallback ignored

fgasperij:


priv/repo/structure.sql
```diff
@@ -0,0 +1,560 @@
+-- MySQL dump 10.13  Distrib 5.5.54, for debian-linux-gnu (x86_64)
```
asterite: This file should probably not be commited

web/templates/coherence/email/password.html.eex
```diff
@@ -1,4 +1,4 @@
-<img src="<%= static_url(Ask.Endpoint, "/images/email-logo@2x.png") %>" alt="Surveda" width="114px" height="60px"/>
+<img src="<%= static_url(Ask.Endpoint, "/images/email-logo@2x.png") %>" alt="Surveda" width="150px" height="60px"/>
```
asterite: These changes don't seem to be related to the issue this PR solves

PR #974, opened 2017-05-10
Feature/946 new disposition states

### LINKED ISSUE #946: New disposition states
ncd - desired disposition diagram

nekron:
Fixes #946

### DIFF
--- a/lib/ask/runtime/broker.ex
+++ b/lib/ask/runtime/broker.ex
@@ -169,8 +169,19 @@ defmodule Ask.Runtime.Broker do
   end
 
   defp set_stalled_respondents_as_failed(survey) do
+    from(r in assoc(survey, :respondents), where: r.state == "stalled" and r.disposition == "queued")
+    |> Repo.update_all(set: [state: "failed", session: nil, timeout_at: nil, disposition: Flow.failed_disposition_from("queued")])
+
+
+    from(r in assoc(survey, :respondents), where: r.state == "stalled" and r.disposition == "contacted")
+    |> Repo.update_all(set: [state: "failed", session: nil, timeout_at: nil, disposition: Flow.failed_disposition_from("contacted")])
+
+
+    from(r in assoc(survey, :respondents), where: r.state == "stalled" and r.disposition == "started")
+    |> Repo.update_all(set: [state: "failed", session: nil, timeout_at: nil, disposition: Flow.failed_disposition_from("started")])
+
     from(r in assoc(survey, :respondents), where: r.state == "stalled")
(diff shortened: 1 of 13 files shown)


lib/ask/runtime/broker.ex
```diff
@@ -164,8 +165,19 @@ defmodule Ask.Runtime.Broker do
   end
 
   defp set_stalled_respondents_as_failed(survey) do
+    from(r in assoc(survey, :respondents), where: r.state == "stalled" and r.disposition == "queued")
```
fgasperij: Is the disposition history needed in this cases? Is it being created?

lib/ask/runtime/broker.ex
```diff
@@ -303,15 +316,20 @@ defmodule Ask.Runtime.Broker do
...
 
   defp sync_step_internal(session, reply, session_mode) do
-    respondent = session.respondent
-
     transaction_result = Repo.transaction(fn ->
       try do
+        session = if reply == Flow.Message.answer() do
+          update_respondent_disposition(session, "contacted")
+        else
+          update_respondent_disposition(session, "started")
```
fgasperij: If a respondent jumps from 'queued' to 'started', do the 'contacted' disposition history be created anyway?

test/lib/runtime/broker_test.exs
```diff
@@ -504,35 +571,184 @@ defmodule Ask.BrokerTest do
...
+    survey = Repo.get(Survey, survey.id)
+    assert survey.state == "completed"
+    respondent = Repo.get(Respondent, respondent.id)
+    assert respondent.state == "failed"
+    assert respondent.disposition == "unresponsive"
+
+    :ok = broker |> GenServer.stop
+  end
+
+  test "stalled respondents are marked failed after the survey is completed" do
```
fgasperij: 'When uncontacted'

lib/ask/runtime/broker.ex
nekron: No, it's not being created.

That condition will only be met if the delivery notification of an sms is missed. And only for that mode.

I think that having this missing transition is clearly an indication of something being wrong somewhere. I wouldn't fake it. I'd rather use it to try to find the problem.

lib/ask/runtime/broker.ex
nekron: Actually, the disposition shouldn't change in this case. Sending an additional commit to fix this.

test/lib/runtime/broker_test.exs
nekron: Fixed

PR #1009, opened 2017-05-19
Questionnaire preview. Fixes #999

### LINKED ISSUE #999: Questionnaire preview
Create a dummy survey and delete it when the popup is closed.

@jkicillof we need a design for a running survey preview

We should display the respondent answers as they are recorded.

Allow the user to select a channel in the test sandbox popup (we also need a design for this @jkicillof).

asterite:
Missing stuff that could probably go into other issues after we merge this:

- The icon used to bring up the dialog for launching a simulation is a gamepad (I couldn't find the "dummy" icon in the material icons)
- On the simulation screen the disposition text is shown. The fancy graph showing all the dispositions states is missing.

### DIFF
--- a/lib/ask/runtime/flow.ex
+++ b/lib/ask/runtime/flow.ex
@@ -300,7 +300,7 @@ defmodule Ask.Runtime.Flow do
     {current_step_index, total_steps}
   end
 
-  defp current_step(flow) do
+  def current_step(flow) do
     flow
     |> steps
     |> Enum.at(flow.current_step)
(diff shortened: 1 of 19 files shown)


waj:
Why did the tests in QuestionnaireControllerTest are deleted? Maybe reading the questionnaire snapshot is needed for the simulation, but updates and deletes should still be forbidden.

asterite:
@waj @nekron I put back the checks in `update` and `delete`

PR #1013, opened 2017-05-23
965: Adapted the SurveyShow for questionnaire comparisons.

### LINKED ISSUE #965: [Survey] When launching a survey, the screen only shows one of the questionnaire names
The [current design](https://projects.invisionapp.com/d/main#/console/8374235/231391853/preview) has a title that is the name of the survey's questionnaire. That design is old now, because a survey can have multiple associated questionnaires. Update the design to reflect this fact.

fgasperij:
Fixes #965


web/static/js/referenceColors.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+import range from 'lodash/range'
+
+const referenceColors = [
```
nekron: Replace this for css classes. For d3 use `.attr("class", "referenceColor0")`

web/static/js/referenceColors.js
```diff
@@ -0,0 +1,28 @@
...
+  '#cddc39', '#ffeb3b', '#ffc107', '#ff9800',
+  '#ff5722', '#f44336', '#e91e63', '#9c27b0'
+]
+
+const singleColor = '#4faf55'
+
+export function referenceColorClasses(totalNeeded: number) {
+  let nextColorIncrement = referenceColors.length / totalNeeded
+  return range(0, totalNeeded).map((i) => {
+    return 'referenceColor' + (i * nextColorIncrement)
```
nekron: It would be nice to define what happens if we exceed the maximum number of colors defined. Maybe we can find a way to reuse the same colors but changing the saturation.

At least we could try not to explode and show what we can with as much color differentiation as possible (ie, if more than 16 colors, show the remaining lines black, or reuse the colors). We can safely assume that we won't have more than 16 questionnaires, but for quotas I cannot be so sure.

PR #1236, opened 2018-02-16
Fixes #1224

### LINKED ISSUE #1224: [Autosave] Show "All changes saved" during the first minute
After the questionnaire is saved, instead of showing the current "Last saved 0 seconds ago", replace with the mesage "All changes saved" for the first minute.

juanboca:


web/static/js/components/layout/SaveStatus.jsx
```diff
@@ -3,17 +3,42 @@ import { connect } from 'react-redux'
...
+    }
   }
 
   render() {
-    const { saveStatus, routes } = this.props
+    const { saveStatus, routes, t } = this.props
 
     let show = false
 
+    const bindedFormatter = this.formatter.bind(this)
```
nekron: Can we bind this in the constructor? to avoid having to do it in every render cycle

web/static/js/components/layout/SaveStatus.jsx
```diff
@@ -3,17 +3,42 @@ import { connect } from 'react-redux'
 import TimeAgo from 'react-timeago'
 import { withRouter } from 'react-router'
 
+import { translate } from 'react-i18next'
+
 class SaveStatus extends Component {
-  static propTypes = {
-    saveStatus: PropTypes.any,
-    routes: PropTypes.any
+
```
nekron: Please avoid empty lines after the class declaration

PR #1285, opened 2018-03-24
Share channels with projects

### LINKED ISSUE #1248: [Channel] Share channel with projects
Clicking on a channel must show a page to edit the sharing of the channel with the existing projects
https://invis.io/3U8BARPXJ#/279907805_Edit_Channel_-_Share

### LINKED ISSUE #1249: [Survey] Show shared channels
Instead of listing the channels belonging to the current user, use the project channels

### LINKED ISSUE #1250: [Channel] Preload shared channels with existing data
Share channels with projects that have at least one survey using them

nekron:
Fix #1248 
Fix #1249 
Fix #1250


web/controllers/channel_controller.ex
```diff
@@ -1,23 +1,67 @@
...
+    channel =
+      Channel
+      |> Repo.get!(id)
+      |> authorize_channel(conn)
+      |> Repo.preload([:projects])
+
+    changeset =
+      channel
+      |> Channel.changeset(channel_params)
+      |> update_projects(channel_params)
```
waj: Is this missing an authorization over the project here? A user should be able to share a channel only with the projects where he/she is a member.

web/controllers/channel_controller.ex
```diff
@@ -1,23 +1,67 @@
...
-  alias Ask.Channel
+  alias Ask.{Channel, Project, Logger}
+
+  def index(conn, %{"project_id" => project_id}) do
+    channels = conn
+    |> load_project(project_id)
+    |> assoc(:channels)
+    |> Repo.all
+
+    render(conn, "index.json", channels: channels |> Repo.preload(:projects))
```
waj: I think it's better to use the `preload` before calling to `render`

web/controllers/channel_controller.ex
nekron: We should allow it if the user has a reader privilege? or we should enforce it to be admin/editor to be able to share?

PR #1420, opened 2018-06-18
Provide a bucket id on the survey view

adam704a:
The bucket id on the survey view allows reletively easy access to the quota target. We need this for this:  https://github.com/ICTatRTI/surveda-status-alerter


waj:
This was waiting here for too long. There are many new features now that might replace the need for this. Please reopen if still necessary.

PR #1432, opened 2018-08-21
Docs - Add Instance Setup guide & templates

matiasgarciaisaia:


matiasgarciaisaia:
Moved this to a Gist to avoid conveying the idea of this being an official, recommended way of setting up Surveda.

PR #1448, opened 2018-12-11
Feature/1444 show update progress when uploading questionnaire

### LINKED ISSUE #1444: Show upload progress when a questionnaire is being uploaded
Specially when the questionnaire being imported includes audios it takes some time to upload. Let's show some kind of progress indicator.

@jkicillof could you provide a design for this?

lmatayoshi:
Closes #1444. Show upload progress when a questionnaire is being uploaded


web/static/js/uploadManager.js
```diff
@@ -0,0 +1,50 @@
...
+  const formData = new FormData()
+  formData.append('file', file)
+  const xhr = new XMLHttpRequest()
+  const uploadId = random(100)
+
+  xhr.open('POST', url, true)
+
+  xhr.onreadystatechange = function() {
+    // state XMLHttpRequest.DONE == 4
+    if (this.readyState === 4 && this.status === 200) {
```
waj: Why not use the `XMLHttpRequest.DONE` constant here?

web/static/js/uploadManager.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+import random from 'lodash/random'
+
+const uploads = {}
+
+export const upload = (file: Object, url: string, onCompleted: ((response: Object) => mixed), onProgress: ((total: number, loaded: number) => mixed), onAbort: (() => mixed), onError: (description: ?string) => mixed) => {
+  const formData = new FormData()
+  formData.append('file', file)
+  const xhr = new XMLHttpRequest()
+  const uploadId = random(100)
```
waj: Let's use a uuid instead

web/static/js/uploadManager.js
```diff
@@ -0,0 +1,50 @@
...
+
+  xhr.open('POST', url, true)
+
+  xhr.onreadystatechange = function() {
+    // state XMLHttpRequest.DONE == 4
+    if (this.readyState === 4 && this.status === 200) {
+      delete uploads[uploadId]
```
waj: Seems like the upload is removed from the dictionary only when it finishes successfully. I think it should be always removed when the request is completed, regardless of the result.

web/static/js/uploadManager.js
```diff
@@ -0,0 +1,50 @@
...
+
+  xhr.send(formData)
+  Object.assign(uploads, {[uploadId]: xhr})
+  return uploadId
+}
+
+export const cancel = (uploadId: number) => {
+  if (uploads[uploadId]) {
+    uploads[uploadId].abort()
+    delete uploads[uploadId]
```
waj: Once the request in effectively cancelled, the `onreadystatechange` callback should be invoked with the `DONE` state so there is no need to remove the upload from the dictionary here.

PR #1460, opened 2019-03-21
Fix/activity log doesnt report section crud

### LINKED ISSUE #1419: Activity log doesn't include entries regarding the questionnaire sections.
Version: ed98638

Steps to reproduce:

- Create a Questionnaire.
- Add a section.
- Add a step to the section.
- Edit the step.
- Remove the step.
- Navigate to Project's `Activity`.
- Observe the log entries.

Obtained:

- Activity log doesn't include entries regarding the questionnaire sections.

hdf1986:
Fixes #1419 

Im not totally sure about the way i did this, mostly because im using recursion and a lot of ifs, but i tried to adapt the old logic without changing it too much


mverzilli:
It's alright, maybe a more idiomatic way would be to move those `if...else` blocks to functions where you'd pattern match to handle the different cases

Issue #1470, opened 2019-06-07
[Quotas] When set to 0 surveda doesn't reject those cases


manumoreira:
In this survey at Malawi RTI has set some quotas to 0 but lots of respondents on those quotas received the complete survey.
https://surveda-mw.org/projects/2/surveys/215/settings

waj:
Copied from @devduarte's PR:

> A new radio button was added for the survey cutoff step to set a default configuration. User should not be allowed to launch a survey when the configured cutoff or quotas buckets are zero and its radio button is checked accordingly. When the default configuration is checked in the step, both quota buckets and cutoff should empty and user should be allowed to launch the survey.
> 
> -If the user blanks the cutoff input, then a cutoff with a zero value is dispatched. Lower bound validation of cutoff was changed to cover this case.
> 
> An error message is displayed if there are no quotas to configure.
> 
> Sum of quotas is being calculated for both SurveyForm and SurveyWizardCutoffStep components, I think it will better to do this calculation in a reducer and use it from there.

mmuller:
Working properly in version f48c221

PR #1474, opened 2019-07-10
Issue #1470 fix

devduarte:
A new radio button was added for the survey cutoff step to set a default configuration. User should not be allowed to launch a survey when the configured cutoff or quotas buckets are zero and its radio button is checked accordingly. When the default configuration is checked in the step, both quota buckets and cutoff should empty and user should be allowed to launch the survey.

-If the user blanks the cutoff input, then a cutoff with a zero value is dispatched. Lower bound validation of cutoff  was changed to cover this case. 

- An error message is displayed if there are no quotas to configure.

- Sum of quotas is being calculated for both SurveyForm and SurveyWizardCutoffStep components, I think it will better to do this calculation in a reducer and use it from there.


matiasgarciaisaia:
Fixes #1470

web/models/survey.ex
```diff
@@ -67,7 +67,7 @@ defmodule Ask.Survey do
     |> validate_required([:project_id, :state, :schedule])
     |> foreign_key_constraint(:project_id)
     |> validate_from_less_than_to
-    |> validate_number(:cutoff, greater_than: 0, less_than: @max_int)
+    |> validate_number(:cutoff, greater_than: -1, less_than: @max_int)
```
waj: I'd rather use `greater_than_or_equal_to` instead of using a negative number as the limit (https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_number/3)

test/controllers/survey_controller_test.exs
```diff
@@ -676,13 +676,13 @@ defmodule Ask.SurveyControllerTest do
...
-    test "rejects update with correct error when cutoff field is less than zero", %{conn: conn, user: user} do
+    test "rejects update with correct error when cutoff field is less than -1", %{conn: conn, user: user} do
       project = create_project_for_user(user)
       survey = insert(:survey, project: project)
 
-      conn = put conn, project_survey_path(conn, :update, survey.project, survey), survey: Map.merge(@valid_attrs, %{cutoff: 0})
+      conn = put conn, project_survey_path(conn, :update, survey.project, survey), survey: Map.merge(@valid_attrs, %{cutoff: -1})
 
-      assert json_response(conn, 422)["errors"] != %{cutoff: "must be greater than 0"}
+      assert json_response(conn, 422)["errors"] != %{cutoff: "must be greater than -1"}
```
waj: The validation message will make much more sense if you apply the change suggested in `survey.ex`

Issue #1472, opened 2019-06-10
Prometheus metrics


waj:
Add the following Prometheus metrics to the existing endpoint:

### Respondent stats

This gauge can be collected on demand by selecting all the rows from `respondent_stats` table
```
surveda_respondents{
  survey_id,
  questionnaire_id,
  state,
  disposition,
  quota_bucket_id,
  mode
}
```

### Polling
These must be counters incremented by the broker
```
surveda_survey_poll{survey_id}
surveda_broker_respondent_start{survey_id}
```

### Verboice interactions
These must be counters incremented every time a call is enqueued in Verboice and status callback is received
```
surveda_verboice_enqueue{response_status}
surveda_verboice_status_callback{status}
```

### Nuntium interactions
These counters are incremented every time a message is enqueued in Nuntium or an incoming message or a status callback is received
```
surveda_nuntium_enqueue{response_status}
surveda_nuntium_incoming{}
surveda_nuntium_status_callback{state}
```

mmuller:
Working properly in version 9a11858

PR #1498, opened 2019-08-20
Prometheus metrics, first commit.

devduarte:
Changes and files for #1472

### DIFF
--- a/lib/ask/runtime/nuntium_channel.ex
+++ b/lib/ask/runtime/nuntium_channel.ex
@@ -2,7 +2,7 @@ defmodule Ask.Runtime.NuntiumChannel do
   @behaviour Ask.Runtime.ChannelProvider
   use Ask.Web, :model
   alias Ask.Runtime.{Broker, NuntiumChannel, Flow, Reply, ReplyStep}
-  alias Ask.{Repo, Respondent, Channel, Stats}
+  alias Ask.{Repo, Respondent, Channel, Stats, SurvedaMetrics}
   import Ecto.Query
   import Plug.Conn
   defstruct [:oauth_token, :name, :base_url, :settings]
@@ -65,6 +65,7 @@ defmodule Ask.Runtime.NuntiumChannel do
           _ -> :ok
         end
     end
+    SurvedaMetrics.increment_counter_with_label(:surveda_nuntium_status_callback, [state])
 
     conn |> send_resp(200, "")
   end
@@ -93,8 +94,8 @@ defmodule Ask.Runtime.NuntiumChannel do
             nil
         end
     end
-
     json_reply = reply_to_messages(reply, from, respondent)
+    SurvedaMetrics.increment_counter(:surveda_nuntium_incoming)
     Phoenix.Controller.json(conn, json_reply)
   end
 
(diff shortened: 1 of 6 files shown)


lib/ask/runtime/nuntium_channel.ex
```diff
@@ -280,9 +281,10 @@ defmodule Ask.Runtime.NuntiumChannel do
 
       respondent = NuntiumChannel.update_stats(respondent)
 
-      Nuntium.Client.new(channel.base_url, channel.oauth_token)
+      response = Nuntium.Client.new(channel.base_url, channel.oauth_token)
       |> Nuntium.Client.send_ao(channel.settings["nuntium_account"], messages)
 
+      SurvedaMetrics.increment_counter_with_label(:surveda_nuntium_enqueue, [elem(response, 0)])
```
waj: This will label the status with "ok" or "error" instead of the HTTP status code from Nuntium

lib/ask/runtime/verboice_channel.ex
```diff
@@ -383,8 +383,9 @@ defmodule Ask.Runtime.VerboiceChannel do
         {:ok, %{"state" => "completed"}} -> false
         {:ok, %{"state" => "failed"}} -> false
         {:ok, %{"state" => "canceled"}} -> false
-        {:ok, %{"state" => "expired"}} -> false
-        {:ok, %{"state" => _}} -> true
+        {:ok, %{"state" => _}} ->
+          SurvedaMetrics.increment_counter_with_label(:surveda_verboice_enqueue, [elem(response, 0)])
```
waj: This is not where a call is enqueued in Verboice. Move this to the `setup` function.

web/models/RespondentStatsMetrics.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Prometheus.RespondentStatsMetrics do
+  use Prometheus.Collector
+
+  alias Ask.Repo
+
+  def collect_mf(_registry, callback) do
+    respondent_stats_query = "select rs.* from respondent_stats rs"
```
waj: Why not use `RespondentStats` model?

web/models/RespondentStatsMetrics.ex
```diff
@@ -0,0 +1,42 @@
...
+    query_results = Repo.query!(respondent_stats_query)
+    columns = query_results.columns|> Enum.map(fn(column_name) -> String.to_charlist(column_name) end)
+    rows = query_results.rows |> Enum.map(fn(row) -> formatRow(row) end)
+    result_list =  Enum.map(rows, fn row ->
+                                      columns
+                                      |> Enum.zip(row)
+                                      |> Enum.into([])
+                            end)
+    callback.(create_gauge(
```
waj: This is not generating the gauge with the right shape. The `count` value from the table should be used as the value for each set of labels.

Also I'd not "reflect" on the columns. The `mode` column currently stores a JSON with an array of strings. Let's convert that into a comma separated string for the `mode` label in the gauge.

web/models/RespondentStatsMetrics.ex
```diff
@@ -0,0 +1,42 @@
...
+
+  def collect_metrics(:surveda_respondents, memory) do
+    Prometheus.Model.gauge_metrics(memory)
+  end
+
+  defp create_gauge(name, help, data) do
+    Prometheus.Model.create_mf(name, help, :gauge, __MODULE__, data)
+  end
+
+  defp formatRow(row) do
```
waj: Don't use camel case

web/models/SurvedaMetrics.ex
```diff
@@ -0,0 +1,60 @@
...
+      )
+    end
+
+    def increment_counter(counter_name) do
+      Counter.inc(
+        name: counter_name
+      )
+    end
+
+end
```
waj: If you use vscode, this can be avoided with these settings 😉 
```json
{
  "files.insertFinalNewline": true,
  "files.trimFinalNewlines": true
}
```

PR #1499, opened 2019-08-23
Add a column to the respondents CSV with the order of the sections

### LINKED ISSUE #1492: Add a column to the respondents CSV with the order of the sections
When a questionnaire has randomized sections, the order in which the sections is actually executed for a respondent is valuable information. Let's add a `section_order` column for these cases containing the name of the sections in the order they were (or going to be) executed for that respondent (separated by commas).

NOTE: the order of the sections is currently stored only during the respondent execution. After a respondent session is finished that info is discarded (the `session` column is set to `NULL`). So a new column must be added to the `respondents` table to persist this information. I suggest a `section_order` column containing a JSON with the index of the sections, just like it's currently stored within the session (for example: `[3, 0, 2, 1]`). When the CSV is generated these indexes will be converted to the section names.

aeatencio:
- [x] Add `section_order` on `respondent` when the session starts
- [x] Add `section_order` on the respondents CSV

 Fix #1492

### DIFF
--- a/web/controllers/respondent_controller.ex
+++ b/web/controllers/respondent_controller.ex
@@ -629,7 +629,7 @@ defmodule Ask.RespondentController do
         |> Enum.map(fn mode -> mode_label([mode]) end)
         |> Enum.join(", ")
 
-        row = row ++ [modes] ++ [Respondent.show_section_order(respondent.section_order)]
+        row = row ++ [modes] ++ [Respondent.show_section_order(respondent.section_order, respondent, questionnaires)]
 
         respondent_group = respondent.respondent_group.name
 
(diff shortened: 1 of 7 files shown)


web/controllers/respondent_controller.ex
```diff
@@ -629,7 +629,7 @@ defmodule Ask.RespondentController do
         |> Enum.map(fn mode -> mode_label([mode]) end)
         |> Enum.join(", ")
 
-        row = row ++ [modes]
+        row = row ++ [modes] ++ [Respondent.show_section_order(respondent.section_order, respondent, questionnaires)]
```
waj: Both lists can be collapsed into one to avoid one concatenation: `row = row + [modes, Respondent.show...]`

Also, the function always receives the `section_order` from the `respondent` and the `respondent` itself. So, actually only the `respondent` could be passed and the function should extract the `section_order` from it.

PR #1534, opened 2019-10-17
Retries histogram

### LINKED ISSUE #1526: Retries histogram
https://projects.invisionapp.com/share/2TMXFAVHVQM#/screens/241966468

Let's start with the "gray" portion of the chart. That means, it will display enqueued respondents and respondents waiting for the next retry.

aeatencio:
Fix #1526


web/models/survey.ex
```diff
@@ -427,6 +434,141 @@ defmodule Ask.Survey do
     }
   end
 
+  def retries_histograms(%Survey{mode: mode} = survey) do
```
matiasgarciaisaia: It may pay off to delegate all the histograms calculations to a different module, and to heavily test its component functions.

Instead of testing the whole histogram, it's easier to test (and understand) what each individual function is calculating, and how they cooperate to get to the final result.

test/models/survey_test.exs
```diff
@@ -111,4 +117,419 @@ defmodule Ask.SurveyTest do
...
+               %{type: mode, delay: 1, label: "1h"},
+               %{type: mode, delay: 2, label: "2h"},
+               %{type: mode, delay: 3, label: "3h"},
+               %{type: "end", delay: 5, label: "5h"}
+             ]
+    end
+
+    test "flow with retries" do
+      flow_with_retries("ivr")
+    end
```
matiasgarciaisaia: `flow_with_retries` is only being used once, and the second definition isn't used at all. There's probably something missing here 😇

web/models/survey.ex
```diff
@@ -427,6 +434,141 @@ defmodule Ask.Survey do
...
+    stats
+      |> RetryStat.count(%{attempt: attempt, mode: mode, retry_time: ""})
+
+  defp count_actives(%{stats: stats, now: now, attempt: attempt, mode: mode, delay: delay}) do
+    count_respondents(stats, now, attempt, mode, delay)
+  end
+
+  defp count_actives(_), do: 0
+
+  defp wating_respondents(%{stats: stats, mode: mode, attempt: %{idx: idx, absolute_delay: absolute_delay, delay: delay, current_mode: current_mode}, now: now, attempts: attempts}) when delay > 1 or (delay == 1 and attempts == idx and current_mode == "end") do
```
matiasgarciaisaia: Typo: `waiting_respondents` 😇

lib/ask/runtime/broker.ex
```diff
@@ -210,7 +210,8 @@ defmodule Ask.Runtime.Broker do
 
     Repo.transaction(fn ->
       try do
-        handle_session_step(Session.timeout(session))
+        substract_retry_stat(respondent)
```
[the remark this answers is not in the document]
aeatencio: I'm pretty sure it's working. So yes, I think the addition is being done in some other place. But I'm not sure at all about where is the best place to add and subtract stats in the current flow. So please let's take a look and try to figure it out together 🙏

PR #1535, opened 2019-10-29
remove underscore from mobile_survey

### LINKED ISSUE #1520: Remove underscore from mobile web URL
Currently the URL generated and sent via SMS to each respondent has an underscore. It was noticed that it causes issues with some carriers/devices: https://github.com/instedd/surveda/issues/1445

Let's remove that character from the URL replacing `mobile_survey` by `mobile`. The resulting URL would be: `https://<server>/mobile/<respondent_id>?token=<token>`

anaPerezGhiglia:
replace /mobile_survey/ with /mobile/
fixes #1520

### DIFF
--- a/lib/ask/runtime/session.ex
+++ b/lib/ask/runtime/session.ex
@@ -121,7 +121,7 @@ defmodule Ask.Runtime.Session do
   end
 
   defp url(respondent_id) do
-    shorten("#{mobile_base_url()}/mobile_survey/#{respondent_id}?token=#{Respondent.token(respondent_id)}")
+    shorten("#{mobile_base_url()}/mobile/#{respondent_id}?token=#{Respondent.token(respondent_id)}")
   end
 
   defp mobile_base_url() do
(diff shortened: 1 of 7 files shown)


lib/ask/runtime/session.ex
```diff
@@ -121,7 +121,7 @@ defmodule Ask.Runtime.Session do
   end
 
   defp url(respondent_id) do
-    shorten("#{mobile_base_url()}/mobile_survey/#{respondent_id}?token=#{Respondent.token(respondent_id)}")
+    shorten("#{mobile_base_url()}/mobile/#{respondent_id}?token=#{Respondent.token(respondent_id)}")
```
matiasgarciaisaia: Have you tried building the URL using the helper?

```elixir
shorten(Ask.Router.Helpers.mobile_survey_url(Ask.Endpoint, :index, respondent_id, token: Respondent.token(respondent_id))})`
```

I'm not sure if the host will be correct, thought 🤔

anaPerezGhiglia: yeah, I thought about it but wasn't sure if it was a good practice using the Helper for this and didn't know how to set a custom url with the helper as in `mobile_base_url()`
I could replace it though with
```elixir
shorten("#{mobile_base_url()}#{mobile_survey_path(Ask.Endpoint, :index, respondent_id, token: Respondent.token(respondent_id))})
```

PR #1558, opened 2019-11-19
Align dates and numeric fields right

### LINKED ISSUE #409: [Respondents] Values other than numbers or dates should be left-aligned in tables
Version: 0.4.0-dev-c76692f (build 1011)

At the moment all values are left aligned. **Numbers and dates** should be right aligned.

screen shot 2016-12-07 at 2 26 20 pm

aeatencio:
Fix #409

### DIFF
--- a/web/static/js/components/respondents/RespondentIndex.jsx
+++ b/web/static/js/components/respondents/RespondentIndex.jsx
@@ -14,6 +14,7 @@ import { modeLabel } from '../../questionnaire.mode'
 import find from 'lodash/find'
 import flatten from 'lodash/flatten'
 import { translate } from 'react-i18next'
+import classNames from 'classnames/bind'
 
 type Props = {
   t: Function,
@@ -441,11 +442,8 @@ class RespondentIndex extends Component<Props, State> {
               <th>{t('Disposition')}</th>
               <SortableHeader className='thDate' text={t('Date')} property='date' sortBy={sortBy} sortAsc={sortAsc} onClick={name => this.sortBy(name)} />
               {this.getModeAttempts()}
-              {respondentsFieldName.map(field => (
-                this.fieldIsNumeric(numericFields, field)
-                  ? <th className='thNumber' key={field}>{field}</th>
-                  : <th key={field}>{field}</th>
-                )
+              {respondentsFieldName.map(field =>
+                <th className={classNames({'thNumber': this.fieldIsNumeric(numericFields, field)})} key={field}>{field}</th>
               )}
               {variantHeader}
             </tr>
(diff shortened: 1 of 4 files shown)


web/static/js/components/respondents/RespondentRow.jsx
```diff
@@ -30,23 +30,23 @@ class RespondentRow extends Component<Props> {
...
+        {this.getModeAttemptsValues()}
         {responses.map((response) => {
           // For the 'language' variable we convert the code to the native name
           let value = response.value
           if (response.name == 'language') {
             value = languageNames[value] || value
           }
 
-          return <td className='tdNowrap' key={parseInt(respondent.id) + response.name}>{value}</td>
+          return <td className={`tdNowrap${response.isNumeric ? ' tdNumber' : ''}`} key={parseInt(respondent.id) + response.name}>{value}</td>
```
bcardiff:
```suggestion
          return <td className={classNames({'tdNowrap' : true, 'tdNumber' : response.isNumeric})} key={parseInt(respondent.id) + response.name}>{value}</td>
```

if `cellClassNames` is added:

```suggestion
          return <td className={cellClassNames(response.field)} key={parseInt(respondent.id) + response.name}>{value}</td>
```

web/static/js/components/respondents/RespondentIndex.jsx
```diff
@@ -400,13 +438,16 @@ class RespondentIndex extends Component<Props, State> {
...
-              )}
-              {variantHeader}
               <th>{t('Disposition')}</th>
-              <SortableHeader text={t('Date')} property='date' sortBy={sortBy} sortAsc={sortAsc} onClick={name => this.sortBy(name)} />
+              <SortableHeader className='thDate' text={t('Date')} property='date' sortBy={sortBy} sortAsc={sortAsc} onClick={name => this.sortBy(name)} />
               {this.getModeAttempts()}
+              {respondentsFieldName.map(field => (
+                this.fieldIsNumeric(numericFields, field)
+                  ? <th className='thNumber' key={field}>{field}</th>
+                  : <th key={field}>{field}</th>
```
bcardiff:
```jsx
<th className={classNames({'thNumber': this.fieldIsNumeric(numericFields, field)})} key={field}>{field}</th>
```

without need to split the two cases

web/static/js/components/respondents/RespondentIndex.jsx
```diff
@@ -431,6 +472,7 @@ class RespondentIndex extends Component<Props, State> {
               const responses = respondentsFieldName.map((field) => {
                 return {
                   name: field,
+                  isNumeric: this.fieldIsNumeric(numericFields, field),
```
bcardiff: Instead of adding `isNumeric` to each data, how about adding a `cellClassNames` property to `<RespondentRow>`.

```jsx
<RespondentRow cellClassNames={(fieldName) => classNames({
    'tdNowrap' : true, 
    'tdNumber' : this.fieldIsNumeric(numericFields, fieldName)})}
/>
```

That way it is clear that all cells of the same field will be styled in the same way.

web/static/js/decls/survey.js
```diff
@@ -119,7 +119,8 @@ export type Integration = {
 
 export type Response = {
   name: string,
-  value: ?string
+  value: ?string,
+  isNumeric: ?boolean
```
bcardiff: this change can go if `cellClassNames` is added.

aeatencio:
@bcardiff, I think we're ok now, but please let me know if otherwise. Thanks!

PR #1561, opened 2019-11-19
Add completion date in incentive CSV

### LINKED ISSUE #1113: Add date in Incentive file.
Feature requested by Charles Lau.

 

> As someone who manages the survey, I want to have a "date" field in the Incentive File, so I can easily distribute incentives.

> Right now, we need to provide incentives to respondents 2 x per week. If we don't have a date field, then I have to manually merge to determine which cases still require incentives. If I had a date field, I could just sort by date, and then cut and paste telephone numbers from the last X days (e.g, 3 days). This would make incentive distributions easy for users, and would reduce the likelihood that we would mistakenly not provide incentives to people.

aeatencio:
Fix #1113


aeatencio:
@bcardiff, I think we're ok now, but please let me know if otherwise.

PR #1563, opened 2019-11-21
Add autcomplete component for timezones

### LINKED ISSUE #297: [Survey] Replace Timezone select with search textbox
There are almost 600 timezone options. The control is unusable because you can't type to look for your timezone directly.

screen shot 2016-11-14 at 15 46 44

devduarte:
Fixes #297


web/static/js/components/surveys/SurveyWizardScheduleStep.jsx
```diff
@@ -2,10 +2,10 @@ import * as actions from '../../actions/survey'
 import * as uiActions from '../../actions/ui'
 import { connect } from 'react-redux'
 import React, { PropTypes, Component } from 'react'
-import TimezoneDropdown from '../timezones/TimezoneDropdown'
```
matiasgarciaisaia: `TimezoneDropdown` is not used anymore on the project, so let's delete it's source file too - and every other dependency that's not used anymore, if any.

Should we want to use it again in the future, we'll always have `git`.

web/static/js/components/surveys/SurveyWizardScheduleStep.jsx
```diff
@@ -125,7 +125,9 @@ class SurveyWizardScheduleStep extends Component {
           : ''
         }
         <div className='row'>
-          <TimezoneDropdown selectedTz={survey.schedule.timezone} onChange={this.updateTimezone} readOnly={readOnly} />
+          <div className='col s12 m6'>
+            <TimezoneAutocomplete selectedTz={survey.schedule.timezone} />
```
matiasgarciaisaia:
```suggestion
            <TimezoneAutocomplete selectedTz={survey.schedule.timezone} readOnly={readOnly} />
```
`TimezoneAutocomplete` seems to be missing a `readOnly` prop to only display the data without allowing editions. That's needed for the `Settings` tabs in a running survey.

web/static/js/components/timezones/TimezoneAutocomplete.jsx
```diff
@@ -0,0 +1,103 @@
...
+    if (this.timeZoneInput) this.timeZoneInput.focus()
+  }
+
+  getTimezoneInputRef = () => {
+    return this.timeZoneInput
+  }
+
+  componentDidUpdate() {
+    const { selectedTz } = this.props
+    this.focusTimezoneInput()
```
matiasgarciaisaia: Setting the focus on the component is making the page scroll down to the timezone autocomplete as soon as the New Survey page loads. This is definitely not the desired behaviour.

Is setting the focus needed at all?

web/static/js/components/timezones/TimezoneAutocomplete.jsx
```diff
@@ -0,0 +1,103 @@
...
+    this.setTimezoneInputVal(formatTimezone(selectedTz))
+  }
+
+  componentDidMount() {
+    const { dispatch, selectedTz } = this.props
+    dispatch(actions.setTimezone(selectedTz))
```
matiasgarciaisaia: Is this `dispatch` needed at all?

I understand it's triggering an Autosave of the survey as soon as the user visits the page - even if no change has been done at all. That autosave should not happen.

web/static/js/components/timezones/TimezoneAutocomplete.jsx
```diff
@@ -0,0 +1,103 @@
...
+    this.focusTimezoneInput()
+    this.setTimezoneInputVal(formatTimezone(selectedTz))
+    dispatch(fetchTimezones())
+  }
+
+  buildTimezonesforAutoselect(options) {
+    return options.map((element) => {
+      return { text: element, id: element }
```
matiasgarciaisaia: Shouldn't we `formatTimezone` the option that's shown on the dropdown while performing the autoselection?

With the code as-is, while you're typing `phno` you see an option like `Asia/Phnom_Penh`, and when you confirm it you see `Asia - Phonm Penh`.

PR #1593, opened 2019-12-13
Fix default for retry stat time in respondent

aeatencio:
For #1594 

The only current valid format is YYYYMMDDHH

Invalid values where set in 9a760ad


bcardiff:
Closing until fixed together with #1594

PR #1609, opened 2020-01-10
New calculations of rate stats for cockpit

### LINKED ISSUE #1600: Cockpit might show progress above 100%
There could be some confusion in the calculation regarding dispositions and states.

devduarte:
Fixes #1600

### DIFF
--- a/web/models/survey.ex
+++ b/web/models/survey.ex
@@ -425,7 +425,7 @@ defmodule Ask.Survey do
     completes = exhausted_respondents(respondents_by_disposition)
     multiplier = Float.ceil(1/estimated_success_rate, 0)
     pending = not_exhausted_respondents(respondents_by_disposition)
-    {needed, missing} = get_needed_missing(survey, respondents_by_disposition, respondents_target, pending, completion_rate, current_success_rate)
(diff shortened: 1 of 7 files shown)


devduarte:
Please double check batch size calculation of the broker since it uses the estimated_success_rate and this rate calculation was changed.

web/models/survey.ex
```diff
@@ -413,22 +413,19 @@ defmodule Ask.Survey do
...
     estimated_success_rate = estimated_success_rate(initial_success_rate, current_success_rate, completion_rate)
-    completes = respondents_by_state["completed"]
-    pending = respondents_target - completes
+    completes = exhausted_respondents(respondents_by_disposition)
     multiplier = Float.ceil(1/estimated_success_rate, 0)
-    needed = pending*multiplier
-    selected_respondents = respondents_count(respondents_by_state)
-    missing = max(needed - selected_respondents, 0)
+    pending = not_exhausted_respondents(respondents_by_disposition)
+    {needed, missing} = get_needed_missing(survey, respondents_by_disposition, respondents_target, pending, completion_rate, current_success_rate)
```
matiasgarciaisaia: To calculate the needed and missing respondents, we have to use the _estimated_ success rate instead of the actual. We're estimating how many respondents are we missing, so it makes sense to use the estimated rate for that.

Issue #1611, opened 2020-01-14
[Histogram] Same respondent appears twice


mmuller:
There are 10 respondents.
10 respondents received the first contact attempt.
One of them will receive the first retry.
The histogram shows 10 respondents in the initial stack and 1 respondent placed in the first retry.

![image](https://user-images.githubusercontent.com/518152/72361667-a35b7980-36d0-11ea-886b-ddf7c0ad42f8.png)

![image](https://user-images.githubusercontent.com/518152/72370408-f1787900-36e0-11ea-9301-966b76e2d4db.png)

This issue was also reproduced using the following:
survey_id: 2502
caller_id: 99925
call_id: 222802
call_id: 222803

anaPerezGhiglia:
Friday 2020-01-17 update (co-auth with @mmuller )

**Update:** this does not happen consistently, in some surveys happens in others not.
**Description:**
verboice called 2 times to the same respondent, the failure is `Call Rejected` for the first call, and for the second it is `Interworking, unspecified`. The distance between the calls is 1 minute. 
On the one hand, the histogram is not removing the respondent from the first bar; on the other hand, verboice is calling the same respondent after 1 minute
The respondent called twice has for retry_stat_id the id corresponding to the second-attempt slot

\* This also happens with nuntium

![image](https://user-images.githubusercontent.com/13237343/72745216-009e7180-3b8e-11ea-8ff7-17022a2320dc.png)

anaPerezGhiglia:
This seems to happen beacause there is a concurrency problem in Surveda. When this happens, `Ecto.StaleEntryError` in respondent entity is raised

The foundation of this issue seems to be that this error is being raised when starting new respondents. Since respondents are being started sequentially, when one fails the consequent ones are not started
Apparently this is happening after enqueuing calls/sms to verboice/nuntium, but before changing respondent's state


The flow would be:
 - respondent **A** has `state: pending`
 - respondent **A** is selected for starting
 - surveda enqueues respondent **A** in verboice/nuntium
 - respondent **A** is added to the retries_histogram
 - surveda tries to update respondent and fails with an uncaught exception - Ecto.StaleEntryError (respondent.state was not updated, thus is still `pending`)

In the next `Broker.poll` (scheduled once a minute), the respondent **A** still has `state: pending` so is eligible for starting so:
 - respondent **A** is selected for starting (again)
 - surveda enqueues respondent **A** in verboice/nuntium (again)
[...]

PR #1612, opened 2020-01-14
Deactivate the respondent inside a transaction

aeatencio:
Technical debt found in the context of #1611.
It might be related to it, but we don't expect it fixes it.

### DIFF
--- a/lib/ask/runtime/retries_histogram.ex
+++ b/lib/ask/runtime/retries_histogram.ex
@@ -18,16 +18,25 @@ defmodule Ask.Runtime.RetriesHistogram do
 
   def respondent_no_longer_active(respondent) do # only makes sense for verboice
     # transition from ivr active to normal retryStat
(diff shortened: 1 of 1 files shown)


aeatencio:
This PR ensures consistency in RetryStats just for one case (respondent deactivation). But we don't want different behaviors and designs depending on the different use contexts.
We won't pay attention to get consistency in RetryStats via transactions until we can run Ecto queries over a dedicated connection or a nested transaction. Maybe `Ecto.checkout/2` could help with that, but it's available since `v.3.0` and we're in `v2.1.6` right now.
The case that we know that can lead to inconsistencies run outside a transaction is `Ask.RetryStat.transition/2`. It was written to be run inside a transaction, but that option was discarded because MySql doesn't support nested transactions and the RetriesHistogram use context includes (sometimes) cases where it's run inside a transaction. 
Maybe we'll cover that case using a single query that makes the transition.

PR #1626, opened 2020-01-27
Feature/1068 add description field to questionnaire

### LINKED ISSUE #1068: Add description field to questionnaire
It would be nice to have an open-ended “description” field that I could use to remind myself what different questionnaires are. For example, we just switched over from a questionnaire that includes “data charges may apply” in SMS to a questionnaire that says “no cost to reply.” So we have 2 questionnaires that are 99% similar, with this one difference. The only place to track that difference (i.e., how is quex A different than quex B) is in a doc outside of the tool OR in my head (not a great idea!)

devduarte:
Fixes #1068 

Visible and editable in `Project -> Questionaires -> Questionnaire` page

questionnaire_description_before_edit

### DIFF
--- a/web/static/js/components/questionnaires/QuestionnaireIndex.jsx
+++ b/web/static/js/components/questionnaires/QuestionnaireIndex.jsx
@@ -141,7 +141,7 @@ class QuestionnaireIndex extends Component<any> {
               if (!questionnaire) return <tr key={-index} className='empty-row'><td colSpan={readOnly ? 3 : 5} /></tr>
 
               return (
-                <tr key={questionnaire.id}>
+                <tr key={questionnaire.id} title={questionnaire.description}>
                   <td onClick={() => this.goTo(questionnaire.id)}>
                     <UntitledIfEmpty text={questionnaire.name} emptyText={t('Untitled questionnaire')} />
                   </td>
--- a/priv/repo/migrations/20200122115222_add_description_to_questionnaires.exs
+++ b/priv/repo/migrations/20200122115222_add_description_to_questionnaires.exs
@@ -1,0 +1,15 @@
+defmodule Ask.Repo.Migrations.AddDescriptionToQuestionnaires do
+  use Ecto.Migration
+
+  def up do
+    alter table(:questionnaires) do
+      add :description, :text
+    end
+  end
+
+  def down do
+    alter table(:questionnaires) do
+      remove :description
+    end
+  end
+end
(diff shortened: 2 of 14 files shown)


anaPerezGhiglia:
# UI comments
 * questionnaire description makes the three-points-menu nex to questionnaire name to look weird (evidences that is not centered)

questionnaire_description_after_edit

* Questionnaire's description does not show anywhere else. As a user you still have to enter the questionnaire to see the description. Perhaps it would be nice to show the description in questionnaires list and/or when selecting a questionnaire when editing the survey? @manumoreira @bcardiff what do you think about this?

anaPerezGhiglia:
# Backend comments
Perhaps I'm missing something, but why adding a new endpoint `/set_description` instead of using the existing one that updates the whole questionnaire `controllers/questionnaire_controller.update()`?
The auto-save is being triggered whenever the state is being modified, so with only modifying it it should be enough. Take for example the questionnaire title: it does not have a specific endpoint for updating it when is modified
 
With this approach, the following code would be unnecessary:
 - `actions/questionnaire.js` -> unnecessary api call, should be as `changeName` action
 - `web/router.ex` -> no need for new endpoint `/set_description`
 - `questionnaire_controller.ex` -> no need for new function `set_description()`
 - `js/api.js` -> no need for new function `setQuestionnaireDescription()`

Moreover, with the current implementation, the elixir-backend is being contacted twice when editing the questionnaire description, once by `questionnaire_controller.set_description()` and once by `questionnaire_controller.update()`

devduarte:
I used the same criteria that is being used in `survey_controller.ex`, where the `set_description` endpoint is separated from the `update`. Questionnaires can get large depending on how many steps, sections and languages they have, so I thought that was convenient to create a new endpoint just for updating the description and not the whole structure. 
I didn't consider the autosave feature while coding this, so I think you are right and it would be better to use the `update` endpoint instead a new one.

bcardiff:
TODO:
* fix layout regarding three dots menu
* use the same endpoint to update de description and name so at least we don’t add a new endpoint. And apply the same save policy as in the name. I’m not aware if this means saving the questionare or not.
* let’s add the description as a tooltip if name for now in the listing page.

web/static/js/components/questionnaires/QuestionnaireIndex.jsx
```diff
@@ -141,7 +141,7 @@ class QuestionnaireIndex extends Component<any> {
               if (!questionnaire) return <tr key={-index} className='empty-row'><td colSpan={readOnly ? 3 : 5} /></tr>
 
               return (
-                <tr key={questionnaire.id}>
+                <tr key={questionnaire.id} title={questionnaire.description}>
```
anaPerezGhiglia: the `tr` title is the description but is not showing in the list

not_showing_description

anaPerezGhiglia: sorry my bad! I thought it should appear as a subtitle of the questionnaire not with the hover

The description is showing well with the hover 👌 
hover_description

Issue #1604, opened 2019-12-30
[Survey] Add clarification of localtime of the selected timezone


mmuller:
When selecting `Etc - GMT-5` the surveys uses `Etc - GMT+5`.

At 4pm (GMT-5) I got the following result:

![image (18)](https://user-images.githubusercontent.com/518152/71588700-37d6b100-2b01-11ea-8282-98d922d01f74.png)

![image (17)](https://user-images.githubusercontent.com/518152/71588701-37d6b100-2b01-11ea-9389-31e981edc22e.png)

Besides that some Capital cities (ie. Quito) are not included as options. We should improve the options listing.

bcardiff:
We will add a caption next to the autocomplete with the current local time of the selected timezone.

bcardiff:
The Etc - GMT-5 timezone corresponds to GMT+5  and vice versa. 
Since there the list of timezone have some missing locations the user is likelly to fall into these options.
The clarification of localtime of the selected timezone should mitigate the confusion.

mmuller:
Working properly in version 672562a

PR #1643, opened 2020-02-06
Add caption to timezones dropdown

danimiba:
In survey schedule step, add current time for selected timezone.
This is the frontend feature in #1604 

Screen Shot 2020-02-06 at 11 43 24
Screen Shot 2020-02-06 at 11 43 38

### DIFF
--- a/web/static/js/components/timezones/TimezoneAutocomplete.jsx
+++ b/web/static/js/components/timezones/TimezoneAutocomplete.jsx
@@ -66,8 +66,14 @@ class TimezoneAutocomplete extends Component {
     dispatch(actions.setTimezone(timezone.id))
   }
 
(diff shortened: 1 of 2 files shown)


web/static/js/components/timezones/TimezoneAutocomplete.jsx
```diff
@@ -66,8 +67,14 @@ class TimezoneAutocomplete extends Component {
     dispatch(actions.setTimezone(timezone.id))
   }
 
+  getTimeForTimezone(timezone, locale) {
+    var date = new Date(Date.now())
```
bcardiff:
```suggestion
    var date = new Date()
```

this already gives the current local time.

PR #1647, opened 2020-02-07
Improve survey stat tests

aeatencio:
- Apply the AAA pattern in every test
- Add error messages to every assert
- Make tests independent of the initial_success_rate config


aeatencio:
It was a good exercise for me, but this PR is too old. It won't be merged.

Issue #1649, opened 2020-02-10
Yarn check errors due to surveda-d3-components requirement


aeatencio:
Run
```bash
docker-compose run --rm webpack yarn check
```
Result: `error "@instedd/surveda-d3-components#react@16.x" doesn't satisfy found match of "react@15.4.2"`

PR #1651, opened 2020-02-10
Prevent dependencies check from being broken again

aeatencio:
### Wait for a solution of #1649 before merging

By running `yarn-check` in `pre-commit.sh` and CI

Currently crashing due to #1649


aeatencio:
Time out. Maybe in the future.

Issue #333, opened 2016-11-22
[Survey] Define quotas for multiple choice variables


waj:
Add the option in the cutoff step to define quotas based on multiple choice questions

nekron:
Quotas will be defined in a separated table with the following structure:
```
condition: {
  smokes: yes,
  gender: female
}, # the bucket definition
quota: 200, # the total number of respondents that must be contacted to fulfill the quota
count: 10, # the current number of respondents contacted
survey_id
```

On the other hand, the survey will have field named `quotas` that will contain an array with the selected variables that generate those buckets. This will be used mostly in runtime to know if the current question defines a quota or not. The idea is to avoid having to load all the buckets to know if the respondent fits into one of them for every question in the questionnaire.

The buckets will be sent to the API as a part of the survey, inside a `buckets` array.

PR #1654, opened 2020-02-14
Add canonical phone number

### LINKED ISSUE #1639: Duplicate respondents due to sanitized phone depending on channel
If a channel applies a pattern to a the respondent's phone number, Surveda updates the `sanitized_phone_number` field with that so it can find the respondent upon channel's callbacks.

When adding more sample to a running survey, the duplicates detection checks using `sanitized_phone_number` - but without running the numbers through the channel patterns (which could be more than one). This means we can end up with duplicate respondents (maybe even having two respondents with the same `phone_number` but different `sanitized_phone_number`).

We should add a new `canonical_phone_number` that stores the initial sanitation of the number (only digits, no dashes/plus/parens), and use that for dupes checking.

anaPerezGhiglia:
Add `canonical_phone_number` to respondents' table
When a new respondent_list file is added, respondents are compared by this new field instead of `sanitized_phone_number` (that changes depending on the channel) to determine whether the respondent has already been added to the survey

fixes #1639

### DIFF
--- a/test/lib/runtime/broker_test.exs
+++ b/test/lib/runtime/broker_test.exs
@@ -3838,7 +3838,8 @@ defmodule Ask.BrokerTest do
     [survey, group, _, _, _] = create_running_survey_with_channel_and_respondent()
 
     phone_number = "1-734-555-1212"
-    respondent = insert(:respondent, survey: survey, respondent_group: group, phone_number: phone_number, sanitized_phone_number: Ask.Respondent.sanitize_phone_number(phone_number))
+    canonical_phone_number = Ask.Respondent.canonicalize_phone_number(phone_number)
+    respondent = insert(:respondent, survey: survey, respondent_group: group, phone_number: phone_number, sanitized_phone_number: canonical_phone_number, canonical_phone_number: canonical_phone_number)
 
     {:ok, logger} = SurveyLogger.start_link
     {:ok, broker} = Broker.start_link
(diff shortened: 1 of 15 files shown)


test/lib/runtime/broker_test.exs
```diff
@@ -3838,7 +3838,8 @@ defmodule Ask.BrokerTest do
     [survey, group, _, _, _] = create_running_survey_with_channel_and_respondent()
 
     phone_number = "1-734-555-1212"
-    respondent = insert(:respondent, survey: survey, respondent_group: group, phone_number: phone_number, sanitized_phone_number: Ask.Respondent.sanitize_phone_number(phone_number))
+    canonical_phone_number = Ask.Respondent.canonicalize_phone_number(phone_number)
+    respondent = insert(:respondent, survey: survey, respondent_group: group, phone_number: phone_number, sanitized_phone_number: canonical_phone_number, canonical_phone_number: canonical_phone_number)
```
bcardiff: the factory can't be used set the sanitized and canonical number from the given phone_number?

web/models/respondent.ex
```diff
@@ -4,8 +4,9 @@ defmodule Ask.Respondent do
   alias Ask.{Stats, Repo, Respondent, Survey}
 
   schema "respondents" do
-    field :phone_number, :string
-    field :sanitized_phone_number, :string
+    field :phone_number, :string # phone_number as-it-is in the respondents_list
+    field :sanitized_phone_number, :string # phone_number with the channel's patterns applied `channel.apply_patterns(canonical_phone_number)`
```
bcardiff: 😍

test/lib/runtime/broker_test.exs
anaPerezGhiglia: yes, hadn't think about it, good idea! ✨ 
I will investigate a little bit more to know how exmachina works and how to capture the arguments passed in the insert, and add a commit with this

anaPerezGhiglia: @bcardiff  Apparently [Allow more control over factory definitions #333](https://github.com/thoughtbot/ex_machina/pull/333) was merged in [ex_machina 2.3.0](https://github.com/thoughtbot/ex_machina/blob/master/CHANGELOG.md#230) and surveda is currently using 1.0.2 😞 
It's a nice-to-have new feature

With 1.0.2, the result of setting a `phone_number` in de `insert(:respondent)` function but not the `canononical_phone_number`/`sanitized_phone_number` is a respondent with inconsistent data

PR #1655, opened 2020-02-14
Deprecated timezones

### LINKED ISSUE #1653: Deprecated timezones breaks in Chrome
**symptom:** 
The "Create Survey" UI fails when selecting "US/Eastern" timezone

**cause:**
The backend returns the list with all the timezones, even those that are deprecated and those that replace the deprecated ones (called links). 

Chrome no longer supports the deprecated timezones ("US/Eastern" links to "America/New_York") 
sources: 
https://stackoverflow.com/a/57711388
http://ftp.iana.org/tz/tzdb-2016g/backward

**solution:**
Surveda uses Timex for getting the timezones, which uses Tzdata that downloads the timezones data every day and has the timezones discriminated by `canonical` or `links`
The TimezoneController should return a map with `timezone -> target_timezone` for all the timezones even if it is a link or not, so the component doesn't have to discriminate. The `timezone` should be use for displaying in the drop-down-list and for configuring the survey schedule, while the `target_timezone` should be used for getting the current time

anaPerezGhiglia:
## Backend
Change surveda api (http://app.surveda.lvh.me/api/v1/timezones) to return a map of timezones instead of a list of them. 

This map links each timezone with its canonical, even if the timezone itself is a canonical one

For example (taken from de endpoint above):
**deprecated** -> canonical:
    "US/Eastern": "America/New_York"

**canonical** -> canonical: 
    "America/Argentina/Buenos_Aires": "America/Argentina/Buenos_Aires"

## Frontend
The `TimezoneAutocomplete` was modified to use the canonical timezone for calculating the current time in the selected timezone 

The `survey.schedule` still uses the generic timezone (canonical or deprecated), since this is what user sees

Fixes #1653

### DIFF
--- a/test/controllers/timezone_controller_test.exs
+++ b/test/controllers/timezone_controller_test.exs
@@ -9,9 +9,33 @@ defmodule Ask.TimezoneControllerTest do
     {:ok, conn: conn, user: user}
   end
 
-  test "returns Timex timezones", %{conn: conn} do
-    conn = get conn, timezone_path(conn, :timezones)
-    assert json_response(conn, 200)["timezones"] == Timex.timezones
-  end
+  describe "the timezones map" do
+    test "should link each deprecated timezone with its canonical timezone", %{conn: conn} do
+      conn = get conn, timezone_path(conn, :timezones)
+      timezones = json_response(conn, 200)["timezones"]
 
+      Tzdata.zone_alias_list()
+      |> Enum.each(fn deprecated_tz ->
+        assert timezones[deprecated_tz] == Tzdata.links()[deprecated_tz]
+      end)
+    end
+
+    test "should link each canonical timezone with itself", %{conn: conn} do
+      conn = get conn, timezone_path(conn, :timezones)
+      timezones = json_response(conn, 200)["timezones"]
+
+      Tzdata.canonical_zone_list()
+      |> Enum.each(fn canonical_tz ->
+        assert timezones[canonical_tz] == canonical_tz
+      end)
+    end
+
+    test "should contain all the timezones defined in Timex.timezones", %{conn: conn} do
+      conn = get conn, timezone_path(conn, :timezones)
+      timezones = json_response(conn, 200)["timezones"]
+
+      assert map_size(timezones) == length(Timex.timezones())
+      Timex.timezones() |> Enum.each(fn tz -> assert Map.has_key?(timezones, tz) end)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


test/controllers/timezone_controller_test.exs
```diff
@@ -9,9 +9,33 @@ defmodule Ask.TimezoneControllerTest do
     {:ok, conn: conn, user: user}
   end
 
-  test "returns Timex timezones", %{conn: conn} do
-    conn = get conn, timezone_path(conn, :timezones)
-    assert json_response(conn, 200)["timezones"] == Timex.timezones
-  end
+  describe "the timezones map" do
+    test "should link each deprecated timezone with its canonical timezone", %{conn: conn} do
```
aeatencio: I would add a test for the specific case we're solving. Something like "should link `US/Eastern` with `America/New_York`"

test/controllers/timezone_controller_test.exs
```diff
@@ -9,9 +9,33 @@ defmodule Ask.TimezoneControllerTest do
...
+      conn = get conn, timezone_path(conn, :timezones)
+      timezones = json_response(conn, 200)["timezones"]
+
+      Tzdata.zone_alias_list()
+      |> Enum.each(fn deprecated_tz ->
+        assert timezones[deprecated_tz] == Tzdata.links()[deprecated_tz]
+      end)
+    end
+
+    test "should link each canonical timezone with itself", %{conn: conn} do
```
aeatencio: I would add a test with a canonical timezone hardcoded. Something like "should link `America/New_York` with itself"

So we cover a specific use case and avoid making all tests depending on the same libraries used in the code we're testing.

test/controllers/timezone_controller_test.exs
anaPerezGhiglia: That case is covered by this test since is asserting all the non-canonical timezones

test/controllers/timezone_controller_test.exs
anaPerezGhiglia: The thing about doing this (or what you suggested in the comment above) is that perhaps in the future "America/NewYork" is no longer a canonical timezone. That is why I didn't test a specific timezone but test it against the Tzdata library that updates all the timezones data from the IANA everyday (or something like that)

aeatencio:
I know both situations are already indirectly covered. Also, I know that both proposed tests could eventually make the test suite break in the future. It's just that IMHO (or maybe it's just a feeling) we should anyway cover at least a very specific (hardcoded) use case in every behavior of `Surveda`.

PR #1660, opened 2020-02-23
Mobile web flow starts with an intro message

### LINKED ISSUE #1521: Remove validation of multiple respondents sharing a mobile survey URL
This is causing some trouble: https://github.com/instedd/surveda/issues/1445

Given that a URL cannot be used twice even from the same device, I think we could remove that validation.

aeatencio:
After that, we generate and check for the cookie, preserving the current flow

Fix #1521


priv/repo/migrations/20200213155156_add_mobile_web_intro_message_to_questionnaires.exs
```diff
@@ -0,0 +1,71 @@
...
+    Map.get(title, default_language)
+
+  defp mobile_web_intro_message(_settings, _default_language), do: nil
+
+  def up do
+    for q <- Repo.all(Questionnaire) do
+      mobile_web_intro_message = mobile_web_intro_message(q.settings, q.default_language)
+        if mobile_web_intro_message do
+          settings = Map.put(q.settings, "mobile_web_intro_message", mobile_web_intro_message)
+          Questionnaire.changeset(q, %{"settings" => settings}) |> Repo.update!()
```
bcardiff: does this migration updates fields with timestamp like update_at?

web/static/js/actions/questionnaire.js
```diff
@@ -54,7 +54,7 @@ export const CHANGE_EXPLANATION_STEP_SKIP_LOGIC = 'QUESTIONNAIRE_CHANGE_EXPLANAT
 export const CHANGE_DISPOSITION = 'QUESTIONNAIRE_CHANGE_DISPOSITION'
 export const TOGGLE_ACCEPT_REFUSALS = 'QUESTIONNAIRE_TOGGLE_ACCEPT_REFUSALS'
 export const CHANGE_REFUSAL = 'QUESTIONNAIRE_CHANGE_REFUSAL'
-export const SET_MOBILE_WEB_SMS_MESSAGE = 'QUESTIONNAIRE_SET_MOBILE_WEB_SMS_MESSAGE'
```
bcardiff: why this key is removed? the web sms message is no longer used?

web/static/mobile_survey/js/components/Step.jsx
```diff
@@ -74,22 +91,28 @@ class Step extends Component<any> {
...
+        case 'language-selection':
+          return <LanguageSelectionStep ref='step' step={step} onClick={value => this.handleValue(value)} />
+        case 'end':
+          return <EndStep ref='step' step={step} />
+        default:
+          throw new Error(`Unknown step type: ${step.type}`)
+      }
+    }
+    else {
+      // Before the first step fetch, show an intro message with user consent button
```
bcardiff: add here a clarification regarding when the cookie is set

web/templates/mobile_survey/index.html.eex
```diff
@@ -4,6 +4,7 @@
     window.mixEnv = "<%= Mix.env %>"
     window.respondentId = <%= @respondent_id %>
     window.token = "<%= @token %>"
+    window.introMessage = "<%= raw @mobile_web_intro_message %>"
```
bcardiff: Why `raw`?  Isn't this leading to some script injection?

priv/repo/migrations/20200213155156_add_mobile_web_intro_message_to_questionnaires.exs
aeatencio: Good question! I hadn't thought about that when I wrote the migration. I've tested it in my dev environment and running the migration up and down didn't change the `updated_at` field. Even when it changed the `settings` field. So I think we're OK. I looked for documentation regarding why it behaves differently from using the actual Questionnaire model, but I couldn't find anything about that.

web/static/js/actions/questionnaire.js
aeatencio: We abstracted this in order to reuse code and avoid adding a key for every text in the mobile web settings. So,  `intro_message` and `sms_message` share the same `SET_MOBILE_WEB_SETTING_TEXT` key here:

https://github.com/instedd/surveda/pull/1660/commits/d346cb911fcc18aa9a3ab317203e50871e68335a#diff-41b4d8946a0a0025e042fcf90b135cd2R480

It seems OK to me, but please let me know if I'm missing something

bcardiff: I see, I prefer to remove the abstraction and expand it for both fields. The rest of the codebase seems to follow that pattern. For example `setMobileWebSurveyIsOverMessage` is still present. setPrimaryColor/setSecondaryColor are not abstract. The only parametrised keys used are regarding the language which are dynamic.

web/templates/mobile_survey/index.html.eex
aeatencio: Good catch! Thanks! I've removed the raw() call and renamed the property name to clarify its content. It was the best solution I've found considering the class you give us about the subject (thanks a lot for that!) and a few script injection tests I did manually. If this solution isn't enough, let's iterate!

web/templates/mobile_survey/index.html.eex
```diff
@@ -4,6 +4,7 @@
     window.mixEnv = "<%= Mix.env %>"
     window.respondentId = <%= @respondent_id %>
     window.token = "<%= @token %>"
+    window.introMessageInnerText = "<%= @mobile_web_intro_message %>"
```
bcardiff:
```suggestion
    window.introMessageInnerText = <%= raw Poison.encode!(@mobile_web_intro_message) %>
```

PR #1664, opened 2020-03-04
IVR attempts count

### LINKED ISSUE #1589: Retries count is failing in surveda-ec
We have several cases in this survey https://surveda-ec.org/projects/3/surveys/24 with 3 in the ivr_attempts column but are still in queued or interim partial dispositions.
As the total amount of tries for the survey is 3 and the mode is `IVR only` after 3 retries there shouldn't be any respondent flagged as queued or interim partials with 3 retries.

[results&interactions.zip](https://github.com/instedd/surveda/files/3941412/results.interactions.zip)

matiasgarciaisaia:
Keep track of whether the last call sent to Verboice has actually been tried or not to omit it from the IVR attempts count if necessary.

Any contact from Verboice (except expiring the call) counts towards activity on the attempt.

Fixes #1589

### DIFF
--- a/lib/ask/ecto_types/stats.ex
+++ b/lib/ask/ecto_types/stats.ex
@@ -8,7 +8,7 @@ defmodule Ask.Stats do
             total_call_time_seconds: nil,
             call_durations: %{},
             attempts: nil,
-            last_call_started: false
+            pending_call: false
 
   def type, do: :longtext
 
@@ -23,7 +23,7 @@ defmodule Ask.Stats do
         total_call_time_seconds: total_call_time_seconds,
         call_durations: call_durations,
         attempts: attempts,
-        last_call_started: last_call_started
+        pending_call: pending_call
       }) do
     {:ok,
      %Stats{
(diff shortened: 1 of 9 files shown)


lib/ask/ecto_types/stats.ex
```diff
@@ -30,7 +32,8 @@ defmodule Ask.Stats do
        total_call_time: total_call_time,
        total_call_time_seconds: total_call_time_seconds,
        call_durations: call_durations || %{},
-       attempts: attempts
+       attempts: attempts,
+       last_call_started: last_call_started || false
```
bcardiff: I think that if last_call_started is null it should be translated to `true` in order to keep the same reports for existing data.

bcardiff: Handle this in data migration instead of code if possible.

bcardiff:
`pending_contact` / `last_attempt_pending` ? Otherwise it still says `call` ...

matiasgarciaisaia:
The flag only affects IVR, so `pending_contact` is more generic than it should. We could change it to `pending_ivr_contact` to avoid "call", but not more than that.

PR #1667, opened 2020-03-16
Synchronize Surveda main events

anaPerezGhiglia:
# Current State
Currently there are concurrent events on same respondents and there is no mechanism for controlling multiple updates to the respondent at the same time. As the schema `Ask.Respondent` has an optimistic lock with the field `lock_version`, when this happens `Ecto.StaleEntryError` is raised. This means that the last change is not applied and thus that new state is lost.

Analyzing surveda's logs and how it works, found that the things that collision the most are the Broker and the Channels (like VerboiceChannel). 

# Solution
The concurrency mechanism selected (after discussing with @waj and @bcardiff) was: use locks to synchronize this actions.
For this, [Mutex](https://hexdocs.pm/mutex/readme.html) dependency was chosen

At the moment of starting this, the `Ask.Runtime.Broker` had more than one responsibility, and made the decision of where to put the locks complicated and difficult for afterwards understanding why one function had a lock and another one didn't. That's why I split the broker in two:
 * `runtime/broker.ex` -> Is the one that uses `GenServer`. Actions that must be triggered by time passing. Iterates once a minute and is the responsible for polling running surveys, retrying respondents and so. Surveda is being “proactive"
 * `runtime/survey.ex` -> Is just a simple elixir module that has the logic of knowing how to handle the next step of the survey and updating the respondent (among a few other tiny things). 

With this change, there is a clearly separated layer

image

With this design, the places where to put the locks is much straightforward: lock every entry-point from the _TriggeringLayer_ to the _CoreLayer_ at the time we have a `respondent_id`. So, all the components at the "TriggeringLayer" are responsible of locking the action per-respondent and all the functions in _CoreLayer_ know nothing of locks

# Considerations:

1. The name `runtime/survey.ex` was discussed with @matiasgarciaisaia . If you do not agree with it, have reservations or opinions about it, please raise your hand now ✋ Re-naming it now will be easier
2. `runtime/broker_test.exs` was split also in two (new `runtime/survey_test.exs`), trying to separate the existing test according to the responsibilities of `runtime/broker` and `runtime/survey`. It was not an easy-task, and there may be tests that belong to the other file. If you identify one, please let me know 🙏 

# Recommendation for the reviewers:

Review this PR commit by commit instead of the whole thing at once, since I moved a lot of code from one file to another. I tried to be as neat as possible with the commits.


bcardiff:
Main diff can eb viewed at https://github.com/instedd/surveda/compare/0f7e634...43aa7d0

test/support/survey_stub.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Ask.BrokerStub do
+defmodule Ask.Runtime.SurveyStub do
```
aeatencio: Do you know what's this module for and how it works? Thanks!

lib/ask.ex
```diff
@@ -18,7 +18,8 @@ defmodule Ask do
       supervisor(Ask.Repo, []),
       # Start the endpoint when the application starts
       supervisor(Ask.Endpoint, []),
-      supervisor(Ask.MetricsEndpoint, [])
+      supervisor(Ask.MetricsEndpoint, []),
+      {Mutex, name: Ask.Mutex}
```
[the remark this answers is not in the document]
anaPerezGhiglia: This line is for starting the process that manages the locks. 

>  Using Mutex
A mutex is handled by a process that you start in your supervision tree with child_spec(name).
Options can be an atom (used as the GenServer name), or a Keyword with GenServer options and a :meta option to set the metadata.

test/support/survey_stub.ex
anaPerezGhiglia: This module is for mocking runtime/Survey responses. For what I saw, it's used for simplifying channel's test. 

For exaple: NuntiumChannelTest test `"callback with :prompts"`
```elixir
GenServer.cast(SurveyStub.server_ref, {:expects, 
  fn {:sync_step, %Respondent{id: ^respondent_id}, {:reply, "yes"}, "sms"} ->
    {:reply, ReplyHelper.multiple(["Hello!", "Do you exercise?"])}
  end})
```

With this, you're telling the Stub that when receives a `{:reply, ...^respondent_id}` in `sync_step` it should answer you with a `ReplyHelper.multiple(...)`

PR #1673, opened 2020-03-20
Name the unnamed test

aeatencio:
Because every test needs a name


bcardiff:
@aeatencio there are conflicts here. Can you check if we still need this is still since the broker was split?

aeatencio:
Yes, @bcardiff, it's still needed and the conflicts were solved.

BTW, the name for the test was taken from 48ee6a5 (the commit where the test was written)

PR #1692, opened 2020-04-17
Interim partial disposition by relevant steps

### LINKED ISSUE #1689: Implement backend logic for relevant questions for partial respondents

anaPerezGhiglia:
## With this changes:
 * a respondent can fetch `interim partial` disposition responding _n_ **relevant** steps
 * the _partial_ flag-step is still available and working as before

## Usage
For using this feature, user must configure the following properties in the questionnaire:
 * the min _n_ relevants steps that must be answered for considering the respondent `interim partial`
* the values to ignore in relevant steps. When a respondent replies with one of this values the step is not considered
* which are the steps in the questionnaire that are relevant

```
partial_config: {
  min_relevant_steps: int,
  ignored_values: string
}

step: {
  ...
  relevant: boolean,
  ..
}
```

**Code considerations**:
Decided to put the main logic in `runtime/session` since `runtime/flow` knows nothing from the respondent, and is needed for fetching the responses from the db

❗This PR contains a migration. Notify when merged

closes #1689


web/models/questionnaire.ex
```diff
@@ -16,6 +16,7 @@ defmodule Ask.Questionnaire do
     field :default_language, :string
     field :valid, :boolean
     field :deleted, :boolean
+    field :partial_config, JSON
```
bcardiff, line 19: Maybe call it

```suggestion
    field :partial_relevant_config, JSON
```

🙈  ?

lib/ask/runtime/flow.ex
```diff
@@ -501,6 +501,31 @@ defmodule Ask.Runtime.Flow do
...
+    steps(flow) |> Enum.any?(is_relevant_step)
+  end
+
+  def min_relevant_steps(%{questionnaire: questionnaire} = _flow) do
+    questionnaire.partial_config["min_relevant_steps"]
+  end
+
+  def relevant_response?(%{questionnaire: questionnaire} = _flow, response) do
+    quiz_step = questionnaire.steps |> Enum.find(fn step -> step["store"] == response.field_name end)
+    not_ignored = fn value -> String.upcase(value) not in ignored_values_from_relevant_steps(questionnaire) end
```
bcardiff, line 516: This is executing `ignored_values_from_relevant_steps` multiple times, right? Can we do it once only?

lib/ask/runtime/session.ex
```diff
@@ -424,6 +425,28 @@ defmodule Ask.Runtime.Session do
     session |> handle_step_answer(step_answer, current_mode)
   end
 
+  # If the respondent has answered at least `min_relevant_steps` relevant steps
+  # and the reply doesn't defines already a disposition
+  # then, 'interim partial' disposition is returned in reply
+  defp relevant_interim_partial_step({:ok, flow, %{disposition: nil} = reply} = step_answer, %{disposition: "started"} = respondent) do
+    new_step_answer = if Flow.interim_partial_by_relevant_steps?(flow) do # Filtered here to avoid fetching the responses unnecessarily
```
bcardiff, line 432: Besides checking if the step is relevant or not we are missing:
* to only set `interim partial` if the respondent is in `started`
* optimization, if the number of answers is less than `min_relevant_steps`, then there is nothing else to check.

What happen if the first question answered (which I guess is setting the respondent to `started`) it is also expected to set the respondent as `interim partial` ?

web/models/questionnaire.ex
anaPerezGhiglia: ✅done

lib/ask/runtime/flow.ex
anaPerezGhiglia: ✅ done 
now is stored if the `Flow` struct
Is calculated once: when the Flow is started or loaded from the db

lib/ask/runtime/session.ex
anaPerezGhiglia: 1. the function definition uses pattern matching to only apply for `disposition: started` respondents
``` elixir
defp relevant_interim_partial_step({:ok, flow, step_answer}, %{disposition: "started"} = respondent)
```
2. Thought about it at first, but when I coded it seemed to me like a premature optimization, so I removed it. The only thing that we would be saving would be the 
``` elixir
|> Enum.count(&Flow.relevant_response?(flow, &1))
```
that's traversing a collection in-memory and comparing it with in-memory values, since we would still have to go fetch the responses to the db

3. _if the first question answered it is also expected to set the respondent as interim partial_: ✅ works. Added a test to validate it

Issue #1672, opened 2020-03-17
Session simulator component


bcardiff:
Create a backend component or agent that given a questionnaire it will be able to handle in-memory sessions for different respondents.

Each session is independent and the goal is to persist nothing in the db. `Flow` and `Sessions` are the one with the semantics to move advance the state of the respondent.

Check that there is a way to extract the information to be shown in https://projects.invisionapp.com/d/main#/console/8374235/385278474/preview
* disposition
* questions state
* messages

This issue is for devolping the component and unit testing it. No need to integrate with the UI or expose it in the API,

anaPerezGhiglia:
## TODO list after #1696 

 - [ ] Currently the questionnaire is being adapted (from there its limitations with sections), and “sms_simulator” mode is used. Check if it is possible to not adapt the questionnaire but instead, use “sms” mode and use the questionnaire as-it-is (change in `QuestionnaireSimualtor` and `SMSSimulatorMode`). 
    - [ ] If the above is not possible, adapt questionnaire so it take into account the sections - use `Questionnaire.all_steps` function

- [ ] Respondent is being created with `disposition: queued`. Why is not being transitioned to `contacted` after starting the session and sending the first reply? According the [dispositions diagram](https://drive.google.com/file/d/0B21iplh0ozbVZTBYdXJ2RjZxMTQ/view) it should. Currently is transitioning from `queued` to `started`

- [ ] Implement logic to support __interim-partial by relevant steps__ feature. It doesn’t work since fetches responses from db

- [ ] When questionnaire doesn’t have a `thank-you-message` the last respondent-response isn’t registered in submissions field (because `{:end, respondent}` instead of `{:end, reply, respondent}`)

- [ ]  `QuestionnaireController` Should return 404 if questionnaire is empty (because does not belong to that project)

- [ ] Validate if questionnaire has “sms” as mode before launching simulation

- [ ] `RetriesHistogram` should not be created and modified for in-memory sessions

Off-topic but necessary
 - [ ] Change `Session.start` optional parameters to a map (_long-parameter-list_)

PR #1696, opened 2020-04-27
Questionnaire Simulator

anaPerezGhiglia:
# How it works:

### Request
* Each POST request to `/simulation` triggers a new simulation session with the same questionnaire
* POST request to `/simulation/message` to sync simulation, with json body: 
```
{
	"respondent_id": <respondent_id>,
	"response": <response>
}
```

### Response
* `”simulation_status"`: indicates whether the session is still active, has ended or expired.  possible values are `“active”`, `“ended”`, `“expired”`
* `”messages_history“`: contains all the messages exchanged during the simulation 
	* each message has a `type` field. Possible values are `“ao”` or `“at”`
* `"submissions"` contains all steps that where completed by the respondent:
	* numeric and multiple-choice steps are completed when respondent answered with a valid response
	* explanations steps are completed when showed to the respondent (don’t require a response)
* `"disposition"`: indicates current respondent disposition

* Thank you message & Error message don’t have an id (since aren’t steps? 🤷) but have title

## Simulator
`start_simulation`: From a given questionnaire and project, creates a new respondent, starts a new session and saves it in the store

`process_respondent_response`: From a given respondent_id and response, fetches the session from the store
* if the store has nothing indexed by that respondent_id, then session expired
* if there is a session for that respondent, the simulator syncs the simulation with the given response

## Store
* stores in memory all current sessions indexed by `respondent_id` (UUID)
* besides of storing the current simulation, also keeps track of the timestamp when that simulation was saved
* every minute the store inspects all stored simulations and deletes those that its ttl has expired
* the ttl is configurable from env_variable: `SIMULATION_TTL`, if not present, defaults to `5 minutes`
* since the granularity is 1 minute, the maximum time a simulations can live (since last updated) is 6 minutes (with the default ttl)

____________

# Main changes
* refactor `survey.update_respondent` since each definition was doing much more than just updating the respondent.  All the respondent modifications are now encapsulated in `survey.respondent_updates` and the remaining logic was bubbled-up to `handle_session_step` definition
 * `survey.handle_session_step` now returns the respondent
 * `survey.sync_step` now returns the respondent
 * Move the responsibility of applying a respondent changes to `Respondent` module 
 * add `persist` parameter to indicate whether the changes must be stored in the db, or only maintained in memory

# Limitations
* Only works for plain questionnaires (without sections)
* Questionnaires without thank-you-message: last response will not be registered in `submissions` field 
* interim-partial-by-relevant-steps feature doesn't work in simulator

_____________
## Api example
```
{
    "submissions": [
        {
            "step": "Introduction",
            "id": "fa827e04-aa66-4def-9b90-c97c4c037fb5"
        },
        {
            "step": "smoke",
            "response": "yes",
            "id": "6fcb2caf-2fc1-4528-96e7-093271724d50"
        }
    ],
    "simulation_status": "active",
    "respondent_id": "8728dc7c-cedd-4131-9aba-08de33f86635",
    "messges_history": [
        {
            "type": "ao",
            "title": "Introduction",
            "id": "fa827e04-aa66-4def-9b90-c97c4c037fb5",
            "body": "This is a survey that will be used for statistical purposes"
        },
        {
            "type": "ao",
            "title": "smoke",
            "id": "6fcb2caf-2fc1-4528-96e7-093271724d50",
            "body": "Do you smoke?"
        },
        {
            "type": "at",
            "body": "yes"
        },
        {
            "type": "ao",
            "title": "sleep",
            "id": "afc79422-7c2b-4d86-a611-7ba896ea00fd",
            "body": "Do you sleep 8 hours a day?"
        }
    ],
    "disposition": "started"
}
```

for #1672


lib/ask/runtime/survey.ex
```diff
@@ -43,138 +43,103 @@ defmodule Ask.Runtime.Survey do
...
+    updated_respondent = respondent_updates(:rejected, respondent, nil, nil, offline)
+    {:end, updated_respondent}
   end
 
-  def update_respondent(%Respondent{} = respondent, :rejected) do
-    respondent
-    |> Respondent.changeset(%{state: "rejected", session: nil, timeout_at: nil})
-    |> Repo.update!
+  def handle_session_step({:stopped, reply, respondent}, _, offline) do
+    updated_respondent = respondent_updates(:stopped, respondent, Reply.disposition(reply), offline) # TODO: update_respondent_and_set_disposition
```
matiasgarciaisaia, line 117: Is the `update_respondent_and_set_disposition` still pending? We should not forget about that.

lib/ask/runtime/survey.ex
```diff
@@ -43,138 +43,103 @@ defmodule Ask.Runtime.Survey do
...
-
-  def update_respondent(%Respondent{} = respondent, {:ok, session, timeout}, disposition, _) do
-    timeout_at = Respondent.next_actual_timeout(respondent, timeout, SystemTime.time.now)
-    update_respondent_and_set_disposition(respondent, session, timeout, %{session: Session.dump(session), timeout_at: timeout_at, disposition: disposition, state: "active"})
+    updated_respondent = respondent_updates(:failed, respondent, offline)
+    Session.log_disposition_changed(updated_respondent, session.current_mode.channel, mode, old_disposition, new_disposition)
+    disposition_changed(updated_respondent, session, old_disposition, offline) #TODO: this knows that the line before changed the disposition
   end
 
-  def update_respondent(%Respondent{} = respondent, :end, reply_disposition, _) do
```
matiasgarciaisaia, line 149: Is this case being considered? Was it deprecated somehow? The `respondent_updates(:end, #...` definition doesn't look anything like this one.

web/views/questionnaire_view.ex
```diff
@@ -25,4 +25,14 @@ defmodule Ask.QuestionnaireView do
       partial_relevant_config: questionnaire.partial_relevant_config,
     }
   end
+
+  def render("simulation.json", %{simulation: simulation}) do
+    %{
+      respondent_id: simulation.respondent_id,
+      simulation_status: simulation.simulation_status,
+      disposition: simulation.disposition,
+      messges_history: simulation.messages_history,
```
aeatencio, line 34:
```suggestion
      messages_history: simulation.messages_history,
```

lib/ask/runtime/survey.ex
anaPerezGhiglia: no, thanks!

lib/ask/runtime/verboice_channel.ex
```diff
@@ -287,13 +287,15 @@ defmodule Ask.Runtime.VerboiceChannel do
...
               num_digits = Reply.num_digits(reply)
               gather(respondent, prompts, num_digits)
-            {:end, {:reply, reply}} ->
+            {:end, {:reply, reply}, _} ->
               prompts = Reply.prompts(reply)
               say_or_play(prompts, channel_base_url(respondent)) ++ [hangup()]
+            {:end, _} ->
+              hangup()
             :end ->
               hangup()
```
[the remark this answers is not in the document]
anaPerezGhiglia: yes, the reason is this function definition:
```elixir
defp sync_step_internal(_, _, :invalid_mode, _, _) do
  :end
end
```

Now, taking a closer look, the first argument is `session` so I think we could return 
```elixir
{:end, session.respondent}
```
to unify the return types, what do you think?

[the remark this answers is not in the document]
matiasgarciaisaia: If we're sure that session has a respondent, it makes sense, yes 👍

lib/ask/runtime/survey.ex
anaPerezGhiglia: Yes, this was one of the functions that was doing way-much-more than just updating the respondent. If you take a look at 
```elixir 
handle_session_step({:end, reply, respondent}...
``` 
you will find out that the logic that is left over `respondent_updates` is there

test/lib/runtime/verboice_channel_test.exs
```diff
@@ -469,41 +469,48 @@ defmodule Ask.Runtime.VerboiceChannelTest do
...
-      {:ok, _broker} = Broker.start_link
+      {:ok, logger} = SurveyLogger.start_link
+      {:ok, broker} = Broker.start_link
       Broker.poll
 
       respondent = Repo.get(Respondent, respondent.id)
       assert respondent.state == "active"
 
-      {:ok, %{conn: conn, respondent: respondent}}
+      {:ok, %{conn: conn, respondent: respondent, logger: logger, broker: broker}}
```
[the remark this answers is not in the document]
anaPerezGhiglia: The explanation is in this commit: f3677fb

PR #1698, opened 2020-05-04
Sms simulator (front-end)

### LINKED ISSUE #1697: Session simulator frontend
Based on https://github.com/instedd/surveda/pull/1696 implement the UI on https://projects.invisionapp.com/d/main#/console/8374235/385278474/preview .

User should be able to do a full page reload to start a new session.

The chat component WIP is in `feature/sms-simulator` branch
The submission and state controls already exist but they might need to be decoupled from the store since in the simulator the response from the server should be used.

There is no need to store the simulated session in the store. It can all be kept in the state of the page component.

aeatencio:
This is how it looks:

active
random section skipped
random section first
refused

Fix #1697


web/static/js/components/questionnaires/QuestionnaireTitle.jsx
```diff
@@ -34,14 +34,20 @@ class QuestionnaireTitle extends Component {
...
-        <EditableDescriptionLabel description={questionnaire.description} emptyText={t('Add description')} onSubmit={(value) => { this.handleSubmitDescription(value) }} readOnly={readOnly} />
+        <EditableDescriptionLabel description={questionnaire.description} emptyText={readOnly ? '' : t('Add description')} onSubmit={(value) => { this.handleSubmitDescription(value) }} readOnly={readOnly} />
       </div>
     )
   }
 }
 
 const mapStateToProps = (state, ownProps) => ({
-  readOnly: state.project && state.project.data ? state.project.data.readOnly : true
+  readOnly: typeof ownProps.readOnly === 'boolean'
```
bcardiff, line 44: why this change?

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
```diff
@@ -34,14 +34,20 @@ class QuestionnaireTitle extends Component {
     return (
       <div className='title-container'>
         <EditableTitleLabel title={questionnaire.name} onSubmit={(value) => { this.handleSubmit(value) }} emptyText={t('Untitled questionnaire')} readOnly={readOnly} more={<QuestionnaireMenu />} />
-        <EditableDescriptionLabel description={questionnaire.description} emptyText={t('Add description')} onSubmit={(value) => { this.handleSubmitDescription(value) }} readOnly={readOnly} />
+        <EditableDescriptionLabel description={questionnaire.description} emptyText={readOnly ? '' : t('Add description')} onSubmit={(value) => { this.handleSubmitDescription(value) }} readOnly={readOnly} />
```
bcardiff, line 37: If it is readonly it should not even have an `onSubmit` attribute, right?

web/static/js/components/simulation/SimulationSteps.jsx
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  wasSkipped = (stepId: string): boolean => {
+    if (this.isActive(stepId) || this.isCompleted(stepId)) return false
+    const { currentStepId } = this.props
+    if (currentStepId) {
+      const { headedBySectionSteps } = this.state
+      return headedBySectionSteps.findIndex(step => step.id == stepId) < headedBySectionSteps.findIndex(step => step.id == currentStepId)
+    } else {
+      return true
```
bcardiff, line 92: Why if there is no current step all steps are marked as "skipped"?

The section headers are also marked as completed or they render different that question steps.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -0,0 +1,169 @@
...
+  state = {
+    simulation: null
+  }
+
+  fetchQuestionnaireForTitle = () => {
+    const { projectId, questionnaireId } = this.props
+    // TODO: this component doesn't do this fetch for itself
+    // It does it just because the title component above needs the questionnaire being in the redux store to render properly
+    // But for some reason that component isn't able to get the projectId and questionnaireId params properly from the url
+    this.props.questionnaireActions.fetchQuestionnaireIfNeeded(projectId, questionnaireId)
```
bcardiff, line 53: Aren't you missing a `withRouter` in the title component?

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
aeatencio: Because we don't want the title to be changed from the simulation page.

It's related to [this other change](https://github.com/instedd/surveda/pull/1698/commits/621d2946311c1e0361cf60df7cd00916bd456ddb#diff-628a34867fbc251edf9230cc7c96e629R66).

So, the intention here was: if the component's user explicitly decides to set it read-only (or the opposite) prioritize that.

web/static/js/components/simulation/SimulationSteps.jsx
aeatencio: The truth is that the case I'm covering defaulting to `true` here is this (simulation ended)

simulation ended

I'm not really sure if defaulting it to `true` is the best solution. What do you think?

Regarding the section headers, they are always being classified as skipped (because of the above default). I think visually they can be OK, but I'm pretty sure the semantic isn't good. I'll try to make it better and let you know if I find that I can't.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
aeatencio: I'm almost sure it wasn't because of that, but I'll try it again and let you know! Thanks!

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
bcardiff: ok

web/static/js/components/simulation/SimulationSteps.jsx
bcardiff: By default questions should not be marked as skipped.

Even if visually they look the same. For now it's fine if by default pending and skipped questions look the same, but let's not add a css class of skipped to default so we can iterate later on the styling if needed.

I don't follow why the headers of sections are marked as skipped. I would expect for them to be in the `headedBySectionSteps` list.

Maybe the step representing the header can also be marked as active/completed/pending=default based on the following rules:

* if the current step is of the section, then the section is active (note, the question step itself will also be active)
* if the current step is NOT of the section, and there are already answers within the section, then section is completed.
* otherwise, is pending=default.

Between these changes and knowing the step type, we can tweak later the styling to show progress in sections nicely I think. Again, I don't mind the styling in this PR.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
aeatencio: OK, yes, adding `withRouter` I was able to access the params.

Now, it's solved [here](https://github.com/instedd/surveda/pull/1698/commits/40b62e16f91fc789f5458a87bcfc0bb2b2559922). Thanks!!!

web/static/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -33,18 +34,27 @@ type Props = {
...
+  questionnaireActions: Object,
+  t: Function
 }
 
 type State = {
   simulation: ?Simulation
 }
 
-class QuestionnaireSimulation extends Component<Props, State> {
+const QuestionnaireSimulation = translate()(class extends Component<Props, State> {
```
bcardiff, line 42: connect with translate at the bottom of the file with the export default.

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
aeatencio: I tried, but `onSubmit` is required. Should we change that?

bcardiff: ok, no worries.

web/static/js/components/simulation/SimulationSteps.jsx
aeatencio: I think this problem was solved in [this commit](https://github.com/instedd/surveda/pull/1698/commits/f9981048f94c70b64358c328fee36bb7d2d9c489)

web/static/js/components/simulation/QuestionnaireSimulation.jsx
aeatencio: Done!

aeatencio:
@bcardiff, now I think it's ready for review! 🤞

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
```diff
@@ -30,18 +40,35 @@ class QuestionnaireTitle extends Component {
...
     )
   }
 }
 
-const mapStateToProps = (state, ownProps) => ({
-  readOnly: state.project && state.project.data ? state.project.data.readOnly : true
-})
-
-export default translate()(connect(mapStateToProps)(withQuestionnaire(QuestionnaireTitle)))
+export default translate()(withRouter(connect(mapStateToProps)(QuestionnaireTitle)))
```
bcardiff, line 74: I'm checking that `withQuestionnaire` is responsible of putting the quex in redux store into de property of the component. Why this need to change? Other usages of `fetchQuestionnaireIfNeeded` reveals that the components referenced in the routes.jsx like `QuestionnaireEditor`and `SurveySimulation` are the ones responsibles to load from params into the redux if needed. Leaving the `QuestionnaireTitle` to grab the quex via `withQuestionnaire`.

Is there any issue to move `fetchQuestionnaireIfNeeded` into the `QuestionnaireSimulation`?

web/static/js/components/simulation/SimulationSteps.jsx
```diff
@@ -0,0 +1,153 @@
...
+      return step.type == 'section'
+        ? !!step.steps && step.steps.some(st => submissions.some(su => su.stepId == st.id))
+        : submissions.some(su => su.stepId == step.id)
+    }
+
+    const wasSkipped = (step: Step, isActive: boolean, isCompleted: boolean, headedBySectionSteps: Array<Step>): boolean => {
+      if (isActive || isCompleted) return false
+      if (simulationIsEnded) {
+        // if the simulation is ended (and the step is neither active nor completed)
+        // then, it was skipped
```
bcardiff, line 67: Append comment that section will be marked as completed because isCompleted will return true.

Otherwise it seems that section will be marked as skipped with simulation ended.

web/static/js/components/simulation/SimulationSteps.jsx
```diff
@@ -0,0 +1,153 @@
...
+          step.steps
+        ]
+      } else {
+        return step
+      }
+    }, 2)
+    return {
+      headedBySectionSteps: headedBySectionSteps.map(step => ({
+        ...step,
+        status: (
```
bcardiff, line 94: extract to stepStatus function

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
aeatencio: Avoid fetching the questionnaire from `QuestionnaireSimulation` was done intentionally [here](https://github.com/instedd/surveda/pull/1698/commits/40b62e16f91fc789f5458a87bcfc0bb2b2559922).
The intention was to avoid fetching something that the component doesn't need for itself.
But I can revert that with no problem if you think it was better that way.

bcardiff: I think is best to keep the same design: allow the top-level component to do the fetch from the url. Otherwise the components like title end up depending on the url, instead of receiving props.

web/static/js/components/simulation/SimulationSteps.jsx
aeatencio: Done!

web/static/js/components/simulation/SimulationSteps.jsx
aeatencio: I modified how this is expressed in [this commit](https://github.com/instedd/surveda/pull/1698/commits/75372fcd8aacf8f74973733085179c0ac2b73b98), in the context of the `stepStatus` extraction.

web/static/js/components/questionnaires/QuestionnaireTitle.jsx
aeatencio: OK, so I'll revert that change. 👍

aeatencio: Done!

aeatencio:
Again, @bcardiff, ready for review! 🤞

PR #1700, opened 2020-05-06
QuestionnaireSimulator supports sections

anaPerezGhiglia:
* Refactor the codebase so `QuestionnaireSimulator` works with `"sms"` mode instead of `"sms_simulator"`. 
With this changes, the simulator uses the questionnaire as-it-is without adapting it, this means that no longer has the limitation of not working with questionnaires with sections

* Do not dump session if `persist = false`

for #1672

### DIFF
--- a/lib/ask/runtime/session_mode.ex
+++ b/lib/ask/runtime/session_mode.ex
@@ -11,7 +11,6 @@ defmodule Ask.Runtime.SessionModeProvider do
   defp mode_provider("sms"), do: SMSMode
   defp mode_provider("ivr"), do: IVRMode
   defp mode_provider("mobileweb"), do: MobileWebMode
-  defp mode_provider("sms_simulator"), do: SMSSimulatorMode
 
   def new(nil, _channel, _retries), do: nil
 
--- a/lib/ask/runtime/session_mode.ex
+++ b/lib/ask/runtime/session_mode.ex
@@ -14,7 +14,7 @@ defmodule Ask.Runtime.SessionModeProvider do
 
   def new(nil, _channel, _retries), do: nil
 
-  def new(_mode, %Ask.Runtime.SimulatorChannel{} = channel, retries) do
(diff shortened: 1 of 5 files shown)


lib/ask/runtime/session_mode.ex
```diff
@@ -11,9 +11,13 @@ defmodule Ask.Runtime.SessionModeProvider do
   defp mode_provider("sms"), do: SMSMode
   defp mode_provider("ivr"), do: IVRMode
   defp mode_provider("mobileweb"), do: MobileWebMode
-  defp mode_provider("sms_simulator"), do: SMSSimulatorMode
 
   def new(nil, _channel, _retries), do: nil
+
+  def new(_mode, %Ask.Runtime.SimulatorChannel{} = channel, retries) do
```
aeatencio, line 17: The reason we don't do this
```suggestion
  def new("sms", %Ask.Runtime.SimulatorChannel{} = channel, retries) do
```
is defensive programming?

Potentially (we don't expect it) we'd be using `SMSSimulatorMode` under unsupported mode scenarios

lib/ask/runtime/session_mode.ex
```diff
@@ -45,22 +49,22 @@ defmodule Ask.Runtime.SMSSimulatorMode do
 
   defstruct [:channel, :retries]
 
-  def new(channel, _retries), do: %SMSSimulatorMode{channel: channel, retries: []}
+  def new(channel, retries), do: %SMSSimulatorMode{channel: channel, retries: retries}
```
aeatencio, line 52: Do we care about retries in Simulator?

lib/ask/runtime/session_mode.ex
anaPerezGhiglia: You are right. There is no reason for doing it as it is

lib/ask/runtime/session_mode.ex
anaPerezGhiglia: No, but as the data was there it seemed to me unnecessary to ignore it

PR #1701, opened 2020-05-07
Simulator improvements

anaPerezGhiglia:
##  Respondents start simulation with `"contacted"` disposition
In real surveys, respondent transitions from `"queued"` to `"contacted"` when nuntium sends the delivery_confirmation through the status callback
Added a call to `delivery_confirm` in `QuestionnaireSimulator.start_simulation` to simulate this behavior

## Interim-partial-by-relevant-steps feature works in simulation
The `Ask.Respondent` struct has a field called `responses`. This is a consequence of defining `has_many :responses, Ask.Response` in its schema.

Used the `persist` parameter to determine if getting the responses from db o directly from `respondent.responses`. When storing responses, if `persist: false`, the responses are kept under this field

## Stop messages work in simulation
Implemented missing pattern `Reply.steps(_)`
Implemented `QuestionnaireSimulator.sync_respondent` for this feature (and probably others too) to work properly with only in-memory respondent

_____

for #1672

### DIFF
--- a/lib/ask/runtime/questionnaire_simulator.ex
+++ b/lib/ask/runtime/questionnaire_simulator.ex
@@ -52,7 +52,8 @@ defmodule Ask.Runtime.QuestionnaireSimulator do
       disposition: "queued",
       phone_number: "",
       canonical_phone_number: "",
-      sanitized_phone_number: ""
+      sanitized_phone_number: "",
+      responses: []
     }
 
     # Simulating what Broker does when starting a respondent: Session.start and then Survey.handle_session_step
@@ -109,7 +110,6 @@ defmodule Ask.Runtime.QuestionnaireSimulator do
   def handle_app_reply(simulation, respondent, reply, status) do
     reply_messages = reply_to_messages(reply) |> AOMessage.create_all
     messages = simulation.messages ++ reply_messages
-
     submitted_steps = simulation.submissions ++ SubmittedStep.build_from(reply, simulation.questionnaire)
 
     QuestionnaireSimulatorStore.add_respondent_simulation(respondent.id, %Ask.QuestionnaireSimulation{simulation | respondent: sync_respondent(respondent), messages: messages, submissions: submitted_steps})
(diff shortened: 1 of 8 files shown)


lib/ask/runtime/questionnaire_simulator.ex
```diff
@@ -124,8 +138,9 @@ defmodule Ask.Simulation.SubmittedStep do
     responses = Reply.stores(reply)
                 |> Enum.map(fn {step_name, value} ->
                   referred_step = questionnaire |> Questionnaire.all_steps |> Enum.filter(fn step -> step["store"] == step_name end)
-                  [id] = referred_step |> Enum.map(fn step -> step["id"] end)
-                  [title] = referred_step |> Enum.map(fn step -> step["title"] end)
+                  # hd function is used since there is a restriction that two steps cannot have the same store variable name
+                  id = referred_step |> Enum.map(fn step -> step["id"] end) |> hd
+                  title = referred_step |> Enum.map(fn step -> step["title"] end) |> hd
```
aeatencio, line 143: If I'm not missing something here, if we replace `filter` with `find` neither `map` nor `hd` are needed
```suggestion
                  referred_step = questionnaire |> Questionnaire.all_steps |> Enum.find(fn step -> step["store"] == step_name end)
                  %{"title" => title, "id" => id} = referred_step
```

lib/ask/runtime/questionnaire_simulator.ex
```diff
@@ -52,25 +52,40 @@ defmodule Ask.Runtime.QuestionnaireSimulator do
...
 
+  # Must update the respondent.session's respondent since if not will be outdated from respondent
+  # This is necessary since some flows:
+  #  - start using the respondent,
+  #  - then use the respondent.session
+  #  - and after that use the session.respondent
+  #
+  # If not synced, respondent data would be loss during simulation and the simulation could behave inaccurately
+  defp sync_respondent(respondent) do
+    synced_session = if respondent.session, do: %{respondent.session | respondent: respondent}, else: respondent.session
```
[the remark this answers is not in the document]
anaPerezGhiglia: yes, is the same

lib/ask/runtime/questionnaire_simulator.ex
anaPerezGhiglia: You are so right! thanks for spotting it! 🙌

PR #1718, opened 2020-05-27
Make QuestionnaireTitle menu optional

### LINKED ISSUE #1710: Remove menu from questionnaire simulation
That menu wasn't supposed to be there:

questionnaire simulation

Related to #1697
Bug from #1698

danimiba:
Now it doesn't appear in the simulation

Fixes #1710 

Related to #1697 and #1698

### DIFF
--- a/web/static/js/components/questionnaires/QuestionnaireTitle.jsx
+++ b/web/static/js/components/questionnaires/QuestionnaireTitle.jsx
@@ -12,7 +12,7 @@ class QuestionnaireTitle extends Component {
     dispatch: PropTypes.func.isRequired,
     questionnaire: PropTypes.object,
     readOnly: PropTypes.bool,
-    noMenu: PropTypes.bool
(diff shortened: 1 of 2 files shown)


web/static/js/components/questionnaires/QuestionnaireTitle.jsx
```diff
@@ -11,7 +11,8 @@ class QuestionnaireTitle extends Component {
     t: PropTypes.func,
     dispatch: PropTypes.func.isRequired,
     questionnaire: PropTypes.object,
-    readOnly: PropTypes.bool
+    readOnly: PropTypes.bool,
+    noMenu: PropTypes.bool
```
bcardiff, line 15: It's better to avoid negative properties. Either:

* rename it to `hideMenu`/`skipMenu`.
* use `menu` with default true value. using `menu={false}`

PR #1719, opened 2020-05-27
Replace cancelling with stopped in activity log

### LINKED ISSUE #1555: Log cancelling to cancelled change in activity log
In the activity log the cancelling, cancelled is not logged. it is logged as instantaneous

Ref: #1517

aeatencio:
Fix #1555

### DIFF
--- a/web/static/js/components/activity/ActivityDescription.jsx
+++ b/web/static/js/components/activity/ActivityDescription.jsx
@@ -69,11 +69,10 @@ class ActivityDescription extends Component {
           case 'start':
             return t('Started <i>{{surveyName}}</i> survey', {surveyName: surveyName})
           case 'request_cancel':
-            return t('Cancelling <i>{{surveyName}}</i> survey', {surveyName: surveyName})
-          case 'completed_cancel':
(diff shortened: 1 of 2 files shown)


web/static/js/components/activity/ActivityDescription.jsx
```diff
@@ -69,11 +69,10 @@ class ActivityDescription extends Component {
           case 'start':
             return t('Started <i>{{surveyName}}</i> survey', {surveyName: surveyName})
           case 'request_cancel':
-            return t('Cancelling <i>{{surveyName}}</i> survey', {surveyName: surveyName})
-          case 'completed_cancel':
-            return t('Cancelled <i>{{surveyName}}</i> survey', {surveyName: surveyName})
           case 'stop':
```
aeatencio, line 72: The `stop` event isn't being registered since [this change](https://github.com/instedd/surveda/commit/3ffaf2e882ffcd960727ef7e6f69bd4569df5c4a#diff-0041c7bf7c8e830ca79758e300963067L197). So, this line remains for backward compatibility.

PR #1734, opened 2020-06-04
Treat Partial and Interim partial as Completed in Survey metrics

### LINKED ISSUE #1726: Include interim partials in cockpit metrics
When a respondent reaches the interim partial disposition it can only go to Partial or Complete so when count partial as complete is selected in the survey settings there's no reason to exclude interim partials from the calculations.
To include interim partials we should add them into:
Exhausted
Partials

So far this change will impact in:

- Progress
- Actual success rate
- Estimated success rate
- Respondents needed
- Available sample

This change creates a more optimistic approach, the success rate will be higher and the number of respondents needed to complete target will decrease.
The positive side of this would be having a more accurate estimation in surveys that are counting partial as completes
The negative side is that as the success rate will increase and the survey might take longer to finish

aeatencio:
When "count partials as completed" is enabled, treat Partial and Interim Partial as Completed in the Survey metrics

Fix #1726

### DIFF
--- a/test/controllers/survey_controller_test.exs
+++ b/test/controllers/survey_controller_test.exs
@@ -459,7 +459,7 @@ defmodule Ask.SurveyControllerTest do
         "success_rate" => 0.667,
         "completion_rate" => 1.0,
         "estimated_success_rate" => 0.667
-      } = testing_survey(%{user: user, cutoff: 2, respondents: respondents})
+      } = testing_survey(%{user: user, respondents: respondents, attrs: %{cutoff: 2}})
         |> get_stats(conn)
     end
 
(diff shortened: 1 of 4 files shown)


aeatencio:
I don't get why `interim partial` when `count_partial_results` is being handled differently here:
 https://github.com/instedd/surveda/blob/ca3d196faac37c3325815c8612a6a5017de94d10/web/models/survey.ex#L451-L453
https://github.com/instedd/surveda/blob/ca3d196faac37c3325815c8612a6a5017de94d10/web/models/survey.ex#L467-L468

matiasgarciaisaia:
The rationale is that, since a respondent can't go _back_ in the disposition "flow", an interim partial respondent on a survey with the "Count partials as completed" flag enabled is _virtually_ completed.

That respondent can't go to any disposition that won't get it counted as a complete, so we count them as that. The only reason they're still counted as `interim partial`s is because we still have at least one more attempt left for them to keep answering questions.

Issue #1725, opened 2020-05-29
Survey respondents grid UI


aeatencio:
Create the following React components:
- Filter: textbox with a placeholder and hints
- Menu with the column checkboxes
- Dropdown with the page size
- Sorting (mind the content alignment)

For #1723 
Related to #1724

danimiba:
Design: https://projects.invisionapp.com/d/main#/projects

Issue #1724, opened 2020-05-29
Survey respondents grid API


aeatencio:
### Support:
- filtering (q=)
  - disposition:`<value>`
  - since:`<yyyy-mm-dd>`
  - `<expr>` `<expr>`
- sorting (`sortBy=field1`): currently supports sorting only by `phoneNumber` and `date`
- ~page-size (`pageSize=<value>`)~ (already supported)

### Response:
Add `fields` to `meta` property:
```
fields:
  [
    {name: t("Disposition"), key: "disposition" }, 
    {name: t("Phone"), key: "phone_number" }, 
    {name: "Do they drink beer?", key: "responses/do_they_drink_beer?" }
  ]
  [
    {name: t("Disposition"), key: "disposition" }, 
    {name: t("Phone"), key: "phone_number" }, 
  ] ++ dynamic_fields
```

### Important

CSV result file must honor the grid filters

For #1723

Issue #1723, opened 2020-05-29
Survey respondents grid improvements


aeatencio:
Follow [this mockup](https://projects.invisionapp.com/d/main#/console/8374235/417605359/preview)

![survey respondents grid mockup](https://user-images.githubusercontent.com/39921597/83265840-764bc700-a198-11ea-960f-77717cac84dc.png)

API support:
- filtering (q=)
  - disposition:`<value>`
  - date:`after <yyyy-mm-dd>`
  - `<expr>` `<expr>`
- sorting (`sortBy=field1`)
- page-size (`pageSize=<value>`)

UI components:
- Filter: textbox with a placeholder and hints
- Menu with the column checkboxes
- Dropdown with the page size
- Sorting (mind the content alignment)

PR #1735, opened 2020-06-04
Filter respondents using a search input

aeatencio:
## 🗒️ The needs

We need to filter respondents from the UI for #1723

Specifically, we decided to start supporting these filters:
- disposition
- since

During the development, we decided to add these other:
- mode
- state

See [the current version](https://github.com/instedd/surveda/wiki/How-to-filter-respondents/3db9ed2cb278648c93767e73d7b0c8b8b5256743) of [the Wiki page](https://github.com/instedd/surveda/wiki/How-to-filter-respondents)

## ✅ The current state

The current API (not the UI) for the CSV results file already supports filtering by:
- disposition
- since (`updated_at > value`)
- final (`r.state == "completed"`)

Every param is received in its own URL param. [Ona Connector](https://github.com/instedd/surveda-ona-connector) uses them

### For #1724 (Back-end)
- Support a new `q` URL param that supports the desired filters
- Do it using a new `RespondentsFilter` module
- Include unit tests for the new module
- Apply it to the respondents index and CSV results file
- Mind backward compatibility for ONA Connector
- When the same filter is received twice, prioritize the specific URL param

### For #1725 (Front-end)
- Add a search input in the respondents index
- Use the new `q` URL param to filter the respondents
- Deep-link using the same URL param in the UI
- Apply the filter debouncing the user input


lib/ask/respondents_filter.ex
```diff
@@ -0,0 +1,82 @@
...
+      {:since, value}, dynamic when value != nil ->
+        if (optimized) do
+          dynamic([_, r], ^dynamic and r.updated_at > ^value)
+        else
+          dynamic([r], ^dynamic and r.updated_at > ^value)
+        end
+
+      {:final, value}, dynamic when value != nil ->
+        if (optimized) do
+          dynamic([_, r], ^dynamic and ^value and r.state == "completed")
```
bcardiff, line 72: Shouldn't it be 

```suggestion
          dynamic([_, r], ^dynamic and r.state == ^value)
```

As in disposition?

aeatencio: I took it from [here](https://github.com/instedd/surveda/pull/1735/files#diff-d110c9705b856d9bca62542ea8d5f04aL550). I don't know the use case, so I don't want to change it.
My only intentional change was: if it receives a falsy value, avoid applying the filter.
So when `final=0` or `final=false` don't apply the filter and when `final=1` or `final=true` do it.
I'd done that change wrong and now I've fixed it [here](https://github.com/instedd/surveda/pull/1735/commits/90652e99c1219e8c96c61bc58adf06d25d471613). Technical debt: add a unit test for it.

bcardiff: Let's parse it as `status:completed` to support other status all together. 

If a query string final es pased on the api endpoint call `RespondentsFilter/put_status` with completed.

web/static/js/components/respondents/RespondentsFilter.jsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class RespondentsFilter extends Component<Props, State> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      inputValue: props.inputValue,
+      applyFilterDebounced: AwesomeDebouncePromise(
+        props.onApplyFilter,
+        props.debounceMilliseconds || 2000
```
bcardiff, line 26:
```suggestion
        props.debounceMilliseconds || 500
```

2 seconds is too much IMO.

web/static/js/components/respondents/RespondentsFilter.jsx
```diff
@@ -0,0 +1,73 @@
...
+        <input
+          value={inputValue}
+          type='search'
+          className='search-input'
+          onChange={this.onInputChange}
+          onKeyPress={this.onInputKeyPress}
+        />
+        <i className='material-icons grey-text'>search</i>
+        <span className='small-text-bellow'>
+          {t('Filter using the following format')}:&nbsp;
```
bcardiff, line 59: The translation key should be 

```
{t('Filter using the following format: {{example}}.', {example: "<em>disposition:completed since:2020-06-20</em>"})}
```

The example does not need to be translated.

web/static/js/components/respondents/RespondentsFilter.jsx
```diff
@@ -0,0 +1,73 @@
...
+          <em>{t('disposition:completed since:2020-06-20')}</em>.&nbsp;
+          <a
+            href='https://github.com/instedd/surveda/wiki/How-to-filter-respondents'
+            target='_blank'
+          >
+          View advanced search syntax
```
bcardiff, line 65:
```suggestion
          {t('View advanced search syntax')}
```

web/static/js/components/respondents/RespondentsFilter.jsx
```diff
@@ -0,0 +1,73 @@
...
+    const { applyFilterDebounced } = this.state
+    const inputValue = event.target.value
+    this.setState({ inputValue: inputValue })
+    onChange(inputValue)
+    applyFilterDebounced()
+  }
+
+  onInputKeyPress = (event) => {
+    const { onApplyFilter } = this.props
+    if (event.key === 'Enter') onApplyFilter()
```
bcardiff, line 41: This should cancel existing de promise execution. Otherwise a later request will fire.

web/static/js/components/respondents/RespondentsFilter.jsx
aeatencio: This is how it looks like when I do that:
filter hint

bcardiff: Ok, I wrongly read some docs in i18n package.

Let's remove the `t` in `<em>{t('disposition:completed since:2020-06-20')}</em>` and leave it directly as `<em>disposition:completed since:2020-06-20</em>` . We will not support i18n in the query

aeatencio:
@bcardiff, hopefully, we finished this PR! 🤞Please let me know!

mmuller:
Working properly in version de6acd1

PR #1745, opened 2020-06-23
Select which respondents index columns are shown

### LINKED ISSUE #1747: Add column picker in respondents table
For #1723

aeatencio:
# Column Picker

Let the user select which respondents index columns are shown

Fix #1747

## 📷 How it looks

column picker

## 📹 How it works

[column picker.mov.zip](https://github.com/instedd/surveda/files/5069303/column.picker.mov.zip)

## 📝 Collateral improvements

Part of the respondent index data manipulation was moved from the front-end to the API.
So the front-end can be mainly focused on visual presentation and user interaction.
We expect this page will deepen this path in the future.

## ❓Why the dropdown was replaced with a modal?

In the original mock-up (see #1723), the column picker was a dropdown with checkboxes.

But the only workaround we found to prevent the dropdown from closing on every user selection included some bad practiced we decided to avoid. See the discussion about this [here](https://github.com/instedd/surveda/pull/1745#discussion_r465821988).

Part of the problem is that we use a very old version of [react-materialize](https://github.com/instedd/surveda/blob/2820ea702bb028a9ebdf6022c68455d2ed2002c4/yarn.lock#L4971).

So after investing some effort trying to make it work using good practices with no results, we decided to use replace it with a modal.


web/static/js/components/respondents/RespondentIndex.jsx
```diff
@@ -453,15 +449,63 @@ class RespondentIndex extends Component<Props, State> {
...
+  toogleColumn(type, key) {
+    this.props.actions.toogleSelectedColumn(type, key)
+  }
+
+  columnsMenuOption({ type, key, displayText, checked }) {
+    const { selectedFields } = this.props
+    const inputId = `toggle_column_${fieldUniqueKey(type, key)}`
+    return (
+      <DropdownDropdownCheckboxItem
+        key={Math.random().toString()}
```
bcardiff, line 489: keys should never be random. It will trigger a recreation of the element on each render.

web/static/js/components/respondents/RespondentIndex.jsx
```diff
@@ -453,15 +449,63 @@ class RespondentIndex extends Component<Props, State> {
...
+        id={inputId}
```
bcardiff, line 490: I think is a code small to pass the id for the input here. See comment in DropdownDropdownCheckboxItem

web/static/js/components/ui/Dropdown.jsx
```diff
@@ -50,6 +52,44 @@ DropdownItem.propTypes = {
   className: PropTypes.string
 }
 
+export class DropdownDropdownCheckboxItem extends Component {
```
bcardiff, line 55:
```suggestion
export class DropdownCheckboxItem extends Component {
```

web/static/js/components/ui/Dropdown.jsx
```diff
@@ -50,6 +52,44 @@ DropdownItem.propTypes = {
...
+    return (
+      <li ref='li' className='checkbox'>
+        <input
+          id={id}
+          ref='input'
+          type='checkbox'
+          defaultChecked={defaultChecked}
+          className='filled-in'
+        />
+        <label htmlFor={id}>{displayText}</label>
```
bcardiff, line 87: if the label is the paren of the input, there will be no need to use htmlFor.

If that can't be done, see https://www.npmjs.com/package/react-html-id for how to generate ids locally

web/controllers/respondent_controller.ex
```diff
@@ -62,13 +102,9 @@ defmodule Ask.RespondentController do
 
   defp sort_respondents(query, sort_by, sort_asc) do
     case {sort_by, sort_asc} do
-      {"phoneNumber", "true"} ->
-        query |> order_by([r], asc: r.hashed_number)
-      {"phoneNumber", "false"} ->
-        query |> order_by([r], desc: r.hashed_number)
-      {"date", "true"} ->
+      {"updated_at", "true"} ->
```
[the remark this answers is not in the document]
aeatencio: No, it isn't. But I'll revert this change. Because it adds more confusion to other related issues we already have.

[Here](https://github.com/instedd/surveda/blob/master/web/views/respondent_view.ex#L17-L20) we're rendering the last response date of every respondent.

And [here](https://github.com/instedd/surveda/blob/master/web/controllers/respondent_controller.ex#L69-L72) we're sorting by the last respondent update.

We're rendering and sorting by different things. It's a bug, isn't it?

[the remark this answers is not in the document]
bcardiff: Ok, revert this change, and create a separate tech-debt issue to unify what data expose as respondent date.

web/views/respondent_view.ex
```diff
@@ -1,8 +1,11 @@
 defmodule Ask.RespondentView do
   use Ask.Web, :view
 
-  def render("index.json", %{respondents: respondents, respondents_count: respondents_count}) do
-    %{data: %{respondents: render_many(respondents, Ask.RespondentView, "respondent.json")}, meta: %{count: respondents_count}}
+  def render("index.json", %{respondents: respondents} = render_index_data) do
+    %{
+      data: %{respondents: render_many(respondents, Ask.RespondentView, "respondent.json")},
+      meta: render_index_meta(render_index_data)
```
anaPerezGhiglia, line 7: Much better now. Easier to understand what is being rendered 🎉

aeatencio: Totally! Thank you!!!

web/static/js/components/respondents/RespondentIndex.jsx
aeatencio: I know `key` isn't intended to be used this way. But this is the only way I found to force the component update. If I remove this line the component (and so the user selection) doesn't update properly.

It's related to [this unconventional solution](https://github.com/instedd/surveda/blob/0e0d0d2a8649415a0191f0d8a7c4491dfd1cd3ad/web/static/js/components/ui/Dropdown.jsx#L65-L73) to prevent the menu to be closed on every user selection.

Could we look for a better solution together, please? 🙏

web/static/js/components/ui/Dropdown.jsx
aeatencio: It seems to me that placing the input as a child of the label complicates styling. Because [_checkboxes.scss](https://github.com/instedd/surveda/blob/74cfc581a131719caf36879894a8f8f0b075e60d/web/static/vendor/css/materialize/components/forms/_checkboxes.scss) assumes that the label is after or before the checkbox.

Regarding the `react-html-id`, Surveda doesn't use it yet. Do you want to start using it now?

For the moment, I generated the id using `lodash.uniqueId`. I know it's not the same because the scope of the uniqueness is different. But we're already using it for similar purposes [here](https://github.com/instedd/surveda/blob/5bc348047de52eaab70397a20ebb747baaac3a10/web/static/js/components/ui/Dropdown.jsx#L18).

I propose opening a technical debt issue to start using `react-html-id` in the whole application.

web/static/js/components/respondents/RespondentIndex.jsx
aeatencio: Removed! Thanks!!

web/static/js/reducers/respondents.js
```diff
@@ -51,14 +54,37 @@ const createOrUpdateRespondent = (state, action) => ({
...
+  const selectedFields = state.selectedFields.some(key => key == fieldKey)
+    ? state.selectedFields.filter(key => key != fieldKey)
+    : [...state.selectedFields, fieldKey]
+  return {
+    ...state,
+    selectedFields
+  }
+}
+
+export const fieldUniqueKey = (type, key) => `${type}_${key}`
```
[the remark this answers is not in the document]
aeatencio: This function isn't exported for being used in component keys. Eventually, I could use it for a component key. But this isn't for what it's intended. The intention is to build a string that identifies a field uniquely vs any other field.

The API provides fields with `type` and `key`. The `type` could be any of `fixed`, `mode`, `variant`, `response`.

The `type` is useful to the front end to make decisions like [these](https://github.com/instedd/surveda/blob/5bc348047de52eaab70397a20ebb747baaac3a10/web/static/js/components/respondents/RespondentIndex.jsx#L589-L596).

The `key` of each field is unique scoped to its own type, and for getting a unique key vs any other field I used `fieldUniqueKey(type, key)` function as a convention.

web/static/js/components/ui/Dropdown.jsx
bcardiff: Using `lodash.uniqueId` is fine, but the value returned from it should be computed during component creation or mount and stored it in component state. Otherwise, each render will keep changing the id.

web/static/js/reducers/respondents.js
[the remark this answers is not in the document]
bcardiff: ok

web/static/js/actions/respondents.js
```diff
@@ -37,7 +38,8 @@ export const fetchRespondents = (
             response.result,
             filter,
             sortBy,
-            sortAsc
+            sortAsc,
+            response.respondentsFields
```
bcardiff, line 42: I understand that this respondentFields will be used as selected fields only in certain situations (empty selectedFields for example). But from this code that is not 100% clear.

PR #1750, opened 2020-07-15
Support long quex in SMS simulation

### LINKED ISSUE #1748: SMS Simulator focus on the quex last step
When using the sandbox and the SMS simulator the default view of the page is doing focus on the last step of the quex. In most of the cases, it makes losing the view focus from the SMS simulator.
defaultSMSSimView

aeatencio:
### 👷 Includes

- refactor simulation layout (using CSS grid)
- minor visual fix in disposition chart
- responsive web design (minimal/draft)

### 📹 Interaction video

- [video](https://drive.google.com/file/d/1FoN7MW-gbTPMF6SnIH_8gwRjL7Y2y2xF/view?usp=sharing)

### 📷 Screens

large
medium
small

Fix #1748

### DIFF
--- a/web/static/css/_quex-simulation.scss
+++ b/web/static/css/_quex-simulation.scss
@@ -3,6 +3,63 @@ $atMessageBackground: color("grey", "darken-4");
 $chatWindowBodyBackground: #999999;
 $chatBorderBackground: color("grey", "darken-3");
 
+// every simulation component heightens 8.5 simulation steps
+// the simulation step height depends on the screen size
+$simulationHeight: calc(74px * 8.5);
+$simulationHeightMediumAndDown: calc(72px * 8.5);
+
+// chatBodyHeight = chatTotalHeight - (chatHeaderHeight + chatFooterHeight)
+$chatWindowHeaderHeight: 90px;
+$chatWindowBodyHeight: calc((74px * 8.5) - (90px + 81px));
+$chatWindowBodyHeightMediumAndDown: calc((72px * 8.5) - (90px + 81px));
+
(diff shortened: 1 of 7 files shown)


web/static/css/_quex-simulation.scss
```diff
@@ -3,6 +3,63 @@ $atMessageBackground: color("grey", "darken-4");
...
 $chatBorderBackground: color("grey", "darken-3");
 
+// every simulation component heightens 8.5 simulation steps
+// the simulation step height depends on the screen size
+$simulationHeight: calc(74px * 8.5);
+$simulationHeightMediumAndDown: calc(72px * 8.5);
+
+// chatBodyHeight = chatTotalHeight - (chatHeaderHeight + chatFooterHeight)
+$chatWindowHeaderHeight: 90px;
+$chatWindowBodyHeight: calc((74px * 8.5) - (90px + 81px));
```
bcardiff, line 13: Maybe you are not able to mention the sass variables in a calc because they should be interpolated with `#{}`

```
$a: 4px;
$b: calc(#{$a} + 7px);
  
div {
  height: $b;
```

But in this case, since the input of the calculation are all px and not mixed with % you can avoid using css calc and let sass compute the value of the expression directly.

```
$a: 4px;
$b: $a + 7px;
```

aeatencio: Yes! I didn't know any of this! Please take a look again! Thanks!!

aeatencio:
One last review and I think we're done, @bcardiff 😇 Thanks!!!

PR #1754, opened 2020-07-28
Avoid rejecting respondents in completed dispositions

### LINKED ISSUE #1752: Failing on retry respondent that completed the quota on IVR
Quex: q1, set partial, q2
Quota: q1

Expected:

- no retry. 
	- a1, hang -> failed, partial
- 1 retry.
    - a1, hang  .... retrying ...  a2 -> completed, partial
    - a1, hang  .... retrying ...  hang -> failed, partial

Currently: rejected, interim_partial.

aeatencio:
Fix #1752 

##  🐛 The bug

Currently, the questionnaires don't go well for `interim partial` respondents that are part of a quota when the flag `Count partials as completed` is enabled in the survey.
We're rejecting them from the moment the quota is completed, even while the quota was reached necessarily because of them.

For _reasons_, it isn't a trivial issue for me using a non-IVR channel. So, I wasn't able to reproduce the bug using a non-IVR channel.

But... as far as I know, this bug shouldn't affect only IVR. It should affect every mode.

## 🕵️ What's happening

During every step (before asking) Surveda tests if the quota is completed. If so, it rejects the respondent.

There are two _free passes_ in this test:

1. the respondent was already rejected
2. the respondent isn't part of a quota

With the exception of the two preceding conditions, if the quota is completed, the respondent is rejected.

Currently, during this test, Surveda doesn't care if the quota was incremented because of the respondent that's being tested. So, it rejects them.

## 🔑 The clue

When the flag `Count partials as completed` is enabled, Surveda internally considers `interim partial` a completed disposition.

Also, every time a respondent reaches the two following conditions its quota is incremented by 1:

1. The respondent has a quota bucked assigned (they're part of a quota)
2. The respondent reached a completed disposition (including `interim partial` when `Count partials as completed` is enabled)

And this is OK because when `Count partials as completed` is enabled Surveda considers that there's enough information about a respondent that reached `interim partial` to increase the quota.

But... we don't want to stop asking questions to them until they finish or we've no more retries. We don't want to reject them because the quota is completed. The reason is simple: if the quota is completed it's necessarily because of them.

## 🔧 The fix

Add a third _free pass_ in the quota completed test (which determines the respondent rejection) for the two conditions already mentioned above:

1. The respondent has a quota bucked assigned (they're part of a quota)
2. The respondent reached a completed disposition (including `interim partial` when `Count partials as completed` is enabled)


lib/ask/runtime/session.ex
```diff
@@ -673,10 +681,43 @@ defmodule Ask.Runtime.Session do
...
+          Respondent.incremented_their_quota?(updated_respondent, count_partial_results)
+      else
+        Respondent.incremented_their_quota?(respondent, count_partial_results)
+      end
+
     cond do
-      flow.in_quota_completed_steps -> false
-      respondent.quota_bucket_id == nil -> false
+      # Was the respondent already rejected?
+      # flow.in_quota_completed_steps ~= (respondent.disposition == "rejected)
```
bcardiff, line 707: what does this comment mean?

aeatencio: I was trying to give the reader the clue that asking for `flow.in_quota_completed_steps` is pretty much the same as asking for `respondent.disposition == "rejected"` but I removed it because I see it looked rare.

lib/ask/runtime/session.ex
```diff
@@ -673,10 +681,42 @@ defmodule Ask.Runtime.Session do
...
+    # the respondent also must be tested using their new disposition.
+    respondent_incremented_their_quota? =
+      if Flow.should_update_disposition(respondent.disposition, reply_disposition) do
+        updated_respondent = %{respondent | disposition: reply_disposition}
+
+        Respondent.incremented_their_quota?(respondent, count_partial_results) ||
+          Respondent.incremented_their_quota?(updated_respondent, count_partial_results)
+      else
+        Respondent.incremented_their_quota?(respondent, count_partial_results)
+      end
```
anaPerezGhiglia, line 703: Look at this solution and tell what you think 

(I thought it considering that we changed `Repsondent.incrementer_their_quota?` parameters to receive the `bucket_it` and the `disposition`)

```elixir
inc_quota = fn disposition ->  Respondent.incremented_their_quota?(respondent.bucket_id, disposition, count_partial_results) end
    
respondent_inc_quota =
      inc_quota(respondent.disposition) || 
      Flow.should_update_disposition(respondent.disposition, reply_disposition) && inc_quota(reply_disposition)
```
For me is easier to understand, and we avoid duplicating `inc_quota(respondent.disposition)`, but please tell me what **you** think. Also, I haven't run and test it, so it could fail

test/lib/runtime/broker_test.exs
```diff
@@ -1238,4 +1281,72 @@ defmodule Ask.Runtime.BrokerTest do
...
+    insert_group_channel.("ivr", ivr_channel, ivr_group)
+    insert_group_channel.("sms", sms_channel, sms_group)
+
+    %{
+      ivr_respondent: insert(:respondent, survey: ivr_survey, respondent_group: ivr_group),
+      sms_respondent: insert(:respondent, survey: sms_survey, respondent_group: sms_group)
+    }
+  end
+
+  defp broker_poll(), do: Broker.handle_info(:poll, nil)
```
anaPerezGhiglia, line 1341: you can directly use `Broker.poll` that is intended for tests 🙂 
```elixir
  # Makes the broker performs a poll on the surveys.
  # This method is intended to be used by tests.
  def poll do
    GenServer.call(@server_ref, :poll)
  end
```

test/lib/runtime/broker_test.exs
```diff
@@ -1238,4 +1281,72 @@ defmodule Ask.Runtime.BrokerTest do
...
+
+  defp respondent_reply(respondent_id, reply_message, mode) do
```
anaPerezGhiglia, line 1343: Since this is a very helpful method, that is already being used in `retries_histogram_test`, perhaps we could move it to `Ask.TestHelpers` or into some of the support modules instead of having it duplicated 😬

lib/ask/runtime/session.ex
```diff
@@ -637,9 +642,12 @@ defmodule Ask.Runtime.Session do
   end
 
   defp assign_bucket_if_one_match([bucket], respondent, session) do
-    respondent = respondent |> Respondent.changeset(%{quota_bucket_id: bucket.id}) |> Repo.update!
+    respondent =
+    respondent |> Respondent.changeset(%{quota_bucket_id: bucket.id}) |> Repo.update!()
```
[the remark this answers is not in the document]
aeatencio: Yes, it was just my formatter that changed that. I applied it for the whole function.

I have installed these two extensions: [vscode-elixir-formatter](https://github.com/jakesorce/vscode-elixir-formatter) and [vscode-elixir-ls](https://github.com/JakeBecker/vscode-elixir-ls)

I'm not sure, but I think both of them use just the standard `mix format` command. I've tested both and they do the same for this line.

Specifically between these two options `Repo.update!` and `Repo.update!()` I think I would choose the one with the parenthesis, just because it's a function call, but... 🤷‍♂️

Would you like I revert it?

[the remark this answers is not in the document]
anaPerezGhiglia: No, if it's the formatter work, just leave it like that

lib/ask/runtime/session.ex
```diff
@@ -673,10 +681,42 @@ defmodule Ask.Runtime.Session do
...
+         in_quota_completed_steps,
+         count_partial_results
+       ) do
+    # The respondent disposition is updated by Runtime.Survey after the current step is handled
+    # by Runtime.Session, but the quota bucket could be assigned to the respondent and the quota
+    # incremented before the disposition update happens. So if the disposition should be updated
+    # the respondent also must be tested using their new disposition.
+    respondent_incremented_their_quota? =
+      if Flow.should_update_disposition(respondent.disposition, reply_disposition) do
+        updated_respondent = %{respondent | disposition: reply_disposition}
```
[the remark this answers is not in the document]
aeatencio: I agree. Updating the respondent this way smells. I'm changing it! Thanks!

lib/ask/runtime/session.ex
aeatencio: Do you see it better now? And... What do you think about the `Respondent.incremented_their_quota?` function. Was it a good decision? I tried to express in the codebase part of the knowledge we gained in these last days including it in the Respondent module.

anaPerezGhiglia: yes, it looks better now 🙂 
I believe it was a good decision encapsulating that logic and giving it a name. It sure is more descriptive now! 🎉

test/lib/runtime/broker_test.exs
aeatencio: Totally! Changed! Thanks!!!

test/lib/runtime/broker_test.exs
aeatencio: I've tried it without success. The error I get is the following:

```
root@75554624e88a:/app# mix test test/lib/runtime/broker_test.exs:204
Including tags: [line: "204"]
Excluding tags: [:test]

  1) test #1752 - respondents in completed dispositions shouldn't be rejected doesn't reject interim partial - IVR (Ask.Runtime.BrokerTest)
     test/lib/runtime/broker_test.exs:204
     ** (exit) exited in: GenServer.call({:global, Ask.Runtime.Broker}, :poll, 5000)
         ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
     stacktrace:
       (elixir) lib/gen_server.ex:766: GenServer.call/3
       test/lib/runtime/broker_test.exs:195: Ask.Runtime.BrokerTest.__ex_unit_setup_5/1
       test/lib/runtime/broker_test.exs:1: Ask.Runtime.BrokerTest.__ex_unit__/2

Finished in 2.2 seconds
50 tests, 1 failure, 49 skipped
```

`Broker.poll()` doesn't appear to be the same as `Broker.handle_info(:poll, nil)` Do you have a clue about their differences?

aeatencio: By the way, I also copied `broker_poll` from [here](https://github.com/instedd/surveda/blob/1873d9f8988e0a5f97e99a0f1bfe07c91944819c/test/lib/runtime/retries_histogram_test.exs#L453) in the retries histogram tests. If it stays, I should also avoid code duplication here, as we did with the `respondent_reply` function.

aeatencio:
@anaPerezGhiglia I think we're done now! Thanks, A LOT!!!

test/lib/runtime/broker_test.exs
anaPerezGhiglia: Right, I had forgot about this 🤦‍♀️ 

If you take a look at the trace, the process - GenServer - isn't alive, so the `GenServer.call` explotes since it doesn't find a process with that name. For using directly `Broker.poll` you have to first have the process running as a GenServer. You can take a look at `test/lib/runtime/survey_test.exs` where this is done: first they start it with 
```elixir 
{:ok, broker} = Broker.start_link
Broker.poll
```

Using directly the `Broker.handle_info(:poll, nil)` is the way of "by-passing" the GenServer and just using the logic it uses but without being run on a separate process - For what I understand, we can use it like this because the `Broker` does nothing with its internal state

aeatencio: 👍 I should learn deeply about how Surveda manages its processes. Maybe, in one of our pairing sessions, we can take a look together 😉

PR #1758, opened 2020-08-03
Select page size in Respondents index

### LINKED ISSUE #1757: Select page size in Respondents index
For #1723

aeatencio:
Fix #1757
 
page size

### DIFF
--- a/test/js/reducers/respondents.spec.js
+++ b/test/js/reducers/respondents.spec.js
@@ -11,10 +11,18 @@ describe('respondents reducer', () => {
   })
 
   it('should start fetching respondents', () => {
-    const surveyId = 200
-    const result = reducer(initialState, actions.startFetchingRespondents(surveyId, 1))
+    const [surveyId, pageSize, pageNumber, filter] = [1, 2, 3, '4']
+
+    const result = reducer(
+      initialState,
+      actions.startFetchingRespondents(surveyId, pageSize, pageNumber, filter)
+    )
+
     expect(result.fetching).toEqual(true)
     expect(result.surveyId).toEqual(surveyId)
+    expect(result.page.size).toEqual(pageSize)
+    expect(result.page.number).toEqual(pageNumber)
+    expect(result.filter).toEqual(filter)
   })
 
   it('should receive respondents', () => {
--- a/test/js/reducers/respondents.spec.js
+++ b/test/js/reducers/respondents.spec.js
@@ -25,6 +25,7 @@ describe('respondents reducer', () => {
     const result = reducer(r1, actions.receiveRespondents(respondents, respondentsCount))
     expect(result.fetching).toEqual(false)
     expect(result.items).toEqual(respondents)
+    expect(result.surveyId).toEqual(surveyId)
     expect(result.page.totalCount).toEqual(123)
   })
 })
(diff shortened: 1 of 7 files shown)


web/static/js/actions/respondents.js
```diff
@@ -101,3 +102,18 @@ export const sortRespondentsBy = (projectId, surveyId, newSortBy) => (dispatch,
   const newSortAsc = sortBy == newSortBy ? !sortAsc : true
   dispatch(fetchRespondents(projectId, surveyId, page.size, 1, filter, newSortBy, newSortAsc))
 }
+
+export const changePageSize = (projectId, surveyId, newPageSize) => (dispatch, getState) => {
+  const { page, filter, sortBy, sortAsc } = getState().respondents
+  if (page.size != newPageSize) {
+    // The following actions should be dispatched synchronously
+    // How could we garantee that?
```
aeatencio, line 110: @bcardiff Can you give me a hand with this? Thanks!!

test/js/reducers/respondents.spec.js
```diff
@@ -11,21 +11,59 @@ describe('respondents reducer', () => {
...
+
+  it('should reset current items while start fetching', () => {
+    const initialState = { surveyId: 1, items: {1: {id: 1}} }
+    const result = reducer(initialState, actions.startFetchingRespondents(1,
+      true
+    ))
+    expect(result.items).toEqual(null)
+  })
+
+  it('shouln\' reset current items while start fetching', () => {
```
bcardiff, line 49:
```suggestion
  it('should not reset current items while start fetching', () => {
```

web/static/js/actions/respondents.js
```diff
@@ -10,34 +11,31 @@ export const RECEIVE_RESPONDENTS_STATS = 'RECEIVE_RESPONDENTS_STATS'
 export const fetchRespondents = (
   projectId,
   surveyId,
-  limit,
-  page,
-  filter = '',
-  sortBy = null,
-  sortAsc = true
+  resetCurrentItems,
```
bcardiff, line 14: I see no need for this argument. The respondent's reducer has a logic that will reset the items if the action's survey differ from the current store state.

aeatencio: OK, thanks for pointing this! I'll try to remove it, while keep everything working 😅

aeatencio:
I'm splitting this PR to avoid adding extra-complexity and delay to the #1757 solution 

I removed the #1759 draft, saved it locally, and removed the following description from this PR.

## Respondents index user decisions persistence

### 🗓️ How the things are today

[Here](https://github.com/instedd/surveda/blob/de6acd151ac0c0d043531d87d7e41cdc5890d895/web/static/js/components/respondents/RespondentIndex.jsx#L82-L118) we handle the user decisions regarding page number and filter from the `RespondentIndex` component.
The page number is set to 1 and the filter is set to the URL param every time the component did mount. So it doesn't honor the user decisions when they navigate to another tab of the same survey.

[Here](https://github.com/instedd/surveda/blob/ed09a2778c417d60e74ce7887514bb3b023a3600/web/static/js/reducers/respondents.js#L28-L45) we handle the user decisions regarding `items` (the respondent rows) and sorting from the `respondents` reducer. 
 
### 🛣️ Exploring solutions

I drafted a solution. It's not necessary THE solution. It's just one I've found on my way. I'd bet money that there's a better one.

Basically, I removed the control of when and why the state is preserved or not from the reducer. I moved this responsibility to `RespondentIndex`. I think the principal reason to do that is the filter deep linking. We reached a workaround for handling its state when we added the feature, but it's not totally working: it doesn't honor the user decisions when they navigate to another tab of the same survey.

I think [this way](https://github.com/instedd/surveda/blob/805b6ba32470bf496f4767c6697bbc823ad43c07/web/static/js/components/respondents/RespondentIndex.jsx#L94-L106) is clearer (and less error-prone) than the one we'd found in our first iteration of the feature.

PR #1762, opened 2020-08-06
Add partial relevant answered counter to the respondents index and the results csv

### LINKED ISSUE #1753: Add partial relevant answered counter to the respondents index and the results csv
Add a variable to the results file that contains the number of partial relevant questions that a respondent answered

aeatencio:
This is how it looks:

Respondents index

CSV results

Fix #1753

### DIFF
--- a/web/controllers/respondent_controller.ex
+++ b/web/controllers/respondent_controller.ex
@@ -49,11 +49,7 @@ defmodule Ask.RespondentController do
 
     survey = Repo.get!(Survey, survey_id)
 
-    # TODO: Check the following question:
-    # We aren't considering the questionnaire of every respondent of the survey.
-    # We're considering only the questionnaires of the filtered respondents.
-    # Should we take in consideration the questionnaire of every respondent of the survey or this is enough?
-    partial_relevant_enabled = Respondent.any_partial_relevant_in_respondents?(respondents)
+    partial_relevant_enabled = partial_relevant_enabled_in_survey?(survey_id)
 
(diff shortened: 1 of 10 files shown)


web/controllers/respondent_controller.ex
```diff
@@ -85,6 +87,37 @@ defmodule Ask.RespondentController do
       |> present_fields("response")
   end
 
+  defp partial_relevant_fields(respondents) do
+    if has_any_respondent_any_partial_relevant_question?(respondents) do
```
bcardiff, line 91: whether to include the partial relevant field or not should depend only on the configuration of the quex. 
* It does not matter if there are or not respondents with relevant questions answered.
* It does not matter how many relevant questions the quex has

Only if the quex toggle is selected.

aeatencio: "It does not matter if there are or not respondents with relevant questions answered" -> I had this in mind.
"It does not matter how many relevant questions the quex has" -> OK, I see that if the toggle is ON, we expect that the questionnaire has at least one relevant question. I'll change that.

bcardiff: If some quex of the survey has the toggle of partial relevant, then the field should appear on all respondents of the survey.

When computing the field value for the respondent, use the quex to determine if/which are the partial relevant questions.

aeatencio: OK, I understand the intention. I'll review the code to see what change is needed to reach that. Thanks!!

PR #1767, opened 2020-08-27
Let the user select every field of the survey

### LINKED ISSUE #1764: Compute fields in the column picker using the Survery's questionnaires
Compute fields in the column picker using the Survey's questionnaires instead of the responses of the first page of respondents

Follow up #1745

aeatencio:
By computing the fields using the Survey's questionnaires
Instead of the responses of the first request respondents

Fix #1764

### DIFF
--- a/test/controllers/respondent_controller_test.exs
+++ b/test/controllers/respondent_controller_test.exs
@@ -269,8 +269,17 @@ defmodule Ask.RespondentControllerTest do
 
       body = json_response(conn, 200)
       fields = body["meta"]["fields"]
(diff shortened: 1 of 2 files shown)


manumoreira:
> @manumoreira there was no decision so far on how to list the fields in the UI. In the scenario where the survey has multiple quex, how should the fields of the questions be shown? Alphabetically? Or we should honor the order one of the quex to relate them better to the timeline if possible.

I'd suggest Alphabetically, I can't find a simple way to choose which quex to honor

Issue #1772, opened 2020-09-02
Page size selector works only on respondents index


mmuller:
Found in version c19b25c

Make page size work on activity, project, quex, channel listing

- Check if it’s possible to add pageSize to activity, project, quex, channel or if pageSize control should be disabled.
- Check which other listings are in the app

When using the page size selector the following error is thrown:

```
console.js:26 ErrorEvent {isTrusted: true, message: "Uncaught TypeError: t is not a function", filename: "https://surveda-stg.instedd.org/js/app-32b5502a0bb06321b4d84f88416d9c12.js?vsn=d", lineno: 96, colno: 454085, …}
```

mmuller:
Working properly in version ad2b6c3

PR #1774, opened 2020-09-03
Fix page size selection on activities index

aeatencio:
And temporarily hide the page size selection in the pages where it's not implemented yet

Also, improve the activities fetch redux design:

 - Copy the pattern from recent respondents index refactor
 - Set the request variables on fetch start
 - Discard the response if it's not for the last request

For #1772


bcardiff (approved):

bcardiff:
On a future iteration:

- It would be good to check why the collection reducer was not used in neither activites nor respondents.
- Could collection be used as a base abstraction for all the listings in the app?
- Why fetch reducer and collection do not have an action counterpart to reduce the boilerplate there.
- This might push some changes in the client-side code to invoke the api to standarize how to pass the listings query strings probably and infer the scope from it.

PR #1778, opened 2020-09-11
Archive questionnaires

### LINKED ISSUE #1771: Archive questionnaires
- [x] In quex index by able to select active or archived quex (front-end  / back-end)
- [x] In the quex list allow the user to archive / unarchive (front-end / back-end)
- [x] Show archived questionnaires as read only in questionnaire editor (front-end)
- [x] Don't update archived questionnaires, they aren't updatable (back-end)
- [x] Don't archive questionnaires of not running surveys (back-end)
- [x] Show the user a message when trying to archive a not running survey quex (front-end)
- [x] Duplicating an archived quex should create an active quex
- [x] When creating a survey list only the active quex
- [x] Log archive and unarchive events in activities

aeatencio:
This is how it looks:

Archive questionnaires

Fix #1771


test/controllers/project_controller_test.exs
```diff
@@ -258,17 +258,30 @@ defmodule Ask.ProjectControllerTest do
 
     test "sets archived status to true", %{conn: conn, user: user} do
       project = create_project_for_user(user)
-      put conn, project_update_archived_status_path(conn, :update_archived_status, project), project: %{"archived" => "true"}
+      put conn, project_update_archived_status_path(conn, :update_archived_status, project), project: %{"archived" => true}
```
bcardiff, line 261: why this is changing from `"true"` to `true`? but on line 289 that is not done, and in questionnaire_controller_test.exs the string alternativa is also used?

aeatencio: Here (and there) I just improved the code to make it test the same data type the application handles in real time.

bcardiff: But the controllers handle http request query string and post body. There are no booleans there, everything is a string `project_id=2&archived=true`.

aeatencio: The body sent in the HTTP request looks like `{project: {archived: true, ... }}`
The function that handles the request extracts `project_params` [here](https://github.com/instedd/surveda/blob/9af2526542c0f1d1fdd1a038c961d39141eee8fd/web/controllers/project_controller.ex#L135). It looks like `%{"archived" => true, ...}`

The test wasn't representing the real scenario where this function is being called.

bcardiff (approved):

PR #1781, opened 2020-09-17
Ensure attempts are at least every 10m

### LINKED ISSUE #570: Issue when re-contact attempts are 1m
@waj said I should log this, so I must.

aeatencio:
This is how it looks:

Retries and fallback configuration

Fix #570


web/static/js/reducers/survey.js
```diff
@@ -123,7 +123,8 @@ const validateFallbackDelay = (state: DataStore<Survey>) => {
   }
 }
 
-const timeSpecRegex = /^\d+[mhd]$/
+// Attempts are at least every 10m
+const timeSpecRegex = /^([1-9]\d+m)|([1-9]\d*[hd])$/
```
bcardiff, line 127: This is abusing a bit the `10` value.

Refactoring the `timeSpecRegex.test` calls to use Regex for matching  `/^\d+[mhd]$/` and then `parseInt` to extract the value of the `\d+` using match data would be less obscure.

bcardiff (changes requested):

web/static/js/reducers/survey.js
aeatencio: I've just improved the code expressiveness. Please take a look now and let me know. Thanks!

bcardiff:
The following is something that uses regex to extract the information of quantity and unit, yet the validation itself is expressed in simpler terms.

```js
function isValid(str) {
  const m = str.match(/^(\d+)([hdm])$/)
  if (m == null) return false
  const quantity = parseInt(m[1])
  const unit = m[2]
  if (quantity == 0) return false
  if (quantity < 10 && unit == "m") return false
  return true
}
```

If we need to limit differently the quantity this code is easier to adapt and understand.

web/static/js/components/surveys/SurveyWizardRetryAttempts.jsx
```diff
@@ -177,7 +177,7 @@ class SurveyWizardRetryAttempts extends Component {
                 />
               </InputWithLabel>
               <span className='small-text-bellow'>
-                {t('Enter a delay like 5m, 3h or 1d to express a time unit')}
+                {t('Enter delays like 10m 2h 1d to express time units (default 1h, use values greater than 10m)')}
```
bcardiff, line 180: The fallback delay is one single delay. That's why it was phrased differently.

bcardiff (changes requested):

web/static/js/components/surveys/SurveyWizardRetryAttempts.jsx
aeatencio: True! Thanks!

bcardiff (approved):

PR #1783, opened 2020-09-21
Show no surveys message when the project is empty

### LINKED ISSUE #1780: Missing message in empty surveys index
When a Project doesn't have Surveys the message "You have no surveys in this project" must be displayed.
In version ad2b6c3 that message is missing.

image

aeatencio:
This is how it looks:

Empty project

Fix #1780

### DIFF
--- a/web/static/js/actions/folder.js
+++ b/web/static/js/actions/folder.js
@@ -105,7 +105,18 @@ export const fetchFolders = (projectId: number) => (dispatch: Function) => {
 
   return api.fetchFolders(projectId)
     .then(response => {
-      dispatch(fetchedFolders(projectId, response.entities.folders))
+      /*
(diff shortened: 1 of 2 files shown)


bcardiff (changes requested):
The fix does not seems the right thing to do.

Why is the `dispatch(folderActions.fetchFolders(projectId))` in the `initialFetch` leaving an `undefined` value in `state.folder.folders`? After the loadingFolders is set to `false` the state.folder.folders should have a map of the folders object.

The `|| []` is a workaround to something else that is not working as expected. Let's try to find that out instead.

aeatencio:
@bcardiff please take a look now. Thanks!

bcardiff (approved):

PR #1787, opened 2020-10-03
Show invitation link on mobileweb simulator

### LINKED ISSUE #1785: Mobile web simulator
Currently, if a user wants to test the mobile web mode needs to have a QST client nuntium channel to send (and receive) the SMS invitation with the link.
It will be useful to implement a mobile web simulator that simply exposes the SMS invitation with a link in the sandbox.

aeatencio:
This is how it looks:

Mobile web simulator (waiting)

Mobile web simulator

Mobile web simulator (iframe)

Fix #1785

At some point, this was the pending work:
- [x] Measure how long the respondent session isn't available.
  - [x] Let the user know they have to wait 
- [x] Avoid 404 if possible when calling `simulation_initial_state`
- [x] Verify tags escaping by `linkifyjs`
- [x] Remove `mobile_web_url` from `simulation_initial_state`
- [x] Tune `stepComponent` render in `SurveySimulation.jsx`
- [x] Show Mobile web page inside an iframe
  - [x] Include a left arrow in the title to let the user back to the SMS phone view
- [x] Unit tests for the new endpoint


web/models/respondent.ex
```diff
@@ -293,4 +293,9 @@ defmodule Ask.Respondent do
     not_ignored = fn value -> String.upcase(value) not in ignored_values end
     quiz_step["relevant"] && not_ignored.(response.value)
   end
+
+  def mobile_web_url(respondent_id) do
```
anaPerezGhiglia, line 297: It looks weird for me that this function is in the `Respondent` module. 
Perhaps belongs somewhere like `web/controllers/controller_helper` ?

anaPerezGhiglia (commented):
I still haven't finished reviewing all the changes, but calling it a day for today 🙂

web/models/respondent.ex
aeatencio: I moved it back to its original place. I extracted it because I planned to expose it from the API but it didn't finally happen. So there's no need to place it in the Respondent (or any other) module. Thanks!

web/static/js/components/surveys/SurveySimulation.jsx
```diff
@@ -212,8 +242,76 @@ class SurveySimulation extends Component {
...
+    const { mode, t } = this.props
+    const { state, initialState } = this.state
+    if (mode != 'mobileweb') return []
+    const respondentIsActive = this.respondentIsActive()
+    const respondentIsPendingOrActive = state == 'pending' || respondentIsActive
+    const initialStateLoaded = this.initialStateLoaded()
+
+    let messages = []
+    if (respondentIsPendingOrActive && !initialStateLoaded) {
+      messages = [t('Waiting for messages to be sent. This could take up to 1 minute.')]
```
bcardiff, line 255: These status messages are displayed as message bubbles? If so, I think that is a bit confusing.

aeatencio: They aren't displayed exactly as bubbles. This is how it looks:

Mobile web simulator (waiting)

But yes, they're similar. The styling rules are [these](https://github.com/instedd/surveda/blob/8325572a92ba0f3b5d2ed336773452905dee318a/web/static/css/_quex-simulation.scss#L227-L234).

bcardiff: Display the text centered on the mobile. No bubble. Probably the text should be dark due to the "light" background. That should be enough.

aeatencio: This is how it looks now:

Mobile web simulator (waiting)

bcardiff (approved):

PR #1792, opened 2020-10-15
Limit survey simulation endpoints

### LINKED ISSUE #1791: Survey simulation endpoints don't limit by simulation flag
Recently, #1787 tried to fix it but it didn't.

Also, unit tests are missing.

aeatencio:
Before #1787, any survey simulation API endpoint wasn't limiting the actions to survey simulations only.

In #1787 we tried to add this restriction: real surveys can't be stopped, or any other action shouldn't apply to them from the simulation endpoints.

But the solution didn't work. Also, we didn't test it.

So, in this PR, we made these restrictions work and added the corresponding tests.

Fix #1791

### DIFF
--- a/test/controllers/survey_controller_test.exs
+++ b/test/controllers/survey_controller_test.exs
@@ -577,7 +577,7 @@ defmodule Ask.SurveyControllerTest do
   describe "stop simulation" do
     setup %{conn: conn, user: user} do
       create_running_survey = fn simulation ->
-        [survey, _group, _test_channel, _respondent, _phone_number] =
+        [survey | _tail] =
           create_running_survey_with_channel_and_respondent_with_options(
             user: user,
             simulation: simulation
@@ -617,7 +617,7 @@ defmodule Ask.SurveyControllerTest do
   describe "simulation status" do
     setup %{conn: conn, user: user} do
       create_running_survey = fn simulation ->
-        [survey, _group, _test_channel, _respondent, _phone_number] =
+        [survey | _tail] =
           create_running_survey_with_channel_and_respondent_with_options(
             user: user,
             simulation: simulation
@@ -705,7 +705,7 @@ defmodule Ask.SurveyControllerTest do
       create_running_survey: create_running_survey,
       get_simulation_initial_state: get_simulation_initial_state
     } do
-      [mode, simulation] = ["sms", false]
+      {mode, simulation} = {"sms", false}
       %{survey: survey} = create_running_survey.(mode, simulation)
 
       assert_error_sent(:not_found, fn ->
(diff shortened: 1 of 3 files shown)


test/controllers/survey_controller_test.exs
```diff
@@ -574,11 +574,96 @@ defmodule Ask.SurveyControllerTest do
...
+      assert_error_sent(:not_found, fn ->
+        stop_simulation.(survey)
+      end)
+    end
+  end
+
+  describe "simulation status" do
+    setup %{conn: conn, user: user} do
+      create_running_survey = fn simulation ->
+        [survey, _group, _test_channel, _respondent, _phone_number] =
```
anaPerezGhiglia, line 620: same as above

test/controllers/survey_controller_test.exs
```diff
@@ -616,12 +701,24 @@ defmodule Ask.SurveyControllerTest do
...
+      end)
+    end
+
     test "SMS return an empty map", %{
       create_running_survey: create_running_survey,
       get_simulation_initial_state: get_simulation_initial_state
     } do
-      mode = "sms"
-      %{survey: survey} = create_running_survey.(mode)
+      [mode, simulation] = ["sms", true]
```
anaPerezGhiglia, line 720: I may be a little rusted with elixir, but this is the same as
```elixir
mode = "sms"
simulation = true
```

right?

anaPerezGhiglia (approved):
LGTM 🙂 
left a few comments but are not mandatory for merging

test/controllers/survey_controller_test.exs
aeatencio: Yes, I think it's exactly the same. It's just a matter of style. May I be missing some details regarding performance, or something else?

test/controllers/survey_controller_test.exs
```diff
@@ -574,11 +574,96 @@ defmodule Ask.SurveyControllerTest do
     end
   end
 
+  describe "stop simulation" do
+    setup %{conn: conn, user: user} do
+      create_running_survey = fn simulation ->
+        [survey, _group, _test_channel, _respondent, _phone_number] =
```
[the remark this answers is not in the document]
aeatencio: Yes, I haven't considered this alternative while programming. And I don't have strong feelings about one of these two options I prefer. This is why I'll take your advice 😉 Thanks!

test/controllers/survey_controller_test.exs
anaPerezGhiglia: No that I can see 
I was just checking that was the same 🙂

aeatencio: [Here](https://elixir-lang.org/getting-started/pattern-matching.html#pattern-matching) there are some similar examples:

```elixir
iex> {a, b, c} = {:hello, "world", 42}
{:hello, "world", 42}
iex> a
:hello
iex> b
"world"
```

```elixir
iex> [a, b, c] = [1, 2, 3]
[1, 2, 3]
iex> a
1
```

Maybe using a tuple (instead of a list) is more expressive?

anaPerezGhiglia: for me the simplest way is to bind the two variables separately, but choose the one that best suits your taste 🙂

aeatencio: So, after taking the advice of my gut (and, ok, also of @bcardiff 😬) tuples they are 🎉

PR #1806, opened 2020-11-25
Create panel surveys

### LINKED ISSUE #1802: Allow surveys to be repeated when creatinig a survey
- Add a Repeat Survey screen when creating surveys https://projects.invisionapp.com/d/main#/console/8374235/437343078/preview
- Add a nilable panel_survey_group_id to group all surveys of the same repetition
- Add a boolean to show if a survey is the latest occurrence of the group.
- When creating a new survey that can be repeated, this will point to self and it should be flagged as the latest occurrence of the group

aeatencio:
### Back-end:

- [x] Add panel survey fields to survey model (including migration)
- [x] Handle the "is panel survey" flag in modification endpoint (including tests)

### Front-end:

- [x] Let the end-user set up panel surveys

## 📷 How it looks

Panel survey

---

After a questionnaire is selected, the panel survey step is marked as completed ✅ (mimic the Cutoff rules step)

Panel survey completed

Fix #1802


web/router.ex
```diff
@@ -70,8 +70,10 @@ defmodule Ask.Router do
 
       get "/timezones", TimezoneController, :timezones
       resources "/projects", ProjectController, except: [:new, :edit] do
+        post "/panel_survey", SurveyController, :create_panel_survey
         resources "/folders", FolderController, only: [:create, :index, :delete] do
           post "/set_name", FolderController, :set_name
+          post "/panel_survey", SurveyController, :create_panel_survey
```
bcardiff, line 76: I would not expect to have a separate post to create panel surveys. I do expect a post to create a new incarnation of a panel survey. But that is another story.

web/controllers/survey_controller.ex
```diff
@@ -41,9 +41,9 @@ defmodule Ask.SurveyController do
     render(conn, "index.json", surveys: surveys)
   end
 
-  def create(conn, params = %{"folder_id" => folder_id}), do: create(conn, params, folder_id)
+  def create(conn, params = %{"project_id" => project_id}) do
```
aeatencio, line 44: This is a minor code improvement. It comes from [this commit](https://github.com/instedd/surveda/commit/d9cce3b425ccca2f38be091f68f3d9e819f11eae). It removes an unneeded function overload.

bcardiff (approved):

Issue #1805, opened 2020-11-20
Add panel surveys


aeatencio:
Here, we're defining a Surveda **panel survey** as a periodically repeated survey.

1. **Regular surveys** support for _cross-sectional studies_:
   - Observations at a single point in time
   - Observes different groups (a “cross-section”) in the population
   - Provides a snapshot of the society at a given point

2. **Panel surveys** support _longitudinal studies_:
   - Repeated observations
   - Observes the same group multiple times
   - Follows changes in participants over time

Roughly, this is how panel surveys will be implemented:
 
- Survey that repeats itself
- Each repetition is a new survey
- Each survey will be named based on the date (YYYY-MM-DD) on which it was launched
- When a new repetition survey is launched the previous survey will be canceled
  - A warning message detailing the impact of the cancellation process might be needed
- Data from all the repetitions will be merged into one dataset
  - Each respondent will be uniquely identifiable across the whole panel survey (the respondent id will be the same for all panel survey incarnations)

Issue #1804, opened 2020-11-19
Show an option to start a new occurrence of a panel survey


bcardiff:
- https://projects.invisionapp.com/d/main#/console/8374235/286487531/comments
- After creation it should go to the new survey overview

aeatencio:
Epic: #1805

bcardiff:
Ref: https://github.com/instedd/surveda/issues/1803#issuecomment-731474378

The buttons should appear disabled if the survey is still ongoing, and show a caption that that is the reason.

aeatencio:
![Repeat survey button](https://user-images.githubusercontent.com/39921597/100449288-6bfc7e00-3092-11eb-94a9-6dcba1d3bdff.png)

Issue #1803, opened 2020-11-19
Allow creation of a new ocurrence of a panel survey via API


bcardiff:
- It should NOT cancel the latest occurrence if it's still running
- It should create a new survey within the same panel_survey_group_id
- It should flag the new survey as the latest occurrence
- It starts the new survey
- It should validate that the survey has a panel_survey_group_id
- Respondent ids should be preserved along the group
  - The encryption algorithm should take into account the panel_survey_group_id when it's != nil

manumoreira:
Small clarification: the creation of a new occurrence will not be responsible for cancelling the latest occurrence, but it should be completed or cancelled to get be able to be repeated.

bcardiff:
- discuss: which respondents should be promoted to the new repetition?
    - all but refused respondents should be promoted
- discuss: how to handle respondent group and multiple channels?
    - s1: 
        - rg1: 
            - jdoe channel1/channel2
        - rg2: 
            - sdoe channel1/channel2
        - rg3: 
            - rdoe channel3
      - s2:
        - rg-channel1-channel2:
            - jdoe channel1/channel2
            - sdoe channel1/channel2
        - rg-channel3
          - rdoe channel3
    - In the new incarnation there should be a rgX for each channel combination needed.
    - respondents should use the same channel
    - respondents will start from the beginning even if the fallback channel was reached.
- discuss: how does quex comparissons work?
    - not supported for now
    - should be disable that from the UI
- discuss: how does mode comparissons work?
    - not supported for now
    - should be disable that from the UI
- discuss: what happen if the channel is removed before creating the incarnation.
    - pending: to be design a replace screen

mmuller:
Working properly in version 0.30.0

PR #1811, opened 2020-12-02
Repeat panel survey (API)

aeatencio:
### 🎯 The goal

This PR adds a new endpoint to create a new occurrence of a panel survey.

For: #1803
See: #1805

### 🚚 The refactors

Because repeating a survey includes starting it, creating respondent groups, and inserting respondents, some refactors were done.

1. Most of the logic for starting a survey was moved from the survey controller to a new module `Ask.Runtime.SurveyAction`. From now, this module is in charge of keeping the logic for repeating and starting a survey.
2. Most of the logic for creating respondent groups and inserting respondents was moved from the respondent group controller to a new module `Ask.Runtime.RespondentGroup`

### 🚧 The new logic

Repeating a survey has its own (new) logic. The top of the logic is in `Ask.Runtime.SurveyAction` and the down is in `Ask.Runtime.PanelSurvey` (also, a new module).

### ✅ Expected behavior
 - [x] Endpoint: `POST /api/v1/projects/1/surveys/300/repeat`
 - [x] It only repeats the latest occurrence of a panel survey, and only after it succeeded.
 - [x] It creates a new survey within the same panel_survey_group_id.
 - [x] It flags the new survey as the latest occurrence.
 - [x] It starts the new survey.
 - [x] It validates that the survey has a panel_survey_group_id
 - [x] It preserves the respondent id through the group

### 🚫 Excluded from this PR
 - [ ] Improve the respondent group creation logic: excluded from this iteration. For now, for each existing group, a new respondent group with the same name is created. Each respondent group has a copy of every respondent (except refused) and channel association.
 - [ ] Add regression tests: included in this iteration, but after merging this PR and doing #1804.

### DIFF
--- a/web/controllers/survey_controller.ex
+++ b/web/controllers/survey_controller.ex
@@ -348,7 +348,6 @@ defmodule Ask.SurveyController do
       render(conn, "show.json",
         survey:
           survey
-          |> Repo.preload(:questionnaires)
           |> Survey.with_links(user_level(survey.project_id, current_user(conn).id))
       )
 
--- a/web/controllers/survey_controller.ex
+++ b/web/controllers/survey_controller.ex
@@ -18,6 +18,7 @@ defmodule Ask.SurveyController do
     dynamic =
       if params["state"] do
         if params["state"] == "completed" do
+          # Same as Survey.succeeded?(s)
           dynamic([s], s.state == "terminated" and s.exit_code == 0 and ^dynamic)
         else
           dynamic([s], s.state == ^params["state"] and ^dynamic)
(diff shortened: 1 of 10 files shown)


web/controllers/survey_controller.ex
```diff
@@ -359,6 +360,69 @@ defmodule Ask.SurveyController do
     end
   end
 
+  def repeat(conn, %{"survey_id" => id}) do
```
bcardiff, line 363: I would expect that repeat would call launch instead of duplicating the logic.

repeat should be

1. create new ocurrence
2. setup respondents
3. launch

PR #1819, opened 2020-12-21
Mobileweb simulation

### LINKED ISSUE #1808: Create simulation channel for mobile web
To behave as a simulation that can be used by any surveda end user we should create a simulated channel for mobile web mode.
The main action of this channel is to send the initial message and the URL for the web form.
We should find a way to create this in the context of the sandbox without the necessity of having a real nuntium channel.

aeatencio:
## 🥕 The motivation

When editing a Survey, having a quick demo in every mode is very useful.
The easiest way of having a demo is the questionnaire simulator because channels aren't needed.
But the mobile web mode isn't supported by the questionnaire simulator yet.

Fix #1808

## 🎯 The goal

This PR makes the questionnaire simulator support the Mobile Web mode.

## 🗒️ The To-Do list:

- [x] Support explanations in Mobile Web mode.
- [x] Split the simulator into the SMS and the Mobile Web simulator. 
- [x] Remove phone from survey simulation screen.
- [x] Restrict endpoints to SMS or Mobile Web only.
- [x] Fix the front-end normalize error to use GET in the new `/simulation/:respondent_id` endpoint.
- [x] Fix missing question in SMS mode.
- [x] Add tests.

### DIFF
--- a/lib/ask/runtime/questionnaire_simulator.ex
+++ b/lib/ask/runtime/questionnaire_simulator.ex
@@ -258,8 +258,12 @@ defmodule Ask.Runtime.QuestionnaireMobileWebSimulator do
   def get_last_simulation_response(respondent_id) do
     simulation = QuestionnaireSimulatorStore.get_respondent_simulation(respondent_id)
     if (simulation) do
-      %{last_simulation_response: last_simulation_response} = simulation
-      last_simulation_response
+      last_simulation_response = Map.get(simulation, :last_simulation_response)
+      if (last_simulation_response) do
+        last_simulation_response
+      else
+        Response.invalid_simulation
+      end
     else
       respondent_id
       |> QuestionnaireSimulationStep.expired
--- a/lib/ask/runtime/questionnaire_simulator.ex
+++ b/lib/ask/runtime/questionnaire_simulator.ex
@@ -10,8 +10,10 @@ defmodule Ask.QuestionnaireSimulationStep do
   alias Ask.Simulation.Status
   alias __MODULE__
 
+  defstruct [:respondent_id, :simulation_status, :disposition, :current_step, :questionnaire, submissions: []]
+
   def expired(respondent_id) do
-    %{respondent_id: respondent_id, simulation_status: Status.expired}
+    %QuestionnaireSimulationStep{respondent_id: respondent_id, simulation_status: Status.expired}
   end
 
   def build(%{respondent: respondent, submissions: submissions, questionnaire: quiz, section_order: section_order}, current_step, status, with_quiz) do
--- a/lib/ask/runtime/questionnaire_simulator.ex
+++ b/lib/ask/runtime/questionnaire_simulator.ex
@@ -45,6 +45,7 @@ defmodule Ask.Runtime.QuestionnaireSimulator do
   alias Ask.Runtime.{Session, Flow, QuestionnaireSimulatorStore}
 
   @sms_mode "sms"
+  @mobileweb_mode "mobileweb"
 
   def start_simulation(project, questionnaire, mode \\ @sms_mode) do
     if valid_questionnaire?(questionnaire, mode) do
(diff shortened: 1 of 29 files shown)


lib/ask/runtime/questionnaire_simulator.ex
```diff
@@ -139,12 +169,16 @@ defmodule Ask.Runtime.QuestionnaireSimulator do
...
   defp handle_app_reply(simulation, respondent, reply, status) do
     messages = simulation.messages ++ AOMessage.create_all(reply)
+    # TODO: support explanations in mobileweb
     submitted_steps = simulation.submissions ++ SubmittedStep.build_from_responses(respondent, simulation) ++ SubmittedStep.new_explanations(reply)
     current_step = current_step(reply)
 
-    QuestionnaireSimulatorStore.add_respondent_simulation(respondent.id, %Ask.QuestionnaireSimulation{simulation | respondent: sync_respondent(respondent), messages: messages, submissions: submitted_steps})
-    |> QuestionnaireSimulationStep.sync_build(current_step, status)
-    |> Response.success
+    simulation = %{simulation | respondent: sync_respondent(respondent), messages: messages, submissions: submitted_steps}
```
anaPerezGhiglia, line 176: I'm rusty with Elixir, but shouldn't be "safer" this struct to be `%Ask.QuestionnaireSimulation{simulation | ...}` as it was before?

lib/ask/runtime/session.ex
```diff
@@ -321,13 +322,12 @@ defmodule Ask.Runtime.Session do
...
-        {:end, reply, respondent}
-      {:ok, flow, reply} ->
-        {:ok, %{session | flow: flow, respondent: respondent}, reply, current_timeout(session)}
-    end
+  defp mode_start(%Session{current_mode: %SMSSimulatorMode{}} = session) do
+    mode_start_simulation(session)
+  end
+
+  defp mode_start(%Session{current_mode: %MobilewebSimulatorMode{}} = session) do
+    mode_start_simulation(session)
```
anaPerezGhiglia, line 330: 🤔  I'm not being able to see the benefit of having this `mode_start` with two different signatures if both are calling the same function

test/lib/runtime/questionnaire_simulator_test.exs
```diff
@@ -84,81 +172,134 @@ defmodule QuestionnaireSimulatorTest do
...
       ]
-      assert expected_submissions == submissions
+
+      Enum.map([true, false], fn nil_thank_you_message ->
+        quiz = questionnaire_with_steps(steps, nil_thank_you_message: nil_thank_you_message)
+        Enum.map(["sms", "mobileweb"], fn mode ->
+          responses = case mode do
+            "sms" ->
+              ["1", "Y", "7", "4"]
+              "mobileweb" ->
```
bcardiff, line 263: format

test/lib/runtime/questionnaire_simulator_test.exs
```diff
@@ -354,6 +520,28 @@ defmodule QuestionnaireSimulatorTest do
...
+    simulation = if mode == "mobileweb" do
+      %{respondent_id: respondent_id} = simulation
+      mobileweb_user_consented(respondent_id)
+    else
+      simulation
+    end
+    simulation
+  end
+
+  defp assert_last_message(simulation, message, mode) do
```
bcardiff, line 539:
```suggestion
  defp on_sms_assert_last_message(simulation, message, mode) do
```

So we know when reading the test that the assertion applies only on sms mode

web/views/questionnaire_view.ex
```diff
@@ -29,15 +29,24 @@ defmodule Ask.QuestionnaireView do
...
+  defp render_simulation(simulation) do
     %{
       respondent_id: simulation.respondent_id,
       simulation_status: simulation.simulation_status,
-      disposition: simulation.disposition,
-      messages_history: simulation.messages_history,
-      submissions: simulation.submissions,
-      current_step: simulation.current_step,
-      questionnaire: render("questionnaire.json", %{questionnaire: simulation.questionnaire})
+      disposition: Map.get(simulation, :disposition),
```
bcardiff, line 45: Why are these changes needed?

aeatencio: I think they were needed because I had removed the `QuestionnaireSimulationStep` struct.

Please take a look at a02f945d. I think it's a better approach.

bcardiff (approved):

PR #1822, opened 2021-01-07
Delete panel surveys

### LINKED ISSUE #1821: Panel surveys cannot be removed
When trying to remove a panel survey the following error is thrown:

`(Ecto.ConstraintError) constraint error when attempting to delete struct: * foreign_key: surveys_panel_survey_of_fkey`

aeatencio:
### Expected Behaviour:
- [x] If it's the only one, just drop it
- [x] Removing one in the middle is fine
- [x] Removing the original should make the second one to act as the original
- [x] Removing the last one should allow the user to create a new incarnation from the previous one from the normal flow.

Fix #1821

### DIFF
--- a/web/controllers/survey_controller.ex
+++ b/web/controllers/survey_controller.ex
@@ -2,7 +2,7 @@ defmodule Ask.SurveyController do
   use Ask.Web, :api_controller
 
   alias Ask.{Project, Folder, Survey, Questionnaire, Logger, RespondentGroup, Respondent, Channel, ShortLink, ActivityLog, RetriesHistogram}
-  alias Ask.Runtime.Session
+  alias Ask.Runtime.{Session, SurveyAction}
   alias Ecto.Multi
   alias Ask.SurveyCanceller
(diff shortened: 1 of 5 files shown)


web/controllers/survey_controller.ex
```diff
@@ -288,14 +288,10 @@ defmodule Ask.SurveyController do
...
-        multi = Multi.new
-        |> Multi.delete(:survey, survey)
-        |> Multi.insert(:log, ActivityLog.delete_survey(project, conn, survey))
-        |> Repo.transaction
-
-        case multi do
+        case SurveyAction.delete(survey) do
           {:ok, _} ->
             project |> Project.touch!
+            ActivityLog.delete_survey(project, conn, survey) |> Repo.insert!
```
bcardiff, line 294: Why move out the activity log insertion outside the transaction?
I would expect this to happen inside the SurveyAction.delete

bcardiff (approved):

PR #1827, opened 2021-01-13
Bring surveda-d3-components and add support of i18n

### LINKED ISSUE #1807: Merge surveda-d3-components and add support of i18n

aeatencio:
### The motivations 🥕 

1. We need to complete the cockpit translations, and doing it is easier with its React components being part of the front-end app.
2. Having these components in a different repo brings no big pros but several cons in our daily work (like unnecessarily duplicated deploys).

### The works 🚧 

- [x] Copy-paste the React components from [their last version](instedd/surveda-d3-components@862f6f0)
- [x] Fix linter errors 
- [x] Add the missing translations to the cockpit
- [x] Upgrade d3 version 
- [ ] ~Add Flow~ (no need for now)

This fixes #1807 

### The technical debt 🐌 

The copy-pasted components don't support [Flow](https://flow.org/), and adding this support isn't a trivial task.
We should add a static type checker to them in the future.

### How it looks 📷 

These changes shouldn't bring significant complications, and the cockpit look and feel shouldn't change. They were smoke tested locally, and they look just fine.

running survey cookpit

The survey card used a previous version of d3. Its code was upgraded to avoid using 2 different versions of d3 in the same project

survey card


bcardiff (commented):
There is no need to add flow for now to the charts.

Rename _d3.scss as _charts.scss and also the component/d3 directory to component/charts.

Other than that looks good to go.

aeatencio:
> There is no need to add flow for now to the charts.

OK, I agree

> Rename _d3.scss as _charts.scss and also the component/d3 directory to component/charts.

Sure, I agree. It's a better name. 

> Other than that looks good to go.

So we're good to go now! 🎉

bcardiff (approved):

PR #1829, opened 2021-01-14
Internationalize texts in SCSS

### LINKED ISSUE #1809: i18n text in _forms.scss
_form.cscc contains lots of text that would be good to get internationalized
https://github.com/instedd/surveda/blob/master/web/static/css/_forms.scss

aeatencio:
### Motivation 🥕 

Internationalize every end-user message.

### Current internalization state 🏁

#1827 and this PR should finish or so the internationalization

### How does it look 📷 

No style should change. The CSS rules were moved to be dynamically written with the same i18n used in every end-user message.

I didn't test that every rule applies OK. I don't expect to have problems with them because they were pretty mechanically copy-pasted. The following translation was made for testing purposes only and wasn't included in this PR.

new i18n

Fix #1809


web/static/css/_forms.scss
```diff
@@ -60,42 +60,29 @@ label {
     top: 50%;
     margin: 0 15%;
     &.csv:before {
-      content: 'Drop your CSV file here, or click to browse';
```
bcardiff, line 63: Let's leave the content: '...' commented in the css and a note to see I18nStyle.jsx

bcardiff (commented):
Looks good. Did you check that going back and forth between different language work and the styles expanded do not accumulate on the client side, right?

aeatencio:
> Looks good. Did you check that going back and forth between different language work and the styles expanded do not accumulate on the client side, right?

Yes, they don't accumulate and they work fine going back and forth between different languages.

They're rendered as a regular React element, this is why they don't accumulate.
And this is also why they render inside the `<body>`, which is rare, but... 🤷 

i18n style element

web/static/css/_forms.scss
aeatencio: OK! Done!

bcardiff (approved):

PR #1832, opened 2021-02-02
Export only the visible settings of the questionnaire (languages)

### LINKED ISSUE #1801: Audios and text data from deleted languages are included in the quex export
Found in version 567ce25
**Preconditions:**

- There is a Questionnaire Q1.
- Q1 has enabled the Phone call mode.
- Q1 has a Step S.
- Q1 has more than one language
- There is a Questionnaire Q2 (a copy of Q1 that has only one language)

**Steps to reproduce:**

- Navigate to the Questionnaire Q1.
- Add audio files to all the steps in all languages
- Export Q1.
- Create a copy of Q1 called Q2
- Export Q2.
- Observe the exported files.
- Obtained result:
- The audio files of the deleted languages are included in Q2

**Examples**

- Multi language https://surveda.instedd.org/projects/355/questionnaires/1091/edit
- One language (copy of Multilanguage and deleted language) https://surveda.instedd.org/projects/355/questionnaires/1092/edit

-

aeatencio:
### The context 🌌 

When some settings are deleted/disabled (languages, modes, partial relevant flag), the associated settings are still there (they become visible after the deleted/disabled setting is back). But they aren't visible to the end-user.

### The symptom 🌡️ 

After exporting and importing the questionnaire, the end-user doesn't expect these settings to be still there.

### The fix 🔧 

This PR avoid exporting the settings of every deleted language.

### The pending work 🚧 

Avoid exporting any other invisible setting to the end-user.

---

It fixes #1801

### DIFF
--- a/lib/ask/runtime/questionnaire_action.ex
+++ b/lib/ask/runtime/questionnaire_action.ex
@@ -81,7 +81,7 @@ defmodule Ask.Runtime.QuestionnaireExport do
     end)
   end
 
-  defp clean_i18n_quiz(quiz) do
+  def clean_i18n_quiz(quiz) do
     clean_i18n_fields = Map.keys(quiz)
 
     Enum.reduce(clean_i18n_fields, quiz, fn field, quiz_acc ->
--- a/lib/ask/runtime/questionnaire_action.ex
+++ b/lib/ask/runtime/questionnaire_action.ex
@@ -109,7 +109,7 @@ defmodule Ask.Runtime.QuestionnaireExport do
   defp clean_18n_steps(quiz, field) do
     clean_i18n_paths = [
       ".[].prompt",
-      ".[].choices.[].responses.[.sms, .ivr, .mobileweb]",
+      ".[].choices.[].responses.[]",
       ".[].refusal.responses"
     ]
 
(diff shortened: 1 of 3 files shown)


lib/ask/runtime/questionnaire_action.ex
```diff
@@ -0,0 +1,245 @@
...
+        true -> List.flatten(element)
+        false -> element
+      end
+    end)
+  end
+
+  def clean_i18n_quiz(quiz) do
+    clean_i18n_fields = Map.keys(quiz)
+
+    Enum.reduce(clean_i18n_fields, quiz, fn field, quiz_acc ->
```
bcardiff, line 87: Why we need to iterate fields this way?

Can't we be more explicit and simple like:

```elixir
quiz |> clean(:setting, "error_message")
     |> clean(:setting, "thank_you_message")
     |> clean(:steps, ".[].prompt")
     |> clean(:steps, ".[].choices.[].responses.[]")
     |> clean(:steps, ".[].refusal.responses")
     |> clean(:quota_completed_steps, ".[].prompt")
     |> clean(:quota_completed_steps, ".[].choices.[].responses.[]")
     |> clean(:quota_completed_steps, ".[].refusal.responses")
```

Also, why is `error_message` wihout a preceding dot, but `.[]` needs one? I would expect to be `.error_message`

lib/ask/runtime/questionnaire_action.ex
```diff
@@ -0,0 +1,245 @@
...
+
+    clean_elem =
+      Enum.reduce(paths, elem, fn path, elem_acc ->
+        clean_i18n_entity(elem_acc, quiz.languages, path)
+      end)
+
+    Map.put(quiz, key, clean_elem)
+  end
+
+  def clean_i18n_entity(nil, _filter_languages, _path), do: nil
```
bcardiff, line 126: all the clean_i18n_entity can be moved to its own module. They are already tested independently of the quiz, so we are half way there.

(Optional) It's worth adding a couple of comments there as an introduction in that module.

test/lib/runtime/questionnaire_action_test.exs
```diff
@@ -0,0 +1,64 @@
...
+
+      clean = QuestionnaireExport.clean_i18n_entity(entity, ["en"], "")
+
+      assert clean == %{"en" => "foo"}
+    end
+
+    test "move forward" do
+      entity = %{"en" => "foo", "es" => "bar"}
+
+      clean = QuestionnaireExport.clean_i18n_entity(entity, ["en"], ".")
```
bcardiff, line 17: There is not such thing as move forward. `.property` or `.[]` should be consumed. empty path is a base.
But there is no `.` alone. In the current implementation `.....property` = `.property` = `property` which make no sense because only `.property` is a JSON path expression as in jq.

aeatencio:
@bcardiff thanks for your review! Please take a look again!

bcardiff (approved):

PR #1835, opened 2021-02-03
Log added respondents when survey is running

### LINKED ISSUE #1830: Log add sample action to the project activity log
When a user adds sample to an existing survey the action does not appear in the activity log 
Screen Shot 2021-01-19 at 14 31 26
It should be added as it is critical action for a survey

The API POST looks like `/api/v1/projects/25/surveys/346/respondent_groups/343/add`

aeatencio:
There are more actions related to the addition and deletion of respondents.

1. Respondents list creation.
2. Respondents' addition to an existing respondent list.
3. Respondents' replacement of an existing respondent list.
4. Respondents list deletion.

For now, to avoid overloading the activity log unnecessarily, we decided to log only the addition of respondents to an existing respondents list when the survey is running.

This is how it looks:

Added respondents log

Fix #1830

### DIFF
--- a/locales/template/translation.json
+++ b/locales/template/translation.json
@@ -37,7 +37,7 @@
   "Added mode <i>{{mode}}</i> to <i>{{questionnaireName}}": "",
   "Added section <i>{{sectionTitle}}</i> to <i>{{questionnaireName}}</i>": "",
   "Added step <i>{{stepTitle}}</i> to <i>{{questionnaireName}}</i>": "",
-  "Added {{respondentsCount}} respondents to <i>{{respondentGroupName}}</i> respondents list in <i>{{surveyName}}</i>": "",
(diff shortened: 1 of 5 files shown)


bcardiff (commented):
Sounds good, but the respondent group name is something that it means nothing to the user I think. Is something more internal. The log message `Added NN respondents to FILENAME.csv respondents list in SURVEY` is odd.

Knowing the filename might be worth to recall. If so the message should be `Added NN respondents from FILENAME.csv to SURVEY`.

Otherwise just `Added NN respondents to SURVEY` seems good enough for now to me.

cc: @manumoreira WDYT?

manumoreira:
Agree, 
`Added NN respondents from FILENAME.csv to SURVEY` is ideal

`Added NN respondents to SURVEY` is good enough

aeatencio:
@manumoreira @bcardiff this is how it looks now:

activity log

bcardiff (approved):

PR #1837, opened 2021-02-10
Upload sample files with RespondentID

### LINKED ISSUE #1833: Allow users to upload sample files with RespondentID
- [x] If it's a respondent id (/^r.*$/) it should validate that it's an existing respondent in the project
- [x] In the preview it should show the respondent id and not the phone number: r354678976
- [x] Allow mixed respondent id and phone numbers if possible
- [x] Check: when creating a survey and when adding respondents

aeatencio:
### The motivation 🥕 

For the end-users, it would be very useful to run surveys using a subset of respondents of previous surveys.
Because the end-users don't have the respondents' phone numbers, they need to upload the respondents' ids.
One of the expected use cases is the creation of Panel Surveys.

### The solution 🎯 

Allow users to upload sample files with RespondentID.

## The limitations 🚧

This will only work in the context of a Surveda project. 

It fixes #1833


web/static/js/components/surveys/SurveyWizardRespondentsStep.jsx
```diff
@@ -59,22 +59,31 @@ class SurveyWizardRespondentsStep extends Component {
...
       content = (
         <div className='card-content card-error'>
-          <div><b>{t('The file you uploaded does not contain any phone number')}</b></div>
+          <div><b>{t('The file you uploaded does not contain any phone number or respondent id')}</b></div>
           <div>{t('Please upload antother file.')}</div>
         </div>
       )
     } else {
-      const lineNumbers = data.invalidEntries.slice(0, 3).map((entry) => entry.line_number)
+      const invalidPhoneNumberEntries = data.invalidEntries.filter(invalidEntry => invalidEntry.type != 'not-found')
```
bcardiff, line 67: given a list of entries, we can extract a function `{count, extraCount, lineNumbers} = reportError(invalidEntries)` to make the following lines more readable.

bcardiff: What is the type for invalid phone number? i don't see it.

I suggest using a more explicit value like `invalid-phone-number`, `invalid-respondent-id`. It will be easier to track in the source code.

web/models/respondent.ex
```diff
@@ -89,11 +89,11 @@ defmodule Ask.Respondent do
   end
 
   def mask_phone_number(phone_number) do
-    Enum.join([replace_numbers_by_hash(String.slice(phone_number, 0..-5)), String.slice(phone_number, -4, 4)], "")
+    Enum.join([replace_alphanumeric_by_hash(String.slice(phone_number, 0..-5)), String.slice(phone_number, -4, 4)], "")
```
bcardiff, line 92: Revert changes on mask_phone_number.
Add a mask_respondent_entry that will be used at

```
/test/controllers/respondent_group_controller_test.exs
  32,53:       sample = group.sample |> Enum.map(&Respondent.mask_phone_number/1)
  55,53:       sample = group.sample |> Enum.map(&Respondent.mask_phone_number/1)
  493,31:       |> Enum.map(&Respondent.mask_phone_number/1)

/web/views/respondent_group_view.ex
  17,62:     sample = respondent_group.sample |> Enum.map(&Respondent.mask_phone_number/1)
```

Make mask_respondent_entry to return the respondentId without masking. If the input is a phone number, then delegate to mask_phone_number.

---

odly, the sample is stored unmasked in the DB. And is masked on rendering at `respondent_group_view.ex`. I would prefer for the stored sample to be masked. That way we could delegate to entries how to mask themself once and leave the stored that way. But that will come later. 

The phone numbers need to be masked to avoid revealing pii in general, although in this case the user has the number in the csv. I guess we are hiding it for demo/screenshot/recording of surveda.

bcardiff (changes requested):

web/static/js/components/surveys/SurveyWizardRespondentsStep.jsx
```diff
@@ -48,6 +48,23 @@ class SurveyWizardRespondentsStep extends Component {
...
   }
 
+  invalidEntriesText(invalidEntries, invalidEntryType) {
+    const { t } = this.props
+    const filteredInvalidEntries = invalidEntries.filter(invalidEntry => invalidEntry.type == invalidEntryType)
+
+    if (filteredInvalidEntries.length == 0) return null
+
+    const lineNumbers = filteredInvalidEntries.slice(0, 3).map((entry) => entry.line_number)
+    const entryType = () => {
```
bcardiff, line 58: Why is this function?

```
const entryType = {'invalid-phone-number': 'phone number', 'invalid-respondent-id': 'respondent id'}[invalidEntryType]
```

or do the if/else if directly.

web/static/js/components/surveys/SurveyWizardRespondentsStep.jsx
```diff
@@ -48,6 +48,23 @@ class SurveyWizardRespondentsStep extends Component {
...
+      if (invalidEntryType == 'invalid-phone-number') return 'phone number'
+      else if (invalidEntryType == 'invalid-respondent-id') return 'respondent id'
+    }
+    const text = t('An invalid {{entryType}} was found at line {{lineNumbers}}', {lineNumbers: lineNumbers.join(', '), entryType: entryType()})
```
bcardiff, line 62: I know this was as is before, but it would be better to have both entries as a whole

- `An invalid {{entryType}} was found at line {{lineNumbers}}`
- `An invalid {{entryType}} was found at line {{lineNumbers}} and {{count}} more.`

concatenating localized texts is not good. There is no context for `and {{count}} more.` for translating, and is not always the case that concatenating two individual translations leads to a valid phrase in the target language.

bcardiff (changes requested):

aeatencio:
@bcardiff please take a look now 🤞

bcardiff (approved):

PR #1844, opened 2021-02-23
Preserve incentives flag in incarnations

### LINKED ISSUE #1846: New occurrence of a panel survey allow to download incentive file when the sample includes IDs
**Found in version:** 6235794

**Steps to reproduce:**

1. Create a repeatable survey.
2. Add a sample using IDs.
3. Stop the survey and press the repeat button in order to create a new occurrence.
4. Navigate to the respondents tab.
5. Download the incentive file.

**Obtained result:**
- The incentive file can be downloaded.

**Expected result:**
- Since the sample includes IDs, the incentive file download is not allowed

aeatencio:
When a respondent id is uploaded, downloading the incentives file is disabled.

Preserve this setting in panel survey incarnations.

Fix #1846


bcardiff:
Can't we have a spec for this?

bcardiff (approved):

Issue #1601, opened 2019-12-20
Add start date for 'scheduled' surveys


mmuller:
Add start date for **scheduled** surveys in the:
 
- [ ] Survey's overview. 
- [ ] Surveys listing.

Besides that we need to improve the information displayed in the Surveys listing. 
 - [ ] ~Replace `Scheduled at 9 AM (America - Buenos Aires)` with `Next contact 7 days`.~
 - [x] Unify the "survey scheduled for" message in the survey card and overview. See #1855.

![image](https://user-images.githubusercontent.com/518152/71278447-6a7cee80-2336-11ea-8f72-6a64e4ebd34c.png)

![image](https://user-images.githubusercontent.com/518152/71278462-736dc000-2336-11ea-9c4f-ce802cbd63e5.png)

manumoreira:
The "Scheduled for"messages were unified in #1855 but we are still not including the start and end dates when start_date is different from `window_start_date`
We might want to add a new message in the survey overview (it will not fit in the survey card) that specify start and end dates.
A simple option would be a message similar to the one that is shown when the survey is closed (defined by surveyRanDescription) in black font below the "Scheduled for" message.

aeatencio:
First part of the message:

1. The survey was launched but it didn't entered in its first active window -> `Didn't start yet`
2. The survey alredy entered in its first active window -> `Started on <first_window_started_at>`

Second part of the message:

1. The survey has an `end_date` -> ` and will end on <last_window_end_at>`
2. The survey doesn't has an `end_date` -> no second part of the message.

mmuller:
Now in version 7235dc9 the `end_date` is showing 1 day before the `end_date`

![image](https://user-images.githubusercontent.com/518152/114453221-b436bb80-9baf-11eb-9b40-12a464598d5c.png)

![image](https://user-images.githubusercontent.com/518152/114453262-bef15080-9baf-11eb-80b4-ff29e1ce2fbf.png)

aeatencio:
> Now in version 7235dc9 the end_date is showing 1 day before the end_date

It will be fixed for #1859. Please let's track this there 🙏

mmuller:
Awesome! Working properly in version 7235dc9

PR #1855, opened 2021-03-09
Unify the "survey scheduled for" message

aeatencio:
This is how it looks in the survey card:

survey card

And this is the full message in the survey overview:

overview-en

overview-es

overview-fr

For #1601


web/static/js/components/surveys/SurveyStatus.jsx
```diff
@@ -36,57 +37,52 @@ class SurveyStatus extends PureComponent {
...
       timeZone: timezones.items[survey.schedule.timezone],
+      year: 'numeric',
+      month: 'short',
+      day: 'numeric',
       hour12: true,
       hour: 'numeric'
     }
-    let time = date.toLocaleTimeString(locale, options)
-    return `${time} (${formatTimezone(survey.schedule.timezone)})`
+    let dateString = date.toLocaleDateString(language, options).replace(/[,.]/g, '')
```
bcardiff, line 82: Why we need to remove `,` and `.`? Some date formats might use `.` as date component separators. I would suggest to `.replace(/a\.m\./, 'AM')` if the goal is to use AM instead of a.m. . Also PM should be handled.

web/static/js/components/surveys/SurveyStatus.jsx
```diff
@@ -215,8 +211,10 @@ class SurveyStatus extends PureComponent {
 }
 
 const mapStateToProps = (state) => {
+  const { settings } = state.userSettings
   return {
-    timezones: state.timezones
+    timezones: state.timezones,
+    language: settings && settings.language || 'en'
```
bcardiff, line 217:
```suggestion
    language: i18next.language
```

I would expect to use the current language from i18n and not duplicate the logic of how the current language is defined and which one is the default.

Does that do not work?

It can also be called within `nextCallDescription`, there is no need to make it a prop since it's a global setting and the `t` func already depends on `i18next`

web/static/js/components/surveys/SurveyStatus.jsx
```diff
@@ -188,13 +184,13 @@ class SurveyStatus extends PureComponent {
...
         color = 'text-error'
-        text = t('Unknown', {context: 'survey'})
+        text = t('Unknown', { context: 'survey' })
     }
 
     let component = (
       <span>
         <i className='material-icons survey-status'>{icon}</i>
-        { text }
+        { text}
```
bcardiff, line 193:
```suggestion
        { text }
```

bcardiff (changes requested):

web/static/js/components/surveys/SurveyStatus.jsx
aeatencio: These are the new messages (the dots aren't removed any more):

```
Programada para 22 mar. 2021 9 AM (America - Buenos Aires)
Scheduled for Mar 22 2021 9 AM (America - Buenos Aires)
Planifié pour 22 mars 2021 à 9 AM (America - Buenos Aires)
````

Comment (aeatencio) [reply web/static/js/components/surveys/SurveyStatus.jsx]: And... after midday:
```
Programada para 16 mar. 2021 3 PM (America - Buenos Aires)
Scheduled for Mar 16 2021 3 PM (America - Buenos Aires)
Planifié pour 16 mars 2021 à 3 PM (America - Buenos Aires)
```

PR #1857, opened 2021-03-16
Improve UI for panel surveys

### LINKED ISSUE #1856: Improve the UI for panel surveys
Change the look and feel of the latest occurrence of a panel survey card [this way](https://invis.io/2TMXFAVHVQM#/180693793_Surveys)

Also, hide the cards of the other occurrences of the panel survey.

When a panel survey has a only survey, show it and link it the same way as panel surveys with multiple occurrences.

When the panel survey is selected, show all the panel survey occurrences using the same design of the [survey folders](https://invis.io/2TMXFAVHVQM#/378385603_Survey_Menu_-_Folder)

Change the `add survey` button of the folder page for a new `repeat` button, without removing the current `repeat` button on the panel survey overview. 

Review if the name of the panel survey occurrences should stay editable.

For #1805

aeatencio:
## The motivation 🥕

Since #1815 panel surveys exist in Surveda. What are panel surveys? See #1805
The 1st iteration of panel surveys left some pending issues. One of these issues is supporting them in the survey index UI.

## The solution 🎯 

Panel surveys are already grouped by their `panel_survey_of` property.
This PR groups them in a single (new, visually different) card that links to their panel survey folder.
The panel survey folder implementation reuses the regular survey folder component.

Fix #1856

### Panel survey model

During the review of this PR, we found the lack of an entity `panel survey` is a problem.

🎯 Temporarily, we mocked a `panel survey` model endpoint, but we should bring a real model to life soon!

### Panel survey folders

The folder membership of a survey is independent of its panel survey membership.
A user can move different surveys of the same panel survey to different folders, which would bring problems to the new UI coherence.

🎯 To prevent this situation, we are prohibiting changing the folder of panel surveys. It's cheaper than moving all the panel survey occurrences together.

### Panel survey deletion

The lack of an entity `panel survey` in the back-end model makes it difficult to delete the first occurrence of a panel survey.

🎯 To handle this limitation, we are prohibiting deleting the first occurrence of a panel survey until there is a `panel survey` entity in the back-end model.

### Odd panel surveys ⚠️ 

During the local Dev QA warnings like these were found in the browser console while navigating the survey index and folders:

warning

The warning message is something like `When merging two panelSurveys, found unequal data in their "latestSurveyId" values. Using the earlier value.`.

This warning is a symptom of a wrong state in the panel surveys. It's shown when more than 1 survey is flagged as the latest survey of a panel survey. It shouldn't happen in production. Let's pay attention to this situation.

### Changes in the breadcrumb 👀 

Surveda supports only 1 level of depth: it doesn't support having a folder inside another folder.
But from now on a panel survey is represented as a folder. So for panel surveys inside a folder, the breadcrumb needs to support 1 level more.

⚠️ Please pay attention to the breadcrumb in the print screens below.

### Project:

project

### Regular survey:

regular survey

### Panel survey:

panel survey

### Panel survey occurrence:

panel survey occurrence

### Folder:

folder

### Regular survey in folder:

regular survey in folder

### Panel survey in folder:

panel survey in folder

### Panel survey occurrence in folder:

panel survey occurrence in folder


web/static/js/components/folders/FolderShow.jsx
```diff
@@ -104,25 +121,50 @@ class FolderShow extends Component<any, any> {
...
+      if (panelSurvey) {
+        primaryButton = (
+          <RepeatButton text={t('Repeat survey')} disabled={panelSurvey.isRepeatable} onClick={() => this.repeatSurvey()} />
+        )
+      } else {
+        primaryButton = (
+          <AddButton text={t('Add survey')} onClick={() => this.newSurvey()} />
+        )
+      }
+    }
```
anaPerezGhiglia, line 135: what do you think if we extract this `primaryButton` to a separate function to avoid scripting in the middle of the `render()` function

Also, with this the `primaryButton` could be a `const` instead of a `let` 😏

web/static/js/components/folders/FolderShow.jsx
```diff
@@ -104,25 +121,50 @@ class FolderShow extends Component<any, any> {
...
+    if (panelSurvey) {
+      hint = <div className='repeat-survey row'>
+        <div className='col s12 hint'>
+          {
+            panelSurvey.isRepeatable
+            ? t('The last occurence of the panel survey is complete, you may follow up with a new survey sent to a subset of the respondents of this panel survey')
+            : t("The last occurence of the panel survey isn't complete yet. After that, you may follow up with a new survey sent to a subset of the respondents of this panel survey")
+          }
+        </div>
+      </div>
```
anaPerezGhiglia, line 152: same as `primaryButton`. I would extract all this logic to a simple and dedicated function

web/static/js/components/folders/FolderShow.jsx
```diff
@@ -28,14 +30,16 @@ class FolderShow extends Component<any, any> {
     totalCount: PropTypes.number.isRequired,
     respondentsStats: PropTypes.object.isRequired,
     params: PropTypes.object,
-    id: PropTypes.number,
+    folderId: PropTypes.number,
```
anaPerezGhiglia, line 33: If now there is a `surveyFolder`, couldn't we remove this prop and when necessary just use `surveyFolder.id` ?

web/static/js/components/folders/FolderShow.jsx
```diff
@@ -138,13 +180,25 @@ class FolderShow extends Component<any, any> {
...
   // Right now we show all surveys: they are not paginated nor sorted
   let surveys = state.surveys.items
 
   if (surveys) {
-    surveys = values(surveys).filter(s => s.folderId == id)
+    if (panelSurvey) {
+      surveys = values(surveys).filter(s => panelSurvey && s.panelSurveyOf == panelSurvey.id)
+    } else {
+      surveys = values(surveys).filter(s => s.folderId == folderId)
+    }
```
anaPerezGhiglia, line 201: a little comment here wouldn't hurt 🙃  (and extra points for extracting this along with the others assignments to `surveys` that are bellow)

If you like it, here are some functions (may have errors since I haven't been writing something in js for a long time)

```javascript
const filterSurveys = (allSurveys, panelSurvey, folder) => {
  const values = allSurveys ? values(allSurveys) : []
  return if (panelSurvey) {
    // If a panelSurvey is present, then we should filter those survey which have the same parent
    values.filter(s => panelSurvey && s.panelSurveyOf == panelSurvey.id)
  } else {
    // Is a common folder, so we should filter those surveys that are inside of it
    values.filter(s => s.folderId == folder.id)
  }
}

const sortAndSliceSurveys = (surveys, pageIndex, pageSize) => {
  // Sort by updated at, descending
  const sorted = (surveys || []).sort((x, y) => y.updatedAt.localeCompare(x.updatedAt))
  // Show only the current page
  return values(sorted).slice(pageIndex, pageIndex + pageSize)
}

```

web/static/js/components/layout/HeaderContainer.jsx
```diff
@@ -48,22 +47,39 @@ HeaderContainer.propTypes = {
...
+  const { params } = ownProps
   const folders = state.folder && state.folder.folders
   const survey = state.survey && state.survey.data
-  const surveyFolder = survey && folders && folders[survey.folderId]
+  const panelSurveyId = params.panelSurveyId && parseInt(params.panelSurveyId)
+
+  let panelSurvey = null
+  if (state.panelSurvey && state.panelSurvey.id == panelSurveyId) {
+    panelSurvey = state.panelSurvey
+  }
```
anaPerezGhiglia, line 63: **I** prefer `const`s to `let`s when possible

web/static/js/components/surveys/SurveyCard.jsx
```diff
@@ -74,54 +75,80 @@ class SurveyCard extends Component<any> {
       {survey.description}
     </div>
 
+    const hideSurveyCard = !inPanelSurveyFolder && survey.isPanelSurvey && !panelSurvey
```
anaPerezGhiglia, line 78: This is for not showing surveys in in the survey index? 
I'm a bit confused with this

web/static/js/decls/survey.js
```diff
@@ -40,6 +40,15 @@ export type Survey = {
   incentivesEnabled: boolean
 };
 
+export type PanelSurvey = {
+  id: number,
+  name: string,
+  folderId: number,
+  projectId: number,
+  isRepeatable: boolean,
+  latestPanelSurveyId: number
```
anaPerezGhiglia, line 49: remember to change it here as well! 🙃

web/static/js/components/surveys/SurveyCard.jsx
anaPerezGhiglia: Shouln't the index ignore those surveys?

aeatencio: Yeah, probably. I'll try to do it this way! 🙃

web/static/js/components/folders/FolderShow.jsx
aeatencio: Yeah, probably! I'll try it! Thanks!

web/static/js/components/layout/HeaderContainer.jsx
aeatencio: Yeah! I prefer `const` too!! I'll try it harder! 😉

anaPerezGhiglia (approved):
LGTM 🙂

Issue #1859, opened 2021-03-17
'end_date' of survey's schedule should stop 5 minutes after the last survey activity window ends


mmuller:
If I select `Mar 18, 2021` as `end_date`, the survey should be stopped 5 minutes after the last survey activity window before `Mar 18, 2021` at `23:59:59` ends. So if the last survey activity window ends Mar 18, 2021 at 3 PM, the survey should end Mar 18, 2021 at 3:05 PM 

![image](https://user-images.githubusercontent.com/518152/111527634-e6830380-873e-11eb-8332-a8b0c886095d.png)

aeatencio:
Check the end-user message is fixed while verifying this issue. See [this comment](https://github.com/instedd/surveda/issues/1601#issuecomment-818107998).

aeatencio:
@mmuller before verifying this issue please read the #1862 description.

aeatencio:
@mmuller please consider the #1862 description while verifying this issue. Thanks!

mmuller:
Working properly in version f4c69e6

PR #1852, opened 2021-03-04
Add end date to schedule

aeatencio:
### Is this a draft ❓ 

No, it isn't. It's ready for review, but I keep it as a draft because before merging it #1850 has to be merged and after that the base branch has to be changed to master.

### What are we doing❓ 

We're adding `start_date` and `end_date` to the survey schedule.

#1850 adds the `start_date` field and this PR adds the `end_date` field.

When a survey is running and the end date is reached, it should stop automatically just as it does when the end-user stops a survey manually.

end date

Fix #1838

PR #1850, opened 2021-02-26
Add start date to Schedule

aeatencio:
### The motivation 🥕 

In regular surveys, and particularly in panel surveys, it'd be useful to schedule the start date of a survey.
This PR adds the `start date` setting to the survey Schedule section.

### Expected behavior 🗒️ 

- [x] use block date UI in settings
- [x] prevent the survey from sending messages before that date. the survey will appear as started.
- [x] add legend start date in card
- [x] (not editable once the survey is launched)

### Done ✅ 

- [x] Select and deselect the start date from the UI
- [x] Honor the end-user selection
- [x] Disable selection after the survey is launched
- [x] UI styling
- [x] Test schedule manually
- [x] Add tests

### Pending 🗒️ 

- [ ] Validate start date is after today and before the first blocked-date

Fix #1838

### How it looks 📷 

Editing survey schedule

---

Scheduled running survey

---

Running survey schedule

---

### Last minor style change 🖌️ 

Reduced space between start-date and block-dates

Reduced space between start-date and block-dates

PR #1860, opened 2021-03-22
Show a schedule clarificacion message

### LINKED ISSUE #1601: Add start date for 'scheduled' surveys

aeatencio:
### The motivation 🥕

Since #1850 and #1852 we need (more than before) to easily let the user know when the survey has started (or when it will start) and when it will be automatically canceled.

Before these PRs, we already had some dates to consider (`started_at` and `ended_at`). They mean different things.

`started_at` -> when the user pressed the launch button
`start_date` -> the first active window won't start before this date
`first_window_started_at` -> when the survey first active window began
`end_date` -> the last active window won't finish after this date

⚠️ There are pending changes for the `end_date` meaning (see #1859). The schedule clarification message will change accordingly to them.

Fix #1601

### Survey cards 📷 :

cards

### Survey overview 📷 :

scheduled

running

completed

cancelled

### The `starting` corner case ⚠️ 

When the survey will start immediately (there will be no distance between `started_at` and `first_window_started_at`) there is a little time window while the survey isn't active only because its first `survey_poll` didn't occur yet. During this time window we tell the user that the survey is `starting`, this way:

starting card

starting overview


locales/template/translation.json
```diff
@@ -428,15 +428,7 @@
...
-  "Started {{count}} day ago": "",
-  "Started {{count}} day ago_plural": "",
-  "Started {{count}} hour ago": "",
-  "Started {{count}} hour ago_plural": "",
-  "Started {{count}} minute ago": "",
-  "Started {{count}} minute ago_plural": "",
-  "Started {{count}} second ago": "",
-  "Started {{count}} second ago_plural": "",
-  "Started {{date}}": "",
+  "Started on {{firstWindowStartedAt}}": "",
```
anaPerezGhiglia, line 431: Shouldn't the new `Didn\'t start yet` string be here as well?

test/controllers/survey_controller_test.exs
```diff
@@ -64,7 +65,7 @@ defmodule Ask.SurveyControllerTest do
...
+        %{"cutoff" => survey.cutoff, "id" => survey.id, "mode" => survey.mode, "name" => survey.name, "description" => survey.description, "project_id" => project.id, "state" => "not_ready", "locked" => false, "exit_code" => nil, "exit_message" => nil, "schedule" => %{"blocked_days" => [], "day_of_week" => %{"fri" => true, "mon" => true, "sat" => true, "sun" => true, "thu" => true, "tue" => true, "wed" => true}, "end_time" => "23:59:59", "start_time" => "00:00:00", "start_date" => nil, "end_date" => nil, "timezone" => "Etc/UTC"}, "next_schedule_time" => nil, "started_at" => started_at |> Timex.format!("%FT%T%:z", :strftime), "ended_at" => nil, "updated_at" => DateTime.to_iso8601(survey.updated_at), "down_channels" => [], "folder_id" => folder.id, "first_window_started_at" => started_at |> Timex.format!("%FT%T%:z", :strftime)}
```
anaPerezGhiglia, line 68: It goes without saying that this is not mandatory for merging the PR, but extracting this `Timex.format!(date, "%FT%T%:z", :strftime)` into a test-util or something would be nice.

If not, in the future, if the format changes we will have to change all this tests instead of just one function 🙃

web/models/survey.ex
```diff
@@ -648,4 +648,16 @@ defmodule Ask.Survey do
   def delete_multi(survey) do
     Multi.delete(Multi.new, :survey, survey)
   end
+
+  def first_window_started_at(%{started_at: nil} = _survey), do: nil
+
+  def first_window_started_at(%{schedule: schedule, started_at: started_at} = _survey) do
+    calculated_started_at = Schedule.next_available_date_time(schedule, started_at)
```
anaPerezGhiglia, line 655: I suppose that the `Schedule.next_available_date_time` will always return the same given the same schedule and date, right?

Also, a survey's schedule never changes?

Is there a reason for calculating this every time instead of storing it as field in the Survey table? 

I'm wondering this ☝️ thinking of reports that may be extracted from the db. Are there such about a survey? is this field necessary?

locales/template/translation.json
aeatencio: Yes, it was missing! Thanks!

test/controllers/survey_controller_test.exs
aeatencio: Agree! Thanks!!

web/models/survey.ex
aeatencio: > is this field necessary?

Yes, it is. I wasn't aware!

Changing it! Thanks!!! 🙌

lib/ask/runtime/broker.ex
```diff
@@ -115,6 +115,11 @@ defmodule Ask.Runtime.Broker do
       reached_quotas = reached_quotas?(survey)
       survey_completed = survey.cutoff <= completed || reached_quotas
       batch_size = batch_size(survey, by_state)
+      survey = if survey.first_window_started_at do
+        survey
+      else
+        Survey.changeset(survey, %{first_window_started_at: SystemTime.time.now}) |> Repo.update!
+      end
```
anaPerezGhiglia, line 122: I would extract this to a separate function for better understanding. I'm thingking of something like
```elixir
def update_survey_on_poll(survey)
```
what do you think?

test/lib/runtime/broker_test.exs
```diff
@@ -587,8 +590,13 @@ defmodule Ask.Runtime.BrokerTest do
 
       survey1 = Repo.get(Ask.Survey, survey1.id)
       survey2 = Repo.get(Ask.Survey, survey2.id)
+
       assert Ask.Survey.completed?(survey1)
       assert survey2.state == "running"
+
+      # Only polled surveys have first_window_started_at
+      assert survey1.first_window_started_at == now
+      refute survey2.first_window_started_at
```
anaPerezGhiglia, line 599: I believe the `first_window_started_at` logic deserves its own tests (like you did un `survey_test` before) 🙃 

We need to test that:
 - this field is set when a survey is polled for the first time
 - when a survey is polled for second time, the field is not updated
 - when a survey is created has this field in `nil`

aeatencio:
@anaPerezGhiglia I think we're OK now. Please take a look again. Thanks!!!

anaPerezGhiglia (approved):
Loving the new simple and descriptive tests 🎉 
Nice work

aeatencio:
Evaluate replacing the `play_arrow` icon by [this one](https://material.io/resources/icons/?search=hour&icon=hourglass_top&style=baseline) on the `starting` corner case.

PR #1861, opened 2021-03-31
Avoid replicating `end_date` on survey repetition

### LINKED ISSUE #1839: Allow setting end date of survey's schedule
The survey will be stopped automatically on that date. This is equivalent to cancelling the survey.

aeatencio:
We don't want to replicate `start_date` nor `end_date` on survey repetition. It doesn't make sense to do it.

This PR avoids replicating `end_date` on survey repetition using the same logic we already used to avoid replicating `start_date` on survey repetition.

Fix #1839

### DIFF
--- a/lib/ask/runtime/panel_survey.ex
+++ b/lib/ask/runtime/panel_survey.ex
@@ -11,6 +11,11 @@ defmodule Ask.Runtime.PanelSurvey do
       survey
       |> Repo.preload([:project])
 
+    schedule =
+      survey.schedule
+      |> Schedule.remove_start_date()
+      |> Schedule.remove_end_date()
+
     new_ocurrence = %{
       # basic settings
       project_id: survey.project_id,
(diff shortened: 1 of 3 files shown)


lib/ask/runtime/panel_survey.ex
```diff
@@ -11,6 +11,11 @@ defmodule Ask.Runtime.PanelSurvey do
       survey
       |> Repo.preload([:project])
 
+    schedule =
+      survey.schedule
+      |> Schedule.remove_start_date()
+      |> Schedule.remove_end_date()
```
matiasgarciaisaia, line 17: Let's try to avoid imperative names for functions, since they don't mutate a thing. This two could be `Schedule.without_start_date()`, for example.

matiasgarciaisaia (approved):
Great one! Merge away! :shipit:

lib/ask/runtime/panel_survey.ex
aeatencio: I see what you mean, but I think this isn't the standard way of naming functions in Elixir.

Here are a couple of examples (the standard library is plenty of them):

 [Map.replace/3](https://hexdocs.pm/elixir/Map.html#replace/3)
 ```
Map.replace(%{a: 1, b: 2}, :a, 3)
%{a: 3, b: 2}
```

[Map.delete/2](https://hexdocs.pm/elixir/Map.html#delete/2)
```
Map.delete(%{a: 1, b: 2}, :a)
%{b: 2}
```

PR #1862, opened 2021-04-09
Honor the end date

### LINKED ISSUE #1859: 'end_date' of survey's schedule should stop 5 minutes after the last survey activity window ends

aeatencio:
## The motivation 🥕

When the end-user picks an `end date`, they expect the survey ends at the ending of the last active window before the last minute of the selected date.

## What this PR does 🎯 

It honors the end-user expectation about the `end date` configuration.

Fix #1859

## ⚠️ A corner case and its technical debt 🐌 

There is some technical debt regarding bad schedule configurations.

⚠️ The corner case: the schedule configuration is bad because there are no active days (e.g., it runs only on Wednesdays, it starts on Tuesday and ends on the same day).

🗺️ The situation: the algorithm that calculates the next and the last active window fails trying to find them.

🎯 The (partial) solution: the algorithm fails with a custom exception, the controller responds 409, and the UI crashes when trying to launch the survey.

🐌 The technical debt: validate bad survey configurations before the survey is launched.

## How it looks 👓 :

survey schedule

survey overview


test/lib/ecto_types/schedule_test.exs
```diff
@@ -142,6 +143,52 @@ defmodule Ask.ScheduleTest do
...
+        end_time: ~T[18:00:00],
+        day_of_week: %Ask.DayOfWeek{sun: true, wed: true},
+        timezone: "America/Argentina/Buenos_Aires",
+        start_date: ~D[2017-03-03],
+        blocked_days: [~D[2017-10-08]]
+      }
+      {:ok, base_schedule: base_schedule}
+    end
+
+    test "gets last available time: unavaible day", %{base_schedule: base_schedule} do
```
anaPerezGhiglia, line 169: typo here: I assume you wanted to write `unavailable` :)

test/lib/ecto_types/schedule_test.exs
```diff
@@ -142,6 +143,52 @@ defmodule Ask.ScheduleTest do
...
+
+    test "gets last available time: this day", %{base_schedule: base_schedule} do
+      schedule = %{base_schedule | end_date: ~D[2017-10-11]}
+
+      time = Schedule.last_window_ends_at(schedule)
+
+      assert time == DateTime.from_naive!(~N[2017-10-11 21:00:00], "Etc/UTC")
+    end
+
+    test "gets next available time: blocked day", %{base_schedule: base_schedule} do
```
anaPerezGhiglia, line 185: `get last available time` right?

test/lib/ecto_types/schedule_test.exs
```diff
@@ -142,6 +143,52 @@ defmodule Ask.ScheduleTest do
...
+    test "gets next available time: start date" do
+      # Long before the start date
+      base = DateTime.from_naive!(~N[2017-02-01 00:00:00], "Etc/UTC")
+      time = @schedule |> Schedule.next_available_date_time(base)
+      # The beginning of the first active window after the start date
+      assert time == DateTime.from_naive!(~N[2017-03-05 12:00:00], "Etc/UTC")
+    end
+  end
+
+  describe "last_window_ends_at" do
```
anaPerezGhiglia, line 156: To understand that these tests were right took me quite a while
- I had to see that only Sundays and Wednesdays where valid
- I had to go look for a calendar (go back to 2017) and see which days (numbers) where valid or not
- And then I had to realize that the schedule timezone is `gmt-3` and the assert is in `UTC`

For the dates, I would at least express the schedule options in a comment. Like:
```
The valid dates for the survey on October '17 are:
- Sun 1
- Wed 4
- ~Sun 8~ (No because is blocked)
- Wed 11
- Sun 15
```

For the times, can we use only one timezone?

aeatencio:
OK, so I'll try to improve the `Ask.Schedule.last_window_ends_at` tests. They were inspired by the `Ask.Schedule.next_available_date_time` tests. I see they are kind of cryptic.

Now please review the new `Survey.expired?` tests. They were all inspired by myself 😉

test/lib/ecto_types/schedule_test.exs
aeatencio: Sure! Totally agree! I'll change it! Thanks!!

test/lib/ecto_types/schedule_test.exs
aeatencio: Yes... I'll review every test to improve its expressiveness. This included!

test/lib/ecto_types/schedule_test.exs
aeatencio: I don't think so, but yes, I'll improve the expressiveness of this test too. Thanks!!

lib/ask/ecto_types/schedule.ex
```diff
@@ -143,37 +143,91 @@ defmodule Ask.Schedule do
...
+    backward = false
+    limit = end_date
+    next_available_date = reversible_next_available_date_time(schedule, from_date_time, backward, limit)
+
+    if limit_exceed(next_available_date, backward, limit) do
+      raise ScheduleError, "next active window not found"
+    else
+      next_available_date
+    end
+  end
```
anaPerezGhiglia, line 158: This logic is repeated here and in `last_window_ends_up`. also, it seems leaked responsibility from `next_available_date`.

Why not instead of returning `erl_date` on `next_available_date` when `limit_exceed` we raise the `ScheduleError`? The error will bubble up and should have the same effect

lib/ask/ecto_types/schedule.ex
```diff
@@ -225,21 +262,39 @@ defmodule Ask.Schedule do
...
-    if day_of_week_available?(schedule, erl_date) do
+  defp next_available_date(schedule, erl_date, backward, limit) do
+    if limit_exceed(erl_date, backward, limit) do
       erl_date
     else
-      next_available_date(schedule, erl_date)
+      shift_days = if backward, do: -1, else: 1
+      erl_date = Timex.shift(erl_date, days: shift_days)
+      if day_of_week_available?(schedule, erl_date) do
+        erl_date
```
anaPerezGhiglia, line 277: here raise `ScheduleError`. So we don't have to validate the dates more than once

lib/ask/ecto_types/schedule.ex
```diff
@@ -225,21 +262,39 @@ defmodule Ask.Schedule do
...
-  defp next_available_date(schedule, erl_date) do
-    erl_date = Timex.shift(erl_date, days: 1)
-    if day_of_week_available?(schedule, erl_date) do
+  defp next_available_date(schedule, erl_date, backward, limit) do
+    if limit_exceed(erl_date, backward, limit) do
       erl_date
     else
-      next_available_date(schedule, erl_date)
+      shift_days = if backward, do: -1, else: 1
+      erl_date = Timex.shift(erl_date, days: shift_days)
```
anaPerezGhiglia, line 275: for better understanding I would rename this variable for `next_date` -> so that it expresses that is a different date than the one passed as parameter

lib/ask/ecto_types/schedule.ex
```diff
@@ -143,37 +143,91 @@ defmodule Ask.Schedule do
...
+          if backward do
+            erl_date
+          else
+            next_available_date(schedule, erl_date, backward, limit)
+          end
       end
     else
-      # If the day is not enabled, find the next day
-      next_available_date_time_internal(schedule, erl_date)
+      next_available_date(schedule, erl_date, backward, limit)
```
anaPerezGhiglia, line 229: What do you think of something like this?

```elixir
def select_available_date(schedule, erl_date, time, backward, limit) do
    date_to_return = if day_of_week_available?(schedule, erl_date) do
      select_date_if_time(compare_time(schedule, time), backward)
    else
      :next_date
    end

    case date_to_return do
      :given_date -> erl_date
      :next_date -> next_available_date(schedule, erl_date, backward, limit)
    end
  end
```

where the new `select_date_if_time` fuction would be defined as follow

```elixir
  defp select_date_if_time(:before = _time_is, true = _backward), do: :next_date
  defp select_date_if_time(:before = _time_is, false = _backward), do: :given_date
  defp select_date_if_time(:after = _time_is, true = _backward), do: :given_date
  defp select_date_if_time(:after = _time_is, false = _backward), do: :next_date
```

I don't like the name for this new function, so If you find a better one, do change it please! 

PS: the tests run green with this 🙂

lib/ask/ecto_types/schedule.ex
```diff
@@ -225,21 +262,39 @@ defmodule Ask.Schedule do
...
+      erl_date = Timex.shift(erl_date, days: shift_days)
+      if day_of_week_available?(schedule, erl_date) do
+        erl_date
+      else
+        next_available_date(schedule, erl_date, backward, limit)
+      end
+    end
+  end
+
+  defp limit_exceed(_date_time, _backward, nil = _limit), do: false
```
anaPerezGhiglia, line 284: Not mandatory (because is just naming), but I would change it to `date_exceeds_limit?` or `limit_exceeded?`

aeatencio:
@anaPerezGhiglia please take a look again! Thanks!!!

anaPerezGhiglia (approved):
Nice work!
I believe not only now we can find the next suitable window going backward, but also the code is easier to understand 👏

PR #1872, opened 2021-04-27
Fix export when choices responses

### LINKED ISSUE #1868: Questionnaire export fails - SMS mode and choices
Found in version f4c69e6 for #1869

### Steps to reproduce:

1. Create a questionnaire.
2. Set SMS mode only
3. Add choices (any quantity, even empty responses) to the question.

### Result:

The export fails: 500 (server error) is thrown

aeatencio:
Under certain conditions, it crashes.

Because of a bug in the clean algorithm when the content of the requested key isn't a map.

Fix #1868

### DIFF
--- a/lib/ask/runtime/questionnaire_action.ex
+++ b/lib/ask/runtime/questionnaire_action.ex
@@ -227,4 +227,6 @@ defmodule Ask.Runtime.CleanI18n do
       Map.delete(entity_acc, lang)
     end)
   end
+
+  defp clean_base_case(entity, _langs), do: entity
 end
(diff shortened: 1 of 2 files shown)


anaPerezGhiglia (commented):
Ok, with this when adding choices the app won't crash, but does it clean the choices list?
I'm probably losing something, but for what I reach to understand, this will not really fix the example you commented on the test, right?

Please enlighten me! 🙏

aeatencio:
> does it clean the choices list?
> this will not really fix the example you commented on the test, right?

Yes, it does clean the choices list. I've tried to clarify it with a new test.

@anaPerezGhiglia, please review it again and let me know. Thanks!

anaPerezGhiglia (approved):
Nice! thanks for the new test 😄

PR #1873, opened 2021-04-28
Add more basic questionnaire tests

### LINKED ISSUE #1867: Write a minimal set of unit tests for exporting questionnaires
Include only some basic cases

aeatencio:
New tests:

1. IVR - exports a simple questionnaire
2. IVR - exports a simple questionnaire with audios
3. Mobile Web - exports a simple questionnaire

Fix #1867

### DIFF
--- a/test/lib/runtime/questionnaire_action_test.exs
+++ b/test/lib/runtime/questionnaire_action_test.exs
@@ -120,6 +120,20 @@ defmodule Ask.Runtime.QuestionnaireExportTest do
       }
     }
 
+    @ivr_audio_id "6fe2fac8-18bf-48f7-970c-204d2f3408b0"
+
+    @ivr_audio_question_prompt %{
+      "en" => %{
+        "sms" => "",
+        "mobileweb" => "",
+        "ivr" => %{
+          "text" => "",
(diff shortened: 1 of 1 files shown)


anaPerezGhiglia (commented):
LGTM 👍 

What I'm thinking is that perhaps it would be useful to have support functions to create 
`choices`, `steps` or `settings` and configure the quiz with them, since are things we want to change from one test to another.
If we have such thing, perhaps instead of building all the possible configurations in the setup, each test would configure the questionnaire to its needs.

I'm thinking of something like (in pseudocode)
```elixir
newQuestionnaire("sms") 
|> withSettings(thankyou_message: "thanks", error_message: "error") 
|> withSteps(buildMultipleChoiceSMS())
```

What do you think?

anaPerezGhiglia (approved):
Nice!
I leave you a message for a small improvement, but it is good to merge!

🤓 I would extract the quiz configurations to a support file, so it can be used in other tests if necessary and to avoid making the test file difficult to follow

aeatencio:
> I would extract the quiz configurations to a support file.

I'll do it (or, at least, I'll try to do it) soon, I promise!

PR #1876, opened 2021-04-30
Do not promote ineligible respondents to survey repetitions

### LINKED ISSUE #1875: Do not promote ineligible respondents to survey repetition
There's no reason to contact again a respondent that has been flagged as 'Ineligible' in a previous survey.
So the respondents that will not be promoted to repetition are: `ineligible and `refused`

aeatencio:
### The motivation (fix #1875) 🥕

> There's no reason to contact again a respondent that has been flagged as 'Ineligible' in a previous survey.
So the respondents that will not be promoted to repetition are: `ineligible` and `refused`


anaPerezGhiglia (approved):
Nice!

Issue #1887, opened 2021-05-11
Split repeat action into clone + run


aeatencio:
When a user clicks on repeat it leads the user to a read-only survey edition screen in which they can click on run.

It includes making the UI totally work again with panel surveys.

Depends on #1893

PR #1891, opened 2021-05-17
Build a panel survey API

### LINKED ISSUE #1890: Panel survey API (1st iteration)
Include every standard endpoint
CRUD PS and its dependencies
For #1886

aeatencio:
## 🚫 Don't merge yet!

Merge to `master` together with #1896 and after solving the following issues: #1893, #1887, and #1897.
Otherwise, Surveda will crash badly in `master`.

---

The new API was manually-directly tested and it seems to be working properly.

Fix #1890 

## Known pending tests:
- _Unhappy_ cases. E.g. when the user doesn't have access to the project.
- Corner cases: I don't know which corner cases, so these aren't really _known_ pending tests.


web/controllers/panel_survey_controller.ex
```diff
@@ -1,48 +1,67 @@
...
 
-    panel_surveys = Enum.map(latest_panel_surveys, fn s -> prepare_panel_survey(s) end)
+    panel_surveys = project
+      |> assoc(:panel_surveys)
+      |> Repo.all()
 
     render(conn, "index.json", panel_surveys: panel_surveys)
   end
 
+  def create(conn, %{"project_id" => project_id, "panel_survey" => panel_survey_params}) do
```
anaPerezGhiglia, line 17: So, currently this is not exposed right? because in the router the `new` and `edit` are excluded.. Or am I missing something?

test/models/survey_test.exs
```diff
@@ -185,4 +185,16 @@ defmodule Ask.SurveyTest do
       assert expired? == true
     end
   end
+
+  describe "panel survey" do
+    test "creates a survey inside a panel survey" do
```
anaPerezGhiglia, line 190: I don't understand what we are asserting here. What we are comparing is the result of the insert, with getting the data from the db (that isn't always the same as what we inserted?)

test/models/panel_survey_test.exs
```diff
@@ -0,0 +1,244 @@
...
+    assert result == :ok
+    assert panel_survey.id == updated_panel_survey.id
+
+    # The name was updated
+    refute panel_survey.name == updated_panel_survey.name
+    assert updated_panel_survey.name == updated_name
+  end
+
+  # TODO: updating a panel survey project shouldn't be allowed.
+  # Where do we put this restriction. At the model level or at the controller level?
```
anaPerezGhiglia, line 94: The business logic should not be in the controller, thus, the restriction should be in the model or action

anaPerezGhiglia: do you have this restriction in mind for the new changes?

anaPerezGhiglia (commented):
Nice work! 👏

web/models/panel_survey.ex
```diff
@@ -0,0 +1,130 @@
...
+  @doc """
+  Returns an `%Ecto.Changeset{}` for tracking panel_survey changes.
+
+  ## Examples
+
+      iex> change_panel_survey(panel_survey)
+      %Ecto.Changeset{source: %PanelSurvey{}}
+
+  """
+  # TODO: find out the purpose of this function
```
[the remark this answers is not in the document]
aeatencio: Yes, it's. It was generated by Phoenix, but we don't use it. Do we remove it?

web/controllers/panel_survey_controller.ex
aeatencio: I think this is just following the Surveda routes standard.

The available routes for panel surveys and their corresponding controller functions are:

```
GET /api/v1/projects/:project_id/panel_surveys Ask.PanelSurveyController :index
GET /api/v1/projects/:project_id/panel_surveys/:id Ask.PanelSurveyController :show
POST /api/v1/projects/:project_id/panel_surveys Ask.PanelSurveyController :create
PATCH /api/v1/projects/:project_id/panel_surveys/:id Ask.PanelSurveyController :update
PUT /api/v1/projects/:project_id/panel_surveys/:id Ask.PanelSurveyController :update
DELETE /api/v1/projects/:project_id/panel_surveys/:id Ask.PanelSurveyController :delete
```

aeatencio: Answering your question: yes, the `create` function of this controller is exposed on this route: `POST /api/v1/projects/:project_id/panel_surveys`

test/models/survey_test.exs
aeatencio: We're just checking that a survey can be assigned to a panel survey. An assertion checking exactly that was missing. I've just added it. Please take a look again.

The comparison between the (factory created) inserted panel survey with the loaded panel survey from the DB is just that. Trying to check that nothing was missing after DB insertion and reading.

test/models/panel_survey_test.exs
aeatencio: > the restriction should be in the model
Done!

web/models/panel_survey.ex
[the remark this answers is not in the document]
aeatencio: I've removed it

aeatencio:
@anaPerezGhiglia, please take a look again! Thanks!!!

test/models/survey_test.exs
```diff
@@ -195,6 +195,7 @@ defmodule Ask.SurveyTest do
       created_survey = Repo.get!(Survey, survey.id) |> Repo.preload([:project, panel_survey: :project])
 
       assert created_survey == survey
+      assert survey.panel_survey_id
```
anaPerezGhiglia, line 198: shouldn't we assert is the same as `panel_survey.id` ?

anaPerezGhiglia (approved):
LGTM 🙂

test/models/survey_test.exs
aeatencio: Yes, it makes a lot of sense! 🙃

Issue #1893, opened 2021-05-27
Make the UI work again


aeatencio:
Using the new API version
Don't mind panel surveys

For #1886
Depends on #1892

PR #1896, opened 2021-05-28
Clean previous panel survey model

### LINKED ISSUE #1892: Clean previous PS model based on new API
It includes the #1887 back-end

For #1886

aeatencio:
## 🚫 Don't merge yet!

Merge to `master` together with #1891 and after solving the following issues: #1893, #1887, and #1897.
Otherwise, Surveda will crash badly in `master`.

---

The panel survey model has changed in #1891.

Remove the obsolete code of the old panel survey and adapt Surveda to use the new one.

This completes the first iteration of the back-end part. The front-end part will be fixed for #1893 and #1887.
The migration of the existing surveys will be done for #1897 after checking that the whole App is working properly with the new model.

Fix #1892


test/controllers/survey_controller_test.exs
```diff
@@ -2576,49 +2572,48 @@ defmodule Ask.SurveyControllerTest do
...
       survey = insert(:survey, project: project)
 
       conn = get(conn, project_survey_path(conn, :show, project, survey))
 
-      assert json_response(conn, 200)["data"]["is_panel_survey"] == false
+      assert json_response(conn, 200)["data"]["panel_survey_id"] == nil
     end
 
-    test "sets up a panel survey", %{conn: conn, project: project} do
+    test "sets a panel survey", %{conn: conn, project: project} do
```
anaPerezGhiglia, line 2591: we should't be able to to this right?

anaPerezGhiglia, line 2591: we should't be able to to this right?

test/controllers/survey_controller_test.exs
```diff
@@ -2576,49 +2572,48 @@ defmodule Ask.SurveyControllerTest do
...
-      assert survey.panel_survey_of == survey.id
-      assert survey.latest_panel_survey == true
+      assert json_response(conn, 200)["data"]["panel_survey_id"] == panel_survey.id
+      assert survey.panel_survey_id == panel_survey.id
     end
 
-    test "sets up a regular survey", %{conn: conn, project: project} do
-      survey = insert(:survey, project: project)
-      survey = Survey.changeset(survey, %{panel_survey_of: survey.id}) |> Repo.update!()
+    test "unsets a panel survey", %{conn: conn, project: project} do
```
anaPerezGhiglia, line 2605: this either..

test/lib/runtime/panel_survey_test.exs
```diff
@@ -0,0 +1,217 @@
+defmodule Ask.Runtime.PanelSurveyTest do
+  use Ask.ModelCase
+  use Ask.TestHelpers
+  alias Ask.Runtime.{PanelSurvey, RespondentGroupAction}
+  alias Ask.{Survey, Repo, TestChannel, Respondent}
+
+  describe "new occurence" do
+    test "creates a new ready occurrence" do
+      panel_survey = completed_panel_survey()
```
anaPerezGhiglia, line 9: I would rename this function. The panel survey is not completed. The last ocurrence of it is completed. In any case, it would be `repeatable_panel_survey` or something like that

test/lib/runtime/panel_survey_test.exs
```diff
@@ -0,0 +1,217 @@
...
+    end
+
+    test "the new occurence is based on the latest occurrence" do
+      panel_survey = completed_panel_survey()
+      previous_occurrence = Ask.PanelSurvey.latest_occurrence(panel_survey)
+      latest_occurrence = insert(:survey, panel_survey: panel_survey, state: "terminated")
+
+      {:ok, %{new_occurrence: new_occurrence}} = PanelSurvey.new_occurrence(panel_survey)
+
+      # TODO: Don't compare the name, compare something more meaningfull
```
anaPerezGhiglia, line 60: pending TODO. also, the `refute` and `assert` aren't straightforward to understand why

test/support/test_helpers.ex
```diff
@@ -124,6 +124,31 @@ defmodule Ask.TestHelpers do
...
+
+      defp complete_panel_survey(panel_survey) do
+        Ask.PanelSurvey.latest_occurrence(panel_survey)
+        |> terminate_survey()
+
+        # Reload the panel survey. One of its surveys has changed, so it's outdated
+        Repo.get!(Ask.PanelSurvey, panel_survey.id)
+      end
+
+      defp completed_panel_survey() do
```
anaPerezGhiglia, line 148: the previous comment belongs here :)

anaPerezGhiglia (commented):
The migration from `panel_survey_of` to `panel_survey_id` with the corresponding creation of panel survey will be addressed in a separate PR?

aeatencio:
> The migration from panel_survey_of to panel_survey_id with the corresponding creation of panel survey will be addressed in a separate PR?

Yes, @anaPerezGhiglia. Tentatively this is my idea because I'm not sure exactly which data we'll need to migrate yet.

This is why I let the following comment on this PR description:
> This completes the first iteration of the back-end part. The front-end part will be fixed for #1893 and #1887.
The migration of the existing surveys will be done for #1897 after checking that the whole App is working properly with the new model.

What do you think about this idea?

test/controllers/survey_controller_test.exs
aeatencio: I think we should allow doing it for now. Because currently, the UI first creates de survey, and then it's "converted" into a panel survey.

test/controllers/survey_controller_test.exs
aeatencio: Same here. Currently, the UI allows to "convert" a survey into a panel survey, and the other way round.

test/lib/runtime/panel_survey_test.exs
aeatencio: Yes, I didn't like this name, but I could find a better one so I didn't change it. I'm not sure if `repeatable` is the best choice. It could be! Let's discuss it today at the call 🙏 . In the meantime, I'll change it using more descriptive (but longer) phrases.

test/lib/runtime/panel_survey_test.exs
aeatencio: Please let's talk about this test on a call 🙏

aeatencio:
@anaPerezGhiglia please take a look at the last changes and comments and let me know. Thanks!

test/controllers/survey_controller_test.exs
anaPerezGhiglia: yo are right!

test/lib/runtime/panel_survey_test.exs
anaPerezGhiglia: So, after what we talked in the call the other day, this test should assert that the channel is the same, the respondents are the same, also the schedule (and so on.., right?)

anaPerezGhiglia:
Looking good! 
It sound reasonable to address the migration in another PR, sorry for missing the PR description 🙇‍♀️

test/lib/runtime/panel_survey_test.exs
aeatencio: Yes! This is pending. Thanks for remembering me it!

Issue #1903, opened 2021-06-24
Panel survey view doesn't update when creating/deleteing ocurrence


manumoreira:
When creating a new occurrence of a panel survey or delete an occurrence hard refresh is needed to get the view updated.
Steps to reproduce:
- Create a panel survey (launch and complete the first occurrence)
- In the panel survey view create new occurrence and before launching go back to the panel survey view
- The new occurrence will not be shown
- Hard refresh, it will be shown
- Delete the last occurrence (still in edition) and the occurrence will still be shown
- Hard refresh......

For #1886

mmuller:
Now (version acf7266) the new occurrence is shown, but it is shown as `Editing` instead of `Running`.

![image](https://user-images.githubusercontent.com/518152/124638908-e62f8500-de61-11eb-96c8-4dcef2022595.png)

PR #1907, opened 2021-06-29
Fix occurrences in panel survey index

aeatencio:
Fid #1903

### DIFF
--- a/web/static/js/reducers/panelSurvey.js
+++ b/web/static/js/reducers/panelSurvey.js
@@ -1,10 +1,22 @@
 // @flow
 import * as actions from '../actions/panelSurvey'
+import * as surveyActions from '../actions/surveys'
 import fetchReducer from './fetch'
 
 export const dataReducer = (state: PanelSurvey, action: any): PanelSurvey => {
   switch (action.type) {
+    case surveyActions.DELETED: return deleteSurvey(state, action)
     default: return state
+  }
+}
+
+const deleteSurvey = (state, action) => {
+  const occurrences = state.occurrences
+  ? state.occurrences.filter(occurrence => occurrence.id != action.id)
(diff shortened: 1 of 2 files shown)


web/static/js/reducers/panelSurvey.js
```diff
@@ -1,13 +1,25 @@
...
 import fetchReducer from './fetch'
 
 export const dataReducer = (state: PanelSurvey, action: any): PanelSurvey => {
   switch (action.type) {
+    case surveyActions.DELETED: return deleteSurvey(state, action)
     default: return state
   }
 }
 
+const deleteSurvey = (state, action) => {
```
anaPerezGhiglia, line 13: Can't we go fetch the panel Survey to the api like you did when creating a `newOcurrence`?

If the backend for some reason (like if is the only occurrence of a PS) does not deletes the occurence, the UI will not show it (when it should)

aeatencio: The `survey.DELETED` action shouldn't be triggered if the survey isn't deleted. If it does, we have a problem in another place, and I don't get why to do defensive programming here. If the action is triggering badly, we'll have problems [here](https://github.com/instedd/surveda/blob/511c8e79df9cf40a9420539755ea48ac3f5b19ca/web/static/js/reducers/surveys.js#L7) also. It's almost exactly the same situation.

As I see it, what we do by fetching the panel survey after creating a new occurrence is a workaround. There we should (ideally) follow a redux flow similar to this and not the other way round.

anaPerezGhiglia: ok, if the action shouldn't be triggered if not _actually_ deleted then seems ok.

anaPerezGhiglia (approved):

PR #1913, opened 2021-06-30
Update Redux with the panel survey new state

### LINKED ISSUE #1903: Panel survey view doesn't update when creating/deleteing ocurrence

### LINKED ISSUE #1911: The UI shows a delete option that shouldn't be there
When a panel survey has 2 occurrences and 1 of them is deleted, the remaining occurrence isn't anymore deletable, because the panel survey can't be empty.

Currently, in this situation, the `is_deletable` property isn't correctly updated. Therefore on the occurrence card, there's a `delete` option that shouldn't be there. If the user tries to delete the occurrence, they can't (the back-end rejects it, and that's OK!), so the UI crashes (without any proper explanation) when the user clicks on the delete option.

Make the `is_deletable` occurrence property update properly after deleting an occurrence. That will fix the problem.

aeatencio:
After occurrence creation and deletion

Fix #1903
Fix #1911

### DIFF
--- a/web/static/js/actions/survey.js
+++ b/web/static/js/actions/survey.js
@@ -3,6 +3,7 @@ import * as api from '../api'
 import each from 'lodash/each'
 import { stepStoreValues } from '../reducers/questionnaire'
 import * as surveysActions from './surveys'
+import * as panelSurveyActions from '../actions/panelSurvey'
 
 export const CHANGE_CUTOFF = 'SURVEY_CHANGE_CUTOFF'
 export const CHANGE_QUOTA = 'SURVEY_CHANGE_QUOTA'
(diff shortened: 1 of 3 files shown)


anaPerezGhiglia (approved):
LGTM

PR #1937, opened 2021-07-29
Avoid future behaviour misunderstanding

aeatencio:
Below I'll try to explain why IMHO the lines this commit removes can only lead to misunderstandings and the reasons to remove them.

In previous iterations of the panel survey development, the new wave took its questionnaires from the latest wave of its panel survey.

But currently, the end-user selects the questionnaire/s after the new wave is created.

Because of them, the new wave could wrongly appear (to future Devs or maintainers) to be associated with a questionnaire.

But that appearance is just that because the UI doesn't show the `wave -> questionnaires` created association/s to the end-user. Conversely, it forces them to select a new questionnaire before launching the new wave.

### DIFF
--- a/lib/ask/runtime/panel_survey.ex
+++ b/lib/ask/runtime/panel_survey.ex
@@ -36,12 +36,9 @@ defmodule Ask.Runtime.PanelSurvey do
       incentives_enabled: survey.incentives_enabled
     }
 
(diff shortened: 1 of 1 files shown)


anaPerezGhiglia (approved):
Nice cleanup 🧹

Issue #1949, opened 2021-09-28
Pages count fail in panel surveys


diegoliberman:
In panel surveys with less than 9 waves, all the waves should be shown on page 1.
Something is not worlimng properly in the page counter (at the bottom right corner)

![image](https://user-images.githubusercontent.com/25181052/135128503-333d1ea6-cb0f-4da6-890c-ab7c3601bdf9.png)

ysbaddaden:
The problem is here:
https://github.com/instedd/surveda/blob/master/web/static/js/components/surveys/PanelSurveyShow.jsx#L162-L163

I think we're simply missing a `pageSize`, which I understand is the number of surveys to list per-page.

ysbaddaden:
Errata: there is no pagination of the list of surveys inside a panel survey. The backend merely returns the panel survey with all its surveys at once. We can work around the issue by faking it until we decide to implement it (or not).

I.e. display "1-X of X", for example "1-9 of 9".

diegoliberman:
@manumoreira we need to decide if it's necessary to implement an actual pagination right now.

ysbaddaden:
I implemented the workaround: fake it until you can make it.

To decide whether we should implement actual pagination, we should look at how many waves there usually are under a panel survey, if it's slowing things down (client and/or backend), displays too many things, or if it's acceptable. It can be done on the server (don't return too many surveys) or the client (server returns everything, client does the pagination).

PR #1957, opened 2021-10-04
Tiny UI corrections

### LINKED ISSUE #1947: Missing `n` of respondents needed in Cockpit
In https://surveda-ph.org/projects/26/surveys/381 the number of respondents needed to complete the target is missing

image

### LINKED ISSUE #1948: Change new panel survey wave icon
Currently, we are using [replay icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Areplay%3A) we should replace that one with [repeat one icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Arepeat_one%3A)

ysbaddaden:
1. fixes #1947

   Capture d'écran du 2021-10-04 11 55 53@2x

2. closes #1948 (the icon is grey because I have a pending survey)
  
   Capture d'écran du 2021-10-04 15 38 20@2x

3. refs #1949 

   There are no pagination. The easy patch is to just fake it, and see later whether we implement an actual pagination, or not.

### DIFF
--- a/web/static/js/components/charts/QueueSize.jsx
+++ b/web/static/js/components/charts/QueueSize.jsx
@@ -105,7 +105,7 @@ class QueueSize extends Component<Props, State> {
               <path style={{display: needed ? 'auto' : 'none'}} className='dottedLine' d={this.connector(left.x1, left.y1, left.x2, left.y2, weight - (left.x1 > left.x2 && right.x1 > right.x2 ? corner : 0), corner)} />
               <path style={{display: needed ? 'auto' : 'none'}} className='dottedLine' d={this.connector(right.x1, right.y1, right.x2, right.y2, weight, corner)} />
               <g ref='content'>
-                <text className='needed'>{t('{{respondentsNeeded}} respondents needed', {needed: toKilo(needed)})}</text>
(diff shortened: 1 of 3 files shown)


devduarte:
Code looks good!

I was thinking about centering this icon with its description, so we continue with another UI fix. What do you think?
Screen Shot 2021-10-04 at 10 42 50

ysbaddaden:
@devduarte That would be better, but this is Firefox not supporting the `baseline-shift` CSS property for SVG, which is deprecated —to be replaced with `vertical-align` that doesn't seem to have any impact whatsoever. There is `dominant-baseline` but that didn't work properly.

The only solution that worked was to use the `dy` attribute instead, but that's mixing CSS (`font-size`) and DOM (`dy`), which ain't ideal, but maybe closer to how SVG works?

```svg
<text class="missing">
  <tspan class="icon">warning</tspan>
  <tspan dy="-6px">Add XX additional respondents</tspan>
</text>
```

There are also occurences in the code of `baseline-shift` that would need to be fixed, too.

devduarte:
Ok! then lets merge as it is now and we can take care of it later on! 🚀

Issue #1968, opened 2021-10-12
Implement actual pagination of surveys in a panel survey


ysbaddaden:
Follow-up of #1949 where we ended up faking the pagination since there is no pagination.

We should to decide if it's necessary to implement an actual pagination right now.

Related to to #1954 where we notice that we don't have any actual pagination of surveys or backend filters, and we load every single survey upfront (even those inside a folder or panel survey).

ysbaddaden:
Merged into the `develop` branch.

Issue #1966, opened 2021-10-07
Upgrade React


diegoliberman:

ysbaddaden:
See https://github.com/instedd/surveda/pull/1977#discussion_r748994888

PR #1977, opened 2021-10-26
Load surveys for the current page

### LINKED ISSUE #1954: Enhance survey loading process in project overview
The survey loading process has very bad performance, every time a user opens a project all surveys in the project are loaded.
## Details
Surveys can be in the project root, in a project folder, or in a panel survey (a survey that holds many repetitions of a survey)
So apart from basic pagination, we should also control the loading of the surveys in folders and panel surveys.

We do not have a selected approach for this, ideas are welcome

ysbaddaden:
Modifications to the backend are complete. Just a few controller tests to fix.

- When listing surveys, we only list them at the root (i.e. not inside a folder or inside a panel survey).
- When listing panel surveys, we only them at the root (i.e. not inside a folder), but also include it's latest occurrence (for respondent stats).
- Fetching a single folder now returns its surveys and panel surveys.
- Fetching a single panel survey now returns its occurrences (i.e. surveys).

The frontend is also capable to handle those cases. It will fetch an individual folder or panel survey or survey as needed, on demand. The components no longer always request the current project then all its folders, surveys and panel surveys all the time. Tests have been fixed (mocha, flow, eslint).

~~One part is still broken: the breadcrumb. I'm still not sure how to fix it. Maybe:~~ The breadcrumb is fixed.

1. ~~by requesting each individual project, folder, panel survey? that will still request lots of data;~~
2. by appending some information about the parent project, folder and panel survey to surveys and panel surveys?

~~I'm trying to implement~~ I implemented solution 2.

closes #1954 #1968

### DIFF
--- a/web/static/js/components/surveys/SurveyIndex.jsx
+++ b/web/static/js/components/surveys/SurveyIndex.jsx
@@ -49,11 +49,7 @@ class SurveyIndex extends Component<any, State> {
     }
   }
 
-  componentWillMount() {
-    this.initialFetch()
-  }
-
-  initialFetch() {
+  componentDidMount() {
     const { dispatch, projectId } = this.props
     dispatch(actions.fetchSurveys(projectId))
     dispatch(foldersActions.fetchFolders(projectId))
--- a/web/controllers/survey_controller.ex
+++ b/web/controllers/survey_controller.ex
@@ -78,8 +78,6 @@ defmodule Ask.SurveyController do
         survey = survey
         |> Repo.preload([:quota_buckets])
         |> Repo.preload(:questionnaires)
-        # |> Repo.preload(:folder)
-        # |> Repo.preload(panel_survey: [:folder])
         |> Survey.with_links(user_level(project_id, current_user(conn).id))
 
         conn
--- a/web/static/js/components/layout/HeaderContainer.jsx
+++ b/web/static/js/components/layout/HeaderContainer.jsx
@@ -66,20 +66,22 @@ const mapStateToProps = (state, ownProps) => {
   const panelSurveyId = params['panelSurveyId']
   let panelSurvey, folder
 
-  // Here the order of the factors does alter the product.
-  // Depending on the case, we need to take the folder from the panel survey taken from params,
-  // from the survey taken from params, or from the panel survey taken from its occurrence.
-  // For example evaluating (surveyFromParams || panelSurveyFromOccurrence) would work well but
-  // just for some cases. It would work for getting the folder of a regular survey, but it
-  // wouldn't work for getting the folder of an occurrence of a panel survey.
+  // We build the breadcrumb in reverse order, and never display the current
+  // level (the currently opened folder, panel survey or survey), only the
+  // parents:
   if (surveyId) {
-    const survey = state.survey.data && state.survey.data.id == parseInt(surveyId) && state.survey.data
-    panelSurvey = survey ? survey.panelSurvey : null
-    folder = panelSurvey ? panelSurvey.folder : (survey ? survey.folder : null)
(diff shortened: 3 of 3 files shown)


web/static/js/components/surveys/PanelSurveyShow.jsx
```diff
@@ -129,11 +88,45 @@ class PanelSurveyShow extends Component<any, any> {
...
+  }
+
+  footer() {
+    const { startIndex, endIndex, totalCount } = this.props
+
+    return <PagingFooter
+      {...{startIndex, endIndex, totalCount}}
+      onPreviousPage={() => this.previousPage()}
+      onNextPage={() => this.nextPage()} />
+  }
```
ftarulla, line 129: Code refactor! 👏

web/static/js/components/surveys/SurveyEdit.jsx
```diff
@@ -34,19 +31,11 @@ class SurveyEdit extends Component {
   componentWillMount() {
```
ftarulla, line 31: I think we should create a tech-debt-issue to move "fetching elements" from `componentWillMount` to `componentDidMount` (or vice versa) so that we are consistent across all components.

web/static/js/components/surveys/SurveyIndex.jsx
```diff
@@ -58,13 +55,8 @@ class SurveyIndex extends Component<any, State> {
 
   initialFetch() {
```
ftarulla, line 56: In this component we use `componentWillMount` but with an auxiliar function (that adds semantic). 
(Not for this PR but) As I mentioned before, we should try to be consistent across components so that reading the code gets easier.

web/static/js/reducers/folder.js
```diff
@@ -2,93 +2,22 @@
...
-    ...newState,
-    error: null
-  }
-}
-
-const includeError = (state, action) => {
-  const newState = {...state}
-  newState.folders[action.id].error = action.error
-  return newState
+  fetching: false,
```
ftarulla, line 5: Just a detail: should we use `loading`?

web/static/js/reducers/folders.js
```diff
@@ -0,0 +1,83 @@
+// @flow
+import * as actions from '../actions/folders'
+
+const initialState = {
+  loading: false,
+  error: null,
+  loadingFetch: false,
```
ftarulla, line 7: I think we should replace `loadingFetch` with another name. It gets confused to read `loading` and `loadingFetch`?
From line 58 it looks like `loading` is actually `saving`?

web/static/js/components/surveys/SurveyEdit.jsx
ysbaddaden: We can add this to #1966 since `componentWillMount()` is deprecated anyway.

web/static/js/components/surveys/SurveyIndex.jsx
ysbaddaden: And... we don't need to have this initialFetch() method anyway!

web/static/js/reducers/folder.js
ysbaddaden: AFAIK `loading` in Surveda is meant as "we're saving something and are waiting to _load_ the updated data", whereas `fetching` is use to mean we're fetching the data.

This seems to be consistent, so I kept the naming.

web/static/js/reducers/folders.js
ysbaddaden: Yeah, it's confusing but it's also consistent across actions & reducers.

ftarulla (approved):
🍟 🚀

PR #1980, opened 2021-10-28
Fix: simulation ended toast can be displayed multiple times

### LINKED ISSUE #1971: 'Simulation ended' toast can be displayed multiple times
**Found in version:** 516879a

**Preconditions:**

- There is a questionnaire `Q1`.

- `Q1` supports the `mobile web` mode.
- The last question in `Q1` is a select one step.

**Steps to reproduce:**

1. Navigate to the mobile web simulator (without channel).
2. Answer the `Q1`'s questions (complete it).
3. Click several times on the last question answer.

**Obtained result:**
- The `Simulation ended` toast has multiple occurrences.

**Screenshot:**
image

ysbaddaden:
Mark a simulation has ended when it expired or ended, so we can
never display the expired/ended toast multiple times. This didn't
happen for the SMS and IVR simulators, but the mobileweb one runs in
its own iframe, and can send a message multiple times.

Refactors to only listen to the `message` event for the mobileweb
simulator, and removes the event manually added to `window` when the
component is unmounted, otherwise we'd have a memory leak with
unmounted components continuing to receive the mobileweb simulator
messages, leading to errors in the console, as well as memory that
the Garbage Collector would never reclaim.

closes #1971

### DIFF
--- a/web/static/js/components/simulation/QuestionnaireSimulation.jsx
+++ b/web/static/js/components/simulation/QuestionnaireSimulation.jsx
@@ -129,10 +129,10 @@ class QuestionnaireSimulation extends Component<Props, State> {
     if (isSimulationTerminated) {
       switch (simulationStatus) {
         case 'expired':
-          window.Materialize.toast(t(`This simulation is expired. Please refresh to start a new one`))
+          window.Materialize.toast(t('This simulation is expired. Please refresh to start a new one'))
           break
         case 'ended':
-          window.Materialize.toast(t(`This simulation is ended. Please refresh to start a new one`))
+          window.Materialize.toast(t('This simulation is ended. Please refresh to start a new one'))
           break
       }
     }


web/static/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -87,21 +104,27 @@ class QuestionnaireSimulation extends Component<Props, State> {
...
+
+    if (event.data.simulationChanged && simulation) {
+      const { projectId, questionnaireId } = this.props
+      return fetchSimulation(projectId, questionnaireId, simulation.respondentId, mode).then(result => {
+        this.onSimulationChanged(result)
+      })
     }
   }
 
   onSimulationChanged = result => {
```
ftarulla, line 120: I think we can re-write `onSimulationChanged`:
```
    const { simulation, isSimulationTerminated } = this.state
    const { t } = this.props

    if (isSimulationTerminated) {
      return
    }

    const { simulationStatus } = result
    const isSimulationTerminated = simulationStatus == 'expired' || simulationStatus == 'ended'

    if (isSimulationTerminated) {
      window.Materialize.toast(t(`This simulation is ${simulationStatus}. Please refresh to start a new one`))
    }

    let newSimulationState = {
      ...simulation,
      simulationStatus: simulationStatus
    }

    if (simulationStatus != 'expired') {
      // When the simulation expires we avoid refreshing (and so, erasing) the current state
      // So we only update the simulation status and we send a message to the end user
      // This allow us to handle this particular situation properly

      newSimulationState = {
        ...newSimulationState,
        messagesHistory: result.messagesHistory, // mode=sms
        prompts: result.prompts,                 // mode=ivr
        submissions: result.submissions,
        disposition: result.disposition,
        currentStep: result.currentStep
      }
    }

    this.setState({
      simulation: newSimulationState,
      isSimulationTerminated: isSimulationTerminated
    })
```
It's easy to see the definition of `simulation terminated` (and when to show the toast). Also it's easy to see when to update the complete simulation state (as stated in the comment). And last (but not least) the state is set in one place at the end of the function.

(note: I didn't try it, please verify this comment)

ftarulla (commented):
I'm leaving some code improvements. Hope they make the code cleaner 🤓

web/static/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -52,26 +52,43 @@ type Props = {
...
-    window.addEventListener('message', event => {
-      const { mode } = this.props
-      const { simulation } = this.state
-      if (event.data.simulationChanged && simulation) {
-        const { projectId, questionnaireId } = this.props
-        return fetchSimulation(projectId, questionnaireId, simulation.respondentId, mode).then(result => {
-          this.onSimulationChanged(result)
+
+    if (mode == 'mobileweb') {
+      this._onMessage = event => { this.onMessage(event) }
```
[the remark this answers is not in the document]
ysbaddaden: I need the lambda so `this` is bound correctly, otherwise I'd do a simple `addEventListener(message, this.message)`.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
ysbaddaden: You're right. That's much better.

ysbaddaden: I only kept the translatable messages as-is, so they can still be extracted automatically, otherwise they'd get removed from the `translation.js` file.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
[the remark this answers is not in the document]
ftarulla: Sorry, you're right! `this` won't be bound correctly

web/static/js/components/simulation/QuestionnaireSimulation.jsx
ftarulla: Great!

ftarulla (approved):
🍟 🚀

PR #1983, opened 2021-10-29
Attempt to silence race condition on audio play/load/pause

### LINKED ISSUE #1982: Bug: play() request was interrupted by a call to pause()
The `play()` method is asynchronous, which means that sometimes a `pause()` may be intertwined between calling `play()` and the audio _really_ starting to play.

Ref. <https://sentry.io/organizations/instedd/issues/2752736081/>
See <https://developers.google.com/web/updates/2017/06/play-request-was-interrupted> for details

refs #1970

ysbaddaden:
fixes #1982 
fixes ASK-STG-RV

### DIFF
--- a/web/static/js/components/simulation/VoiceWindow.jsx
+++ b/web/static/js/components/simulation/VoiceWindow.jsx
@@ -37,6 +37,7 @@ function audioURL(ivr: IVRPrompt): string {
 
 const VoiceWindow = translate()(class extends Component<VoiceWindowProps, VoiceWindowState> {
   audio: HTMLAudioElement
+  playPromise: Promise<any>
 
   message: string
   messageTimer: TimeoutID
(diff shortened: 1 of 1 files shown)


web/static/js/components/simulation/VoiceWindow.jsx
```diff
@@ -64,7 +64,9 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps, VoiceW
       this.audio.pause()
       this.audio.src = audioURL(ivr)
       this.audio.onended = () => { this.play() }
-      this.audio.play()
+
+      const promise = this.audio.play()
+      promise && promise.catch(() => {}) // silence async play errors
```
ftarulla, line 69: Should we do something like in the example [here]( https://developers.google.com/web/updates/2017/06/play-request-was-interrupted)
```
const promise = this.audio.play()
if (promise !== undefined) {
    promise.then(_ => {
      // We can now safely pause video...
      this.audio.pause();
    })
    .catch(error => {
      // Auto-play was prevented
      // Show paused UI.
    });
  }
```

And remove line 64 `this.audio.pause()` that I think is part of the problem

ysbaddaden: I wasn't sure, but we can indeed delay the handling of the `play()` promise to the next call to `VoiceWindow.play()` and we can silence the error at this moment to pause/load/play the next IVR audio prompt, with a fallback for the first audio _and_ older browsers that don't return a promise.

ftarulla (approved):

Issue #1978, opened 2021-10-26
~~Upgrade Flow~~ Migrate to TypeScript


ysbaddaden:
The currently version of Flow (0.71) lags far behind the current release (0.162). We may consider upgrading to a more recent version.

The upgrade process probably won't be smooth. A quick try at Flow 0.162 for example leads to 915 errors, mostly because Flow now expects almost everything to be statically typed (e.g. full function signatures, including anonymous functions).

Alternative: consider [TypeScript](https://www.typescriptlang.org/), with an [automated translation](https://github.com/khan/flow-to-ts).

diegoliberman:
@ysbaddaden excellent analysis! As it seems this will require quite some effort, my only recommendation is to analyze it in the context of the rest of the upgrades that we want to do.  Perhaps it's more convenient to start with a different upgrade from the ones that we have in the tech debt backlog.

diegoliberman:
Removed pending estimation: 32h

PR #1989, opened 2021-11-15
Sandbox: add waveform to IVR simulator

ysbaddaden:
Animated waveform while audio is playing (canvas):

Capture d'écran du 2021-11-15 11 15 52@2x

### DIFF
--- a/web/static/js/components/simulation/VoiceWindow.jsx
+++ b/web/static/js/components/simulation/VoiceWindow.jsx
@@ -60,7 +60,9 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps, VoiceW
   }
 
   componentWillUnmount() {
-    this.spectrum?.stop()
+    if (this.spectrum) {
+      this.spectrum.stop()
+    }
   }
 
   play() {
@@ -163,6 +165,7 @@ class VoiceSpectrum {
 
   constructor() {
     this.audioContext = new AudioContext()
+    // $FlowFixMe: shut up, AnalyserNode's constructor does take arguments: https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/AnalyserNode
     this.audioData = new AnalyserNode(this.audioContext, { fftSize: 256 })
     this.amplitudeData = new Uint8Array(0)
     this.playing = false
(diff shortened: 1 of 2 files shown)


web/static/js/components/simulation/VoiceWindow.jsx
```diff
@@ -122,9 +139,121 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps, VoiceW
...
+        height,
+        halfHeight: height / 2
+      }
+    }
+  }
+
+  start(audio) {
+    const audioSource = this.audioContext.createMediaElementSource(audio)
+    audioSource.connect(this.audioContext.destination)
+    audioSource.connect(this.audioData)
```
ftarulla, line 193: From an example I was reading (https://github.com/mdn/voice-change-o-matic/blob/gh-pages/scripts/app.js#L128-L205) I thought we should do something like this:

```
audioSource.connect(this.audioData)
this.audioData.connect(this.audioContext.destination)
```
Maybe is the same? (or at least, it has the same result?) 🤔

ysbaddaden: Oh, you got me wondering why it worked: AudioAnalyser doesn't process the sound (its output is equivalent to its input) and merely provides an interface to extract data about the audio. Connecting the audio source to both the analyser and the destination, or connecting through the analyser are equivalent :nerd_face:

Yet, it would be different with a Gain or Equalizer that would process/transform the audio, so it's better to properly do the connection!

web/static/js/components/simulation/VoiceWindow.jsx
```diff
@@ -122,9 +139,121 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps, VoiceW
...
+  amplitudeData: Uint8Array
+  canvas: HTMLCanvasElement
+  context: CanvasRenderingContext2D
+  sizes: { width: number, height: number, halfHeight: number }
+  playing: boolean
+
+  constructor() {
+    this.audioContext = new AudioContext()
+    // $FlowFixMe: shut up, AnalyserNode's constructor does take arguments: https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/AnalyserNode
+    this.audioData = new AnalyserNode(this.audioContext, { fftSize: 256 })
```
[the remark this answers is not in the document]
ysbaddaden: Sigh, the constructor is the recommended solution, but yeah, let's go with `createAnalyser()`. I'll remember this as one more reason to migrate to TypeScript (#1978)!

web/static/js/components/simulation/VoiceWindow.jsx
ftarulla: yeah it seemed that it was equivalent. Great!

ftarulla (approved):
🍟 🚀

PR #1993, opened 2021-11-15
Add restart button to simulator ended/expired notification

### LINKED ISSUE #1985: Refreshing the session when the simulation is ended throws a JS error
**Steps to reproduce:**

1. Select the phone mode
2. Start a simulation
3. Complete the "Survey"
4. Refresh the session

**Obtained:**
- A JS error is thrown

**Error details:**
https://sentry.io/organizations/instedd/issues/2761590658/?project=105692&referrer=alert_email

ysbaddaden:
Basically: one big hack :computer: :fire: 

1. The toast is outside's react scope/context, so we can't call a method on the component, and have to create a global function on the `window` object that can access the react router.

   Not pretty, and I tried to use the `Toast` component, but it's still outside of React scope. Maybe a new version of `react-materialize` would allow to specify a HTML content that's linked to a React context?

2. `react-router` won't reload a route if nothing changed (and we can't force it). One solution is to router to another path (e.g. `/`) then replace it back. 

   Again, not pretty. There may be other solutions?

fixes #1985


locales/translations/es.json
```diff
@@ -451,6 +451,7 @@
     "Response must not be blank": "La respuesta no debe estar en blanco",
     "Responses": "Respuestas",
     "Responsive": "Responsivo",
+    "Restart simulation": "",
```
ysbaddaden, line 454: Missing spanish translation of the button.

ftarulla: It should be something like: _Reiniciar simulación_ or _Reiniciar la simulación_
@diegoliberman, who should approve this translations?

diegoliberman: Let's translate what we feel confident ourselves.  "Reiniciar simulación" sounds good.  If we come up with a case that we don't know how to translate, we'll ask Manu.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -83,12 +83,30 @@ class QuestionnaireSimulation extends Component<Props, State> {
...
+      window.event.preventDefault()
+
+      // HACK: react-router won't navigate to the same URL and doesn't have a
+      // solution to force a refresh... we thus force a refresh by pushing a
+      // fake route, that will unload the current context, and immediately
+      // replace it for the simulator page, that will trigger a refresh.
+      //
+      // See https://dev.to/zbmarius/react-route-refresh-without-page-reload-1907
+      const pathname = window.location.pathname
+      router.push('/')
```
ftarulla, line 100: Do we have this problem? (from the post)
_But because the app's router was mounted into a nested route, doing history.push('/') un-mounts the entire react app loaded there. This means the whole context gets wiped._

locales/translations/es.json
```diff
@@ -547,9 +548,8 @@
     "The system only accepts CSV files": "El sistema sólo acepta archivos CSV",
     "This question is not relevant for partial flag": "Esta pregunta no es relevante para alcanzar la marca parcial",
     "This question is relevant for partial flag": "Esta pregunta es relevante para alcanzar la marca parcial",
-    "This simulation has ended": "La simulación ha finalizado",
-    "This simulation is ended. Please refresh to start a new one": "La simulación ha finalizado. Actualice para comenzar una nueva",
-    "This simulation is expired. Please refresh to start a new one": "La simulación expiró. Actualice para comenzar una nueva",
+    "This simulation is ended": "La simulación ha finalizado",
```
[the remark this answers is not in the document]
ysbaddaden: I think it used to be "has ended" but it somehow got changed to "is ended" at some point, probably when specifying "please refresh". I didn't touch it... thought, yes, I'd use "has ended" myself.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
ysbaddaden: @ftarulla This is the actual behavior we want: to wipe the current context, so React will recreate it. Without this hack, React won't reload the context.

There is probably a better solution, maybe call a method that would reset the simulation, instead of playing with the router? If you want to take a stab at it, please do! I'm not confident in my React-fu.

ysbaddaden:
@ftarulla I fixed the translations. We have both "simulation is ended" and "simulation has ended" messages. The former is used for the toast, the latter by the mobileweb simulator.

web/static/js/components/simulation/QuestionnaireSimulation.jsx
ftarulla: Yeah! From what I was reading this is a missing part in `react-router` and all solutions are workarounds/hacks so I don't think we could do much about it.

ftarulla (approved):
🍟 🚀

PR #1986, opened 2021-11-02
Upgrade to Ecto 3, Elixir 1.8 + latest Phoenix 1.3

ysbaddaden:
**The road to Ecto 3**:

- [x] remove direct dependency on Timex.Ecto;
- [x] remove usages of deprecated Timex.Ecto.DateTime —replaced with `:utc_datetime`;
- [x] remove usages of deprecated Ecto.DateTime —replaced with `:utc_datetime` for fields (only found in SurveyLogEntry) and DateTime for methods;
- [x] start using DateTime over Timex.DateTime when possible;
- [x] upgrade Ecto from v2.1 to v2.2;
- [x] upgrade Coherence:
  - [ ] :warning: may impact `alto_guisso` (nope),
  - [ ] :warning: may need to also upgrade Elixir (yes -> 1.7) and Phoenix (nope);
  - [x] **alternative**: fix upgrade warnings in Coherence 3.2;
- [x] upgrade phoenix_ecto:
  - [ ] :warning: ~~may require to upgrade to Phoenix 1.4~~ (not needed);
- [x] upgrade ex_machina (>= 2.2.2);
- [x] upgrade to Ecto 3;
- [x] migrate from [mariaex](https://hex.pm/package/mariaex) to [myxql](https://hex.pm/package/myxql) driver (breaks CI & mariaex support was removed in Ecto 3.2)

**Bumps on the road**:
- [x] upgrade to Elixir 1.6 —mostly for the `DateTime.truncate/2` method;
- [x] upgrade to Elixir 1.7 —required by dependencies;
  - [x] upgrade to Sentry 7.x to take advantage of Logger of OTP 21+ (instead of the deprecated `:error_logger`);
- [x] upgrade to Elixir 1.8 —required by Ecto 3.5;
- [x] upgrade to Phoenix 1.3.4 —fixes a compilation warning after upgrading to Elixir 1.6.

**Bugs**:
- [x] The Floip tests exhibit an issue where DateTime's microseconds are being rounded instead of truncated ([MySQL issue](https://bugs.mysql.com/bug.php?id=76948)), which results in the date comparison failing in tests because of a one-second difference —for example: we expect 00:00:09 but we get 00:00:10 instead.

  ~~This should be fixed after upgrading to Ecto 3, since `:utc_datetime` will have seconds precision instead of microseconds~~ That seems to be fixed :crossed_fingers:

PR #1994, opened 2021-11-19
Create enums for Respondent state & disposition

### LINKED ISSUE #1975: Create unique list of valid RespondentState and RespondentDisposition
Currently, the codebase is full of strings defining states and dispositions from scratch, we should define them as constants only once.

ysbaddaden:
Leverages Ecto.Enum to validate respondent state/disposition values and raise errors when trying to use invalid values (e.g. typos, mixed up state vs disposition vs survey state, ...).

- [x] Respondent state enum (see f026354)
- [x] Respondent disposition enum (see b1ac21a)

requires #1986 (Ecto 3.5+)
closes #1975

### DIFF
--- a/test/lib/runtime/flow_test.exs
+++ b/test/lib/runtime/flow_test.exs
@@ -1287,7 +1287,7 @@ defmodule Ask.FlowTest do
       |> test_step(mode)
 
   defp test_step(flow, mode), do:
-    Flow.step(flow, test_visitor(mode), :answer, "contacted")
+    Flow.step(flow, test_visitor(mode), :answer, :contacted)
 
   defp test_reply(flow, mode, nil), do:
     Flow.step(flow, test_visitor(mode), Flow.Message.no_reply, :contacted)
--- a/test/lib/runtime/flow_test.exs
+++ b/test/lib/runtime/flow_test.exs
@@ -247,7 +247,7 @@ defmodule Ask.FlowTest do
     disposition = Reply.disposition(reply)
 
     assert prompts == []
-    assert disposition == "refused"
+    assert disposition == :refused
   end
 
   test "retry step (sms mode)" do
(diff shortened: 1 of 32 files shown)


ysbaddaden:
Note: I can probably go a step further with the respondent state, with the respondent_count also returning atoms (it currently returns strings) as I had to do for the disposition.

web/models/respondent_stats.ex
```diff
@@ -114,6 +114,6 @@ defmodule Ask.RespondentStats do
 
   def respondents_by_disposition(survey) do
     respondent_count(survey_id: ^survey.id, by: :disposition)
-    |> Enum.into(%{})
+    |> Enum.into(%{}, fn ({k, v}) -> {String.to_existing_atom(k), v} end)
```
ftarulla, line 117: Maybe we could use:
```
|> Map.new(fn {k, v} -> { String.to_existing_atom(k), v } end)
```
Just to emphasize that we are creating a new `Map` (instead of using `Enum.into(%{}, ...)`

test/lib/runtime/flow_test.exs
```diff
@@ -1304,7 +1304,7 @@ defmodule Ask.FlowTest do
 
   defp reply_ivr(flow, reply), do: test_reply(flow, "ivr", reply)
 
-  defp test_reply(flow, mode, reply, old_disposition \\ "any_disposition"), do:
+  defp test_reply(flow, mode, reply, old_disposition \\ :contacted), do:
```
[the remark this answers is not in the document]
ysbaddaden: Because we can't use whatever disposition name anymore, it must be a valid value, otherwise we get a runtime error that the disposition doesn't exist.

web/models/respondent_stats.ex
ysbaddaden: @ftarulla I also made `respondents_by_state` above use `Map.new` and `Map.merge` because `Enum.into` is deprecated for non empty enumerables :shrug: 

I also build the default Map from the actual Ecto.Enum. It was missing `:cancelled` for example.

ftarulla (approved):
🍟 🚀

PR #2000, opened 2021-11-26
Fix: report crashed process exceptions to Sentry

ysbaddaden:
Relies on :error_logger to log exceptions happening in processes that live outside Plug.

See https://github.com/getsentry/sentry-elixir/tree/6.4.2#capture-crashed-process-exceptions


matiasgarciaisaia (approved):
👍

🚀 🍟

ysbaddaden:
Once deployed we should start seeing some _new_ errors in Sentry (especially in production). That's a little sad, but how we smash bugs if we don't know about them?

matiasgarciaisaia:
_Seen_ the errors is great as long as the causes of those errors are not new 😇

Issue #2002, opened 2021-11-29
IVR Simulation Hang up flag as breakoff


manumoreira:
In an IVR simulation if the user presses the hangup button it should be flagged as break off, a while after that the simulation will end.

manumoreira:
It looks ok for me, thanks!

PR #2012, opened 2021-12-07
IVR Simulation Hang up as breakoff

ftarulla:
Related #2002 

In an IVR Simulation when choosing hangup button moves disposition to `breakoff`

### DIFF
--- a/lib/ask/runtime/session.ex
+++ b/lib/ask/runtime/session.ex
@@ -141,7 +141,11 @@ defmodule Ask.Runtime.Session do
         update_respondent_disposition(session, "started", current_mode, persist)
     end
     respondent = session.respondent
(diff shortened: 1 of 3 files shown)


ysbaddaden (commented):
After reading the PR, maybe just sending a "STOP" AT message would do the job? We wouldn't have to tweak Ask.Runtime.Flow nor Ask.Runtime.Session just for the IVR simulator.

ysbaddaden (changes requested):
`VoiceWindow` doesn't need the `onCloseSimulation()` prop, so let's remove it, and then LGTM!

```console
$ eslint --ext .jsx,.js web/static/js/ test/js

/app/web/static/js/components/simulation/VoiceWindow.jsx
  108:25  error  'onCloseSimulation' is assigned a value but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

ysbaddaden (approved):

PR #2013, opened 2021-12-07
Fix: cache-control header for AudioController and AudioDeliveryController

ysbaddaden:
The former did set the cache-control header properly, but the latter didn't, leading to poor performance with Twilio.

There were no tests for AudioDeliveryController either, or for the cache-control header being set in AudioController. This is now tested.

### DIFF
--- a/test/controllers/audio_controller_test.exs
+++ b/test/controllers/audio_controller_test.exs
@@ -90,27 +90,20 @@ defmodule Ask.AudioControllerTest do
   end
 
   describe "show" do
-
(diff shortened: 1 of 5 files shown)


matiasgarciaisaia (approved):
💯 Awesome, Julien!

ysbaddaden:
@matiasgarciaisaia this is merged to `develop` which gathers the refactored, and cherry-picked to `master` (current staging/production).

matiasgarciaisaia:
Screen Shot 2021-12-09 at 09 28 19

Works great in staging! 🎉

PR #2015, opened 2021-12-07
Fix: can't export some questionnaires as CSV

### LINKED ISSUE #1995: Quex export CSV isn't working
Quex CSV export is not working properly, the export crashes sometimes and when it doesn't it does not include the randomized sections.

This feature was created long before adding most of the current quex features.

Depending on quex export epic #1988

ysbaddaden:
- [x] Special characters (such as `#`) break the export.
- [ ] ~~SMS prompts with split parts will only export the first part, skip the n+1 parts.~~ Errata: they're actually joined with a special character (`\u001E`)

fixes #1995


ftarulla (approved):
🍟 🚀

PR #2018, opened 2021-12-09
Fix: IVR crash on quex step with refusal disabled

### LINKED ISSUE #2017: IVR simulator ends session instead of recording the current step submission
I continue finding interesting bugs in the new questionnaire I’ve imported to the blue team project:
<https://surveda-stg.instedd.org/projects/343/questionnaires/3112/edit>

While running the IVR simulator it automatically break-off in after step gender. That’s not happening in the original quex (I’ve exported).

Refs #1988

ysbaddaden:
There was a bug in `Ask.Runtime.Step` that considered refusal values for a step where refusal had been disabled. The list of values was empty, so this ended up introducing a `nil` value in the list, and an eventual internal crash.

fixes #2017

### DIFF
--- a/lib/ask/runtime/step.ex
+++ b/lib/ask/runtime/step.ex
@@ -139,7 +139,7 @@ defmodule Ask.Runtime.Step do
     end
   end
 
-  def fetch(:num_digits, step, "ivr", _language) do
+  def fetch(:num_digits, step, "ivr", language) do
     case step["type"] do
       "language-selection" ->
         # If we have 9 choices (1..9) then we can set numDigits to 1,
@@ -165,18 +165,12 @@ defmodule Ask.Runtime.Step do
       "numeric" ->
         # Only send numDigits if the min and max values have the same string length,
         # also taking into account the values of refusal responses
-        refusal = step["refusal"]
-
-        values = if refusal do
-          refusal["responses"]["ivr"]
-        else
-          []
-        end
-
         min_value = step["min_value"]
         max_value = step["max_value"]
+        refusal_values = fetch(:refusal, step, "ivr", language) || []
+
         if min_value && max_value do
-          values = [min_value, max_value | values]
+          values = [min_value, max_value | refusal_values]
           |> Enum.map(fn v -> v |> to_string |> String.length end)
           |> Enum.uniq
 
--- a/test/lib/runtime/flow_test.exs
+++ b/test/lib/runtime/flow_test.exs
@@ -202,6 +202,7 @@ defmodule Ask.FlowTest do
         skip_logic: numeric_skip_logic(min_value: 12345, max_value: 56789, ranges_delimiters: "25,75", ranges: []),
         alphabetical_answers: false,
         refusal: %{
+          "enabled" => true,
           "responses" => %{
             "ivr" => ["#", "12"]
           }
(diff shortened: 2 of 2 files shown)


ftarulla:
I've tried setting up a Questionnaire where the first Question is of type `Numeric` with `responses` From 1 To 2 without accepting `refusals` and I cannot get the simulation to crash. Do I need to setup something else? 🤔

lib/ask/runtime/step.ex
```diff
@@ -165,18 +165,12 @@ defmodule Ask.Runtime.Step do
...
-
         min_value = step["min_value"]
         max_value = step["max_value"]
+        refusal_values = fetch(:refusal, step, "ivr", language) || []
+
         if min_value && max_value do
-          values = [min_value, max_value | values]
+          values = [min_value, max_value | refusal_values]
           |> Enum.map(fn v -> v |> to_string |> String.length end)
           |> Enum.uniq
```
ftarulla, line 176: We can improve code in lines 177-181:
```
  if length(values) == 1 do
    hd(values)
  else
    nil
  end
```
with: 
```
List.first(values)
```
Which returns `nil` on `empty` lists
https://hexdocs.pm/elixir/List.html#first/2

ysbaddaden:
@ftarulla as per [this comment](https://github.com/instedd/surveda/issues/2017#issuecomment-990054780): you must enable the refusal feature, add some values, disable the feature, export the questionnaire (ZIP) then import it as a new questionnaire to trigger the bug. It won't happen on the original questionnaire because the refusal values are kept.

In any case, if the feature is disabled, we musn't consider the refusal values.

lib/ask/runtime/step.ex
ysbaddaden: We can't use `List.first/2` because there must be _exactly_ one remaining entry in the list.

ftarulla:
> @ftarulla as per [this comment](https://github.com/instedd/surveda/issues/2017#issuecomment-990054780): you must enable the refusal feature, add some values, disable the feature, export the questionnaire (ZIP) then import it as a new questionnaire to trigger the bug. It won't happen on the original questionnaire because the refusal values are kept.
> 
> In any case, if the feature is disabled, we musn't consider the refusal values.

Great!! I didn't see the step by step for this bug. Thanks!!

lib/ask/runtime/step.ex
```diff
@@ -165,18 +165,12 @@ defmodule Ask.Runtime.Step do
...
-          []
-        end
-
         min_value = step["min_value"]
         max_value = step["max_value"]
+        refusal_values = fetch(:refusal, step, "ivr", language) || []
+
         if min_value && max_value do
-          values = [min_value, max_value | values]
+          values = [min_value, max_value | refusal_values]
```
ftarulla, line 173: Just one more suggestion (because it is misleading at first sight)
If I got it right, the `Enum.map` (in line 174) is counting the digits, so maybe we could re-write it as:
```
values = [min_value, max_value | refusal_values]
|> Enum.map(& (&1 |> Integer.digits |> Enum.count))
|> Enum.uniq
```

An example of the `Integer.digits |> Enum.count` composition
```console
iex(1)> [1, 2, 42] |> Enum.map(& (&1 |> Integer.digits |> Enum.count))
[1, 1, 2]
```

I hope that with this change it makes the code more readable.

ftarulla (approved):

lib/ask/runtime/step.ex
ysbaddaden: It's usually a number, but it doesn't have to, for example it can be `#` or `*`.

Issue #2003, opened 2021-11-29
IVR simulation if no answer repeat questions 3 times then breakoff


manumoreira:
To emulate the real IVR behavior if there is no answer after 3 seconds we should repeat the question 2 more times and if no answer after that flagged as a breakoff

mmuller:
Version b6ec048

If the question is not the first one, the respondent is flagged as a `break-off` as expected, but when there is no answer to the `first` question, the respondent is flagged as a `refused`:

<img width="1284" alt="image" src="https://user-images.githubusercontent.com/518152/154565588-fb424980-a20f-48e2-a08f-09a273b977c0.png">

diegoliberman:
@mmuller shouldn't we create an issue for the bug you mentioned?

ysbaddaden:
Here is the reason: we don't have an endpoint to tell the backend to break-off the simulation, since sending a STOP message did the trick, we used that for now.

@diegoliberman yes, let's create an issue about it!

PR #2023, opened 2021-12-13
IVR simulation: if no answer repeat last question 3 times

ftarulla:
Related #2003 

We repeat the last question 3 times. 
If we have no input from the user then we hangup (resulting in Disposition = `Refused`, the same as if the user hit the hangup button)

### DIFF
--- a/web/static/js/components/simulation/VoiceWindow.jsx
+++ b/web/static/js/components/simulation/VoiceWindow.jsx
@@ -69,12 +69,8 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps> {
   }
 
   play() {
-    console.log("play")
-
     const ivr = this.props.prompts.shift()
     if (ivr) {
-      console.log("ivr")
-
       if (this.playPromise) {
         this.playPromise
           .then(() => this.playIVR(ivr))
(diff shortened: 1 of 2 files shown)


web/static/js/components/simulation/VoiceWindow.jsx
```diff
@@ -74,25 +80,48 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps> {
         this.playIVR(ivr)
       }
     } else {
-      this.spectrum.stop()
```
ysbaddaden, line 77: We should still stop the spectrum, to avoid having a background timer being called at 60fps. `playIVR()` restarts it, and you can restart it just before calling `this.audio.play()` for the repeat.

web/static/js/components/simulation/VoiceWindow.jsx
```diff
@@ -74,25 +80,48 @@ const VoiceWindow = translate()(class extends Component<VoiceWindowProps> {
...
+
+        this.repeatMessageTimer = setInterval(() => {
+          if (this.timesRepeated >= 3) {
+            this.hangUp()
+            return
+          }
+
+          this.audio.play()
+          this.timesRepeated += 1
+        }, 5000)
```
ysbaddaden, line 100: Maybe extract a `repeat()` method? So we could:

```js
if (ivr) {
  // ...
} else {
  this.spectrum.stop()

  if (!this.props.readOnly) {
    this.repeat()
  }
}
```

ysbaddaden: There can be an issue with `setInterval()` because a question may be very long (see a questionnaire in staging), so we should just start a timeout after the question has finished, or aborting if it's the third time.

Issue #2024, opened 2021-12-15
Questionnaire downloads: csv and zip filenames don't match


ftarulla:
When downloading a questionnaire using `Export questionnaire` (.zip format) or downloading its content via  `Download contents as csv` (csv format) the resulting filenames don't match. 
The first option results is the id of the questionnaire (example `42.zip`) 
The second option results in the name of the questionnaire (without spaces) (example `Examplequestionnaire_translations.csv`)

We could try:
- `42-Example_questionnaire.zip`
- `42-Example_questionnaire_translations.csv`

PR #2025, opened 2021-12-15
Fix questionnaire download filenames

ftarulla:
Related #2024 

For a questionnaire with the name "Example questionnaire", the following download-filenames are generated:

- `42-Example_questionnaire.zip` (when Export questionnaire)
- `42-Example_questionnaire_translations.csv` (when Download contents as csv)


ysbaddaden (commented):
What about using a simplified regular expression to allow any unicode letters and numbers, and discard everything else?

- `~r/[^\p{L}\p{N}]/u` in Elixir;
- `/[^\p{L}\p{N}]+/gu` in JavaScript.

ysbaddaden:
Apart from the regex comment, this is :+1:

Issue #2026, opened 2021-12-15
Handling long questionnaire names


ftarulla:
When entering a long questionnaire name the `name input` in the UI is misplaced.
Also we need to check the filenames generated when downloading `.csv` or `.zip` (the filenames are based on the questionnaire name)

- [x] Fix name input
- [x] Review download filenames

PR #2027, opened 2021-12-16
Fix handling long questionnaire names (UI + csv zip filenames)

ftarulla:
Related #2026


ysbaddaden:
I think we don't need that with #2028 properly truncating the title?

ftarulla:
Replaced by #2028

PR #2030, opened 2022-01-04
Fix: dev annoyances

ysbaddaden:
Fixes a few tiny developer annoyances. See individual commits.

### DIFF
--- a/web/static/js/store/configureStore.dev.js
+++ b/web/static/js/store/configureStore.dev.js
@@ -2,5 +2,5 @@ import prodStore from './configureStore.prod'
 import createLogger from 'redux-logger'
 
 export default function configureStore(preloadedState) {
-  return prodStore(preloadedState, [createLogger()])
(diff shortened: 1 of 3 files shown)


web/static/js/store/configureStore.dev.js
```diff
@@ -2,5 +2,5 @@ import prodStore from './configureStore.prod'
 import createLogger from 'redux-logger'
 
 export default function configureStore(preloadedState) {
-  return prodStore(preloadedState, [createLogger()])
+  return prodStore(preloadedState, [createLogger({ collapsed: true })])
```
ftarulla, line 5: Do we need to do the same at `web/static/mobile_survey/js/store/configureStore.dev.js`?

ftarulla (approved):

web/static/js/store/configureStore.dev.js
ysbaddaden: Not needed but :+1:

PR #2032, opened 2022-01-06
Fix: #1471 (restored branch by Mati)

### LINKED ISSUE #1471: Overview not loading in a particular survey
In this survey the overview doesn't load 
https://surveda-mw.org/projects/2/surveys/216

### LINKED ISSUE #1999: Missing bucket error when calculating percentages
A bug reminiscent of #1471 is happening in production servers.

There is a recent example in the LK instance for the survey id=368 (see [SURVEDA-LK-2X](https://sentry.io/organizations/instedd/issues/1654548537/?project=1263912)):

> **function nil.quota/0 is undefined or private**

The bug happens in [Ask.RespondentController](https://github.com/instedd/surveda/blob/master/web/controllers/respondent_controller.ex#L375-L381). We're trying to find a bucket for a group but there is no such bucket, so Surveda crashes when it tries to calculate the group's percentage. I assume we don't validate properly that the configured buckets match the whole range, or we're missing a fallback solution, or a proper error reporting?

ysbaddaden:
I believe #1471 has never been really fixed. I dig into the pull request and commit histories, and couldn't find anything even remotely related. The notifications created by Juan & Brian also seem to indicate that: the issue was moved from 0.22 upto 0.25.0 then removed, with nothing happening.

If the fixes were never merged, it makes sense that it still appears in Sentry (see #1999), though not that frequently, since it's an edge case.

I noticed lost commit references by @matiasgarciaisaia's in #1471 which I could restore — :bow: to GitHub for keeping those commits somewhere (they're not in Git) and returning a patch when adding `.patch` to the commit URL!

I eventually rebased the restored branch against the current `master` branch, and the patches still applied properly (nice).

closes #1471 
closes #1999

### DIFF
--- a/test/js/reducers/survey.spec.js
+++ b/test/js/reducers/survey.spec.js
@@ -525,11 +525,11 @@ describe('survey reducer', () => {
             {'condition': [{store: 'Age', value: [20, 29]}, {store: 'Smokes', value: 'Yes'}], 'quota': 0},
             {'condition': [{store: 'Age', value: [30, 39]}, {store: 'Smokes', value: 'Yes'}], 'quota': 0},
             {'condition': [{store: 'Age', value: [40, 49]}, {store: 'Smokes', value: 'Yes'}], 'quota': 0},
-            {'condition': [{store: 'Age', value: [50, 119]}, {store: 'Smokes', value: 'Yes'}], 'quota': 0},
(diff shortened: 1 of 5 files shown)


ysbaddaden:
Merging. This was written by @matiasgarciaisaia and I rebased/reviewed.

PR #2033, opened 2022-01-10
Use plugs to assign/authorize project (+ some per-controller resources)

### LINKED ISSUE #1974: Use plugs to load/validate current project
We currently load/validate projects manually in each controller action, but we could simplify and DRY things up by having a function plug that would load/validate and assign a project to the conn instead.

ysbaddaden:
Note: based on the [chore/update-to-phoenix-1.4](https://github.com/instedd/surveda/tree/chore/upgrade-to-phoenix-1.4) branch to avoid conflicts.

Avoids repetitive calls to `load_project/2` and `load_project_for_change/2`, while still allowing to access the loaded project from `conn.assigns`.

Introduces a `:append_assigns_to_action` method to Ask.Web to change actions from `action_name/2` to `action_name/3` by passing `conn.assigns` as the third argument. This allows to match against `conn.assigns` just as nicely as we already match against `conn.params` which is the second argument. _This behavior is opt-in for each controller_.

Also adds a few per-controller `assign_resource/2` methods to avoid the same repetitive method calls inside numerous actions of the controller, preferring to access the resource directly from the `conn.assigns`.

fixes #1974


web/controllers/questionnaire_controller.ex
```diff
@@ -21,14 +22,26 @@ defmodule Ask.QuestionnaireController do
     QuestionnaireAction
   }
 
+  plug :assign_project when action in [
+    :index,
+    :show,
+    :export_zip,
+    :start_simulation,
+    :sync_simulation,
+    :get_last_simulation_response,
```
ysbaddaden, line 31: It could be interesting to extract these 3 methods into a QuestionnaireSimulationController.

web/controllers/survey_controller.ex
```diff
@@ -1,14 +1,39 @@
...
+    :set_description,
+    :delete,
+    :launch,
+    :simulate_questionanire,
+    :create_link,
+    :delete_link,
+    :refresh_link,
+    :stop,
+  ]
+  plug :assign_project_for_owner when action in [:update_locked_status]
```
ysbaddaden, line 34: We could extract:
- the 4 simulation methods into a SurveySimulationController
- the 3 link methods into a SurveyLinkController

ysbaddaden:
While doing this refactor, I noticed that some controllers were a little overloaded with external things that could be extracted into their own controller.

I attempted to extract QuestionnaireSimulationController, SurveySimulationController and SurveyLinkController. At worst it would be a good learning material at manipulating routes, controllers and views in Phoenix. It turned out that it did help to refocus the original QuestionnaireController and SurveyController to what they should: CRUD operations around their resources, with the new controllers either manipulating a related resource (e.g. survey links) or executing peripheral actions on questionnaires and surveys (e.g. running simulations).

These refactorings are internal to the backend. The frontend API didn't change.

ysbaddaden:
I also noted that surveys have a state, and we could maybe use an enum to handle it, instead of plain strings?

ysbaddaden:
Hum, the more I think about this PR, the more I believe that the controller extractions are good, but the `:assign_project` plugs are only half baked and the `when actions in [...]` are far too big.

I think I miss [Pundit](https://rubygems.org/gems/pundit) which I _always_ use with Rails, to extract the authorization policies out of controllers and have simple `:load_and_authorize_xyz` filters.

That being said, someone did port Pundit to Elixir: https://hex.pm/packages/pundit
Maybe I should give it a try.

ggiraldez (approved):
Looks good as far as my understanding of the project and the framework goes. Great job!

I like and agree that the controller extractions are the way to go. I'd try to do more of them. The only drawback to the plugs is what @ysbaddaden already pointed out: needing to specify different plugs depending on the action is prone to mistakes on the `when action in ...` conditionals when writing new actions or changing existing ones. Those mistakes should be caught by tests but that's not ideal either. Something like Pundit may help, as it doesn't erase the problem, but makes the intent more explicit.

ysbaddaden:
Thanks for the review @ggiraldez!

I'll extract the controller extractions into their own PR, then rework the load and authorize project plugs, see if Pundit helps alleviate that mixed feeling.

Issue #1962, opened 2021-10-07
Upgrade Phoenix to 1.4


diegoliberman:
Let's invest up to 2 days to migrate.  Expected outputs:
1. If we didn't finish: estimation of how much work is left
2. Estimation of how much work would require the upgrade to Phoenix 1.5.
3. Plan how to continue the upgrade.

ysbaddaden:
QA: merged to preview/upgrades branch #2049

PR #2039, opened 2022-01-18
Upgrade Coherence dependency

### LINKED ISSUE #2036: Upgrade Coherence
During the recent upgrades (#1986 and #2019) it proved easier and much quicker to upgrade our fork of Coherence 0.3 rather than trying to upgrade to the latest Coherence.

For obvious reasons, it's not a good idea to continue this way, and before or in parallel, to upgrading to Phoenix 1.5 it would be interesting to try and assess a proper upgrade to the latest `master` branch (sadly there have been no releases for many years, while `master` did get updates) that already supports Phoenix 1.5.

refs #1963
refs #1965

ysbaddaden:
Upgrades [Coherence](https://github.com/smpallen99/coherence) to the latest `master` on top of #1962

- [x] Upgrade dependencies:
  - [x] upgrade the `coherence` dependency (latest `master`);
    - [x] create migration to add the `unconfirmed_email` column to the `users` table :warning: ;
  - [x] fix [`alto_guisso`](https://github.com/instedd/alto_guisso_ex) to work with the latest `coherence` (see https://github.com/instedd/alto_guisso_ex/pull/1);
  - [ ] upgrade `alto_guisso` dependency when the above PR is merged;
  - [x] update `timex` —we locked v3.3 but `coherence` now requires v3.6;
  - [x] extract `timex` 3.6 upgrade to its own PR;
- [x] Refactor the integrations:
  - [x] rework the current `coherence` integration with custom controllers;
  - [x] fix the `alto_guisso` integration with the new `coherence` controllers (use plugs);
  - [x] rework the current `coherence` integration with custom layout;
- [x] Manual tests (aka local QA):
  - [x] Views are using the expected layout;
  - [x] I can login & logout using Guisso;
  - [x] I can register using Guisso _and_ a pending invite is translated to a project membership;
  - [x] Pending invite is translated to a project membership after registration through Guisso;
  - [x] I can login, logout & reset password (Guisso disabled);
  - [ ] I can register & confirm (Guisso disabled);
  - [ ] Pending invite is translated to a project membership (Guisso disabled);
membership;
- [x] Write unit tests to verify:
  - [x] the custom redirections;
  - [x] the `alto_guisso` integration;
  - [x] that pending invitations are transformed to project memberships after a successful registration :warning: 

closes #2036


test/controllers/coherence/session_controller_test.exs
```diff
@@ -0,0 +1,107 @@
...
+    disable_guisso()
+
+    conn = get conn, session_path(conn, :new)
+    assert html_response(conn, 200)
+  end
+
+  describe "oauth_callback" do
+    # setup do
+    #   enable_guisso()
+    # end
```
devduarte, line 29: Is this commented for any specific reason? If its not being executed I would just delete it. What do you think?

test/controllers/coherence/session_controller_test.exs
```diff
@@ -0,0 +1,107 @@
...
+    #   end)
+
+    #   conn = get conn, session_path(conn, :oauth_callback), code: "CODE", state: "STATE"
+    #   assert redirected_to(conn, 302) == "/"
+
+    #   # it created the user and added her to the project (and deleted the invite)
+    #   assert user = Repo.one(from u in User, where: u.email == "someone@ask.dev")
+    #   assert Repo.one(from m in Ask.ProjectMembership, where: m.user_id == ^user.id and m.project_id == ^project.id)
+    #   refute Ask.Invite |> Repo.get(invite.id)
+    # end
```
devduarte, line 79: If this a test case that we should be considering, we can create a tech-debt and tackle it later, if it is not then I would just delete it. But I'm not sure if it makes sense to have a whole test case commented. Maybe we can add this code to the new issue to use it as a guide?

test/controllers/coherence/session_controller_test.exs
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  describe "authenticated" do
+    # setup do
+    #   disable_guisso()
+    # end
```
devduarte, line 85: Kind of the same here about commented code.

lib/ask/runtime/session.ex
```diff
@@ -79,9 +79,10 @@ defmodule Ask.Runtime.Session do
         # do not retry since the respondent was never contacted, thus the retries should not be consumed
         session = contact_respondent(session, runtime_channel)
 
+        now = DateTime.utc_now
         next_available_date_time = session.schedule |> Schedule.next_available_date_time
 
-        base_timeout = Interval.new(from: DateTime.utc_now, until: next_available_date_time)
+        base_timeout = Interval.new(from: now, until: next_available_date_time)
```
devduarte, line 85: How about leaving `DateTime.utc_now` ? Are we storing it in a variable just for improving readability a bit or for any other reason I might be missing?

devduarte (changes requested):

lib/ask/runtime/session.ex
ysbaddaden: This is actually a fix to a bug that is introduced by the upgrade to Timex 3.6 (if I recall correctly). We must retain the `now` value _before_ we called `next_available_date_time` otherwise we end up with `from > until` which now fails.

I should add a comment about this!

ysbaddaden: The code was duplicated, so I extracted it and I added a comment explaining the reason!

devduarte (approved):
:shipit:

PR #2042, opened 2022-01-26
GHA: avoid duplicated runs and useless builds

ysbaddaden:
Avoids the duplicated run on `push` and `pull_request` events for internal branches. Also avoids building the Docker image unless we really want one: main, release or preview branches, and release tags.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,15 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+    branches:
+      - "**"
+
+  pull_request:
+    branches:
+      - "**:**"
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


ysbaddaden:
This PR should be the perfect example: only one test run and no build :crossed_fingers:

.github/workflows/ci.yaml
```diff
@@ -23,13 +32,15 @@ jobs:
           docker-compose run --rm webpack yarn eslint
 
   build:
-    needs: test
-    if: github.event_name == 'push'
     runs-on: ubuntu-latest
+    needs: test
+    if: contains('refs/heads/main', github.ref) || startsWith(github.ref, 'refs/heads/release') || startsWith(github.ref, 'refs/heads/preview') || startsWith(github.ref, 'refs/tags/')
```
matiasgarciaisaia, line 37:
```suggestion
    if: contains('refs/heads/main', github.ref) || startsWith(github.ref, 'refs/heads/release/') || startsWith(github.ref, 'refs/heads/preview/') || startsWith(github.ref, 'refs/tags/')
```

.github/workflows/ci.yaml
```diff
@@ -1,6 +1,15 @@
...
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+"
+    branches:
+      - "**"
+
+  pull_request:
+    branches:
+      - "**:**"
```
matiasgarciaisaia, line 12: I fear this is always getting more verbose.

Code reuse is hard in workflows, isn't it?

matiasgarciaisaia (approved):
Looks good! 👍 

I'd rather add the trailing slashes on the branch names we check, to avoid a `previews-in-app-links` branch triggering a preview image 😇

.github/workflows/ci.yaml
ysbaddaden: Yes! Hopefully the list of events won't change much.

But it was much simpler with CircleCI (thanks to `&`). Or maybe we must change how we look at it, and must write actual actions, that could be shared across repositories?

matiasgarciaisaia: Maybe something in the spirit of manastech/ci-docker-builder but that also checks whether to run the tests or not.

Like, having the CI-provider-specific workflow be "get this script, always run these tests, then build me a docker image", and have the script (shared between most repos) decide when to run the workflow (depending on branch names, tag names, etc).

🤷

ysbaddaden: Yes. Exactly!

Issue #1965, opened 2021-10-07
Upgrade Elixir Libraries


diegoliberman:
Try to upgrade the different Elixir dependencies of Surveda to, if not their latest, at least a supported version.

This is a long-running **EPIC** issue to eventually bring Surveda up to date.

PR #2045, opened 2022-02-07
Upgrade Timex to v3.6

ysbaddaden:
Extracted from #2036 to have a distinct commit in main for this upgrade instead of squashing it _with_ the Coherence upgrade. it's touching on time & timezones, and can have some impact.

Note: we don't need Timex much since Date, Time and DateTime in stdlib evolved _a lot_ since Elixir 1.0. Even timezones are properly handled, now. The only feature we need from Timex is shifting non-naive DateTime objects by an arbitrary granularity, for example adding 2 days or 1 hour, because stdlib only allows to add seconds.

refs #1965


ysbaddaden:
@devduarte I extracted the Timex upgrade and applied your suggestions from #2036.

PR #2050, opened 2022-02-08
Refactor: extract controllers and simplify SurveyController and RespondentController

ysbaddaden:
Extracted from #2033 to only keep the controller extractions, and the simplification of the SurveyController and RespondentController, as previously discussed with @ggiraldez.


diegoliberman:
@devduarte do you have this one in your radar?

PR #2062, opened 2022-04-08
Truncate long messages in IVR simulator

### LINKED ISSUE #2053: Add ellipsis on step text detail in the mobile phone emulator [IVR Simulator]
Currently, when a step is reproduced in the simulator it also shows the text detail of the step in the mobile phone emulator

Screen Shot 2022-03-02 at 14 11 54
Screen Shot 2022-03-02 at 14 12 00

When the text exceeds 140 characters or 6 lines the emulator automatically scrolls down to show the last characters of the step.

To enhance the UX we should add a clickable ellipsis after 140 chars or 6 lines, this way the user will always see the first lines of the step by default and only scroll down to the last lines after clicking on the ellipsis.

ysbaddaden:
Truncates long message in the IVR simulator, and adds an ellipsis.

It also doesn't automatically scroll to the bottom when expanding a message, only when sending and receiving the message.

Maybe the message bubble should have the `cursor: pointer` CSS property when the message can be expanded? Maybe it could have an explicit `(click to expand)` or alike message?

closes #2053

### DIFF
--- a/web/static/js/components/simulation/MessagesList.jsx
+++ b/web/static/js/components/simulation/MessagesList.jsx
@@ -12,20 +12,48 @@ type MessagesListProps = {
   truncateAt: ?number,
 }
 
-type MessagesListState = {
-  toTruncate: Array<boolean>,
+type MessageProps = {
+  message: ChatMessage,
+  truncateAt: ?number,
 }
 
-export class MessagesList extends Component<MessagesListProps, MessagesListState> {
-  messagesBottomDivRef: any
+type MessageState = {
+  truncated: boolean,
+}
 
+class Message extends Component<MessageProps, MessageState> {
   constructor(props) {
     super(props)
 
     this.state = {
-      toTruncate: [],
+      truncated: props.truncateAt && props.message.body.length > props.truncateAt,
     }
   }
+
+  render() {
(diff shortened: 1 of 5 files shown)


web/static/js/components/simulation/MessagesList.jsx
```diff
@@ -50,33 +39,42 @@ export class MessagesList extends Component<MessagesListProps> {
...
+    if (this.mustTruncate()) {
+      messages = this.truncate(messages)
+    }
 
     return (
       <div className="chat-window-body">
         <ul>
-          {groupedMessages.map((messages, ix) => (
-            <MessageBulk key={`msg-bulk-${ix}`} messages={messages} />
+          {messages.map((message, index) => (
```
ysbaddaden, line 64: I'm dropping the `MessageBulk` component because it didn't add much, and actually required a bunch of useless indirection. 

BUT now that I'm writing this, maybe I should have have keep it, and put the individual `truncated` state there, instead of a using a `truncated` array in MessagesList. Hum :thinking: yeah, that would be better :facepalm: 

I also dropped grouping the subsequent AO and AT messages, since it led to invalid DOM (UL > DIV > LI) and I failed to notice any visual difference?

PR #2067, opened 2022-04-11
Migrate to Phoenix 1.3 directory structure

### LINKED ISSUE #2061: Migrate to new Phoenix 1.3 project structure
Surveda still uses the Phoenix project structure from Phoenix 1.2 and before.

As we upgrade Phoenix, the difference is more and more noticeable and confusing. Trying to upgrade Coherence, for example, was challenging partly because of that project structure change.

The new project is detailed in https://www.phoenixframework.org/blog/phoenix-1-3-0-released

This should be mostly moving files around (from `web` to `lib/ask_web`) and to rename some modules (from `Ask.Web` to `AskWeb`).

A simple way to know what to do is to create a brand new `Ask` project with Phoenix 1.4, add coherence, and see what files have been created and which modules have been renamed, and to make the changes in Surveda.

Since assets currently live inside `web/assets` we'll probably have to update the webpack configuration, too. See https://phoenixframework.org/blog/phoenix-1.4.0-released (they merely put assets into `/assets`)

ysbaddaden:
Finishes the migration to Phoenix 1.3 by moving files and some modules from the current directory structure to the new directory structure, which is expected by Phoenix and dependencies from now on.

It helps to clarify a little the modules: we still have Ask (under `/lib/ask`) but now have AskWeb (under `/lib/ask_web`) that replaces Ask.Web and is much more widely used.

Models were moved right under `/lib/ask`. It's a little weird to no longer have `models` directory, but it makes sense since they're under `Ask` module. ~~One oddity though, is that we still `use AskWeb, :model` inside models. It looks like newly generated models don't use that anymore (or it was only done in Surveda :shrug:) I kept it because it seemed useful.~~ I added Ask.Model and changed models to `use Ask.Model` !

Also moves the `/web/static` directory as `/assets` since this is the new place for assets. The `/web/static/assets` directory was renamed as `/assets/static`.

Some modules from `/web`, such as Ask.Metrics have been kept under the Ask module, and thus moved into `lib/ask` to better reflect which module they belong to.

closes #2061


ysbaddaden:
@aeatencio I pushed a commit that adds Ask.Model :)

aeatencio:
> Models were moved right under /lib/ask. A little weird to no longer have models directory, but it makes sense since they're under Ask module. One oddity though, is that we still use AskWeb, :model inside models. It looks like newly generated models don't use that anymore (or it was only done in Surveda 🤷) I kept it because it seemed useful.

Two thoughts:

1. `AskWeb, :model` was replaced by `Ask.Model` 👏  => We should update the PR description to express that.
2. I didn't get what "newly generated models don't use that anymore" means. Can't we keep a folder for models? Maybe using `lib/ask/models`, instead of placing them all in `lib/ask`?

aeatencio:
> @aeatencio I pushed a commit that adds Ask.Model :)

I saw it! Nice!!! 🤩

test/models/short_link_test.exs
```diff
@@ -1,5 +1,5 @@
-defmodule Ask.ShortLinkTest do
-  use Ask.ModelCase
+defmodule AskWeb.ShortLinkTest do
+  use Ask.DataCase
```
aeatencio, line 2: What is this renaming for?

aeatencio:
I think the `test` directory may be outdated. It should follow the shape of the folder it tests. Most of the files should be inside `lib/ask` and `lib/ask_web`.

test/models/short_link_test.exs
ysbaddaden: This is what Phoenix uses from now on :shrug:

ysbaddaden:
Yeah, I should see what Phoenix' generators do now.

test/models/short_link_test.exs
ysbaddaden: I think it goes with the rest of the refactoring: Phoenix doesn't seem to want models anymore, but data stores, abstracted through contexts. It may also be to avoid thinking the ModelCase is meant to test models, when it's meant to help testing modules _with_ models (I need data)?

ysbaddaden:
@aeatencio I moved the tests around... within `test/ask` or `test/ask_web`!

aeatencio (approved):
Every change I read seems to follow the goal of the migration. It looks like a very well-done job.

Nevertheless, because the very big number of files changed (so many files were renamed and little changes were made everywhere), I'd do some QA before merging this.

ysbaddaden:
AFAIK: Elixir compiles the application and doesn't complain about any missing module or function, so I don't expect any issues with the renames. Elixir did complain while I made the changes, with some modules moving between `Ask` and `AskWeb`, leading to fix some alias/import statements.

aeatencio:
> AFAIK: Elixir compiles the application and doesn't complain about any missing module or function, so I don't expect any issues with the renames. Elixir did complain while I made the changes, with some modules moving between `Ask` and `AskWeb`, leading to fix some alias/import statements.

As we talked about today, let's merge it, then! 🎉

ysbaddaden:
Commits manually cherry-picked to `main`.

PR #2069, opened 2022-04-12
IVR simulator breakoff on first question

### LINKED ISSUE #2052: [Bug] Respondent incorrectly flagged as refused
Reported [here](https://github.com/instedd/surveda/issues/2003#issuecomment-1043470498).

If the question is not the first one, the respondent is flagged as a `break-off` as expected, but when there is no answer to the `first` question, the respondent is flagged as a `refused`:

image

ysbaddaden:
The IVR simulator UI now acts more as a channel, following the Verboice channel it now sends a `timeout` message to the simulator, instead of abusing the `stop` message that didn't work as expected in some cases (e.g. the disposition becomes `rejected` when it should be `unresponsive`).

This change, along with the proper support of the `no_reply` flow message in the IVR simulator allowed to go one step further, and to remove the whole repeat question logic in the UI, because it's already properly handled by the session flow running on the backend. The IVR simulator is now behaving closer to reality :crossed_fingers: 

Note: I made a first commit to extract the IVR, MobileWeb and SMS simulators into their own files, which makes reading and hacking on the simulators a little less messier to follow :face_with_spiral_eyes: 

fixes #2052


aeatencio:
> I made a first commit to extract the IVR, MobileWeb and SMS simulators into their own files, which makes reading and hacking on the simulators a little less messier to follow.

Thank you **so much** @ysbaddaden for doing it!

assets/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -197,18 +197,20 @@ class QuestionnaireSimulation extends Component<Props, State> {
   handleATMessage = (message) => {
     const { projectId, questionnaireId, mode } = this.props
     const { simulation } = this.state
-    if (simulation) {
+    let reply
```
aeatencio, line 200: The following sounds a little odd to me: `reply` can take its value from `message` or `message.body` depending on the context. The type of `message` depends on the context? Or the type of `reply` depends on the context?
We may want to add some sort of type check or at least a comment to clarify the expected types.

assets/js/components/simulation/QuestionnaireSimulation.jsx
```diff
@@ -197,18 +197,20 @@ class QuestionnaireSimulation extends Component<Props, State> {
...
+
+    if (!simulation) return
+
+    if (message.body) {
```
aeatencio, line 204: I see from below (`onSendMessage`) that the `message` type can be `Message` or `String`. This still seems a little odd to me. I'm not sure about how it could be improved. Maybe considering a new type of Message? Something like `{ body: "timeout", type: "failure" }`, and then handling them differently based on the `message.type` ?

test/lib/runtime/questionnaire_simulator_test.exs
```diff
@@ -442,6 +442,78 @@ defmodule QuestionnaireSimulatorTest do
     end
   end
 
+  defp ivr_questionnaire_with_steps(steps) do
+    questionnaire_with_steps(steps)
+    |> Questionnaire.changeset(%{modes: ["ivr"]})
+    |> Repo.update!()
+  end
+
+  describe "IVR - timeout message" do
```
aeatencio: These tests make lots of sense to me. Nicely done! Also, reading them made me think we have some technical debt testing the whole IVR simulator. These new tests are the only ones testing it, right? Do we have an open issue for solving this debt?

assets/js/components/simulation/VoiceWindow.jsx
```diff
@@ -124,51 +135,37 @@ const VoiceWindow = translate()(
...
+    stopSimulation(): void {
+      this.audio.pause()
+      this.cancelAnswerTimeout()
+      this.cancelMessageTimer()
     }
 
-    repeatLastAudio(): void {
-      if (!this.audio) return
+    // Simulates a phone hangup by sending a STOP message. This is kinda
+    // hackish, we could probably send a proper hangup message that would be
```
aeatencio, line 152: Agree. Also because using stop here may be confused with the SMS concept of "stop mo" message. Maybe we could have a new type "conn" and then messages like { body: "hangup", type: "conn" } for this case and { body: "timeout", type: "conn" } or something similar for an unanswered call?

aeatencio (approved):
You fixed it with a lot of class! I leave you this modest review so you can consider some thoughts I had while doing it. But feel free to merge it. The code seems to work much better this way!

test/ask/respondents_filter_test.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Ask.RespondentsFilterTest do
```
aeatencio, line 1: Ups, sorry, @ysbaddaden, I think I changed this (in this file and others) by error while merging changes from `main` into this branch 🤦 . Do you want to review this, or do you prefer I fix it before merging it?

test/lib/runtime/questionnaire_simulator_test.exs
ysbaddaden: There are enough tests, at least for the IVR and SMS simulators, but they're _very_ repetitive. We could extract some helper methods and heavily simplify each test, improving the readability!

lib/ask/runtime/questionnaire_ivr_simulator.ex
```diff
@@ -125,6 +121,25 @@ defmodule Ask.Runtime.QuestionnaireIvrSimulator do
...
+  defp append_ao_message(simulation, nil) do
+    simulation
+  end
+
+  defp append_ao_message(simulation, reply) do
+    messages = simulation.messages ++ AOMessage.create_all(reply)
+    Map.put(simulation, :messages, messages)
+  end
+
+  defp handle_at_message(simulation, "timeout") do
```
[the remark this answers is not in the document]
ysbaddaden: For the backend `message` is a `String.t()`, hence it can either be one or more digits (0-9) or a special message (e.g. timeout), and it looks like the proper way to handle the situation in Erlang/Elixir, and it helped reduce the complexity in the calling functions.

assets/js/components/simulation/QuestionnaireSimulation.jsx
ysbaddaden: You're right. I don't like it either. There is a `Message|string` type but it feels wrong. The backend doesn't care about the type: it only receives a message (the `message.body`).

Let's make `message.type` optional? So we'd have `{ body: "1", type: "at" }` and `{ body: "timeout" }`, oh and we can also have `{ body: "stop" }` to avoid logging the stop message :)

test/lib/runtime/questionnaire_simulator_test.exs
aeatencio: I don't see them. For IVR, I only see the tests you wrote in this PR (all of them covering the `timeout` cases). Please help me: what tests am I missing?

assets/js/components/simulation/VoiceWindow.jsx
ysbaddaden: I made some failed attempts, but I finally found a mean to implement a proper hangup! Please see my newest commit @aeatencio

lib/ask/runtime/questionnaire_ivr_simulator.ex
```diff
@@ -110,8 +110,12 @@ defmodule Ask.Runtime.QuestionnaireIvrSimulator do
     end
 
     sync_simulation = fn simulation, response ->
-      {simulation, reply} = handle_at_message(simulation, response)
-      QuestionnaireSimulator.sync_simulation(simulation, reply, "ivr", handle_app_reply)
+      case handle_at_message(simulation, response) do
```
ysbaddaden, line 113: Maybe using `handle_at_message` isn't the best solution here. Maybe just checking `if response == "hangup"` would suffice?

aeatencio: If you ask this because of [this conversation](https://github.com/instedd/surveda/pull/2069#discussion_r858751321), considering the type is String, I agree with you [here](https://github.com/instedd/surveda/pull/2069#discussion_r858946573), the current version of the code is the best Elixir way of expressing this conditional. I changed my mind. I wouldn't change it 🙂

aeatencio: On the other hand, I'm not sure if "hangup" is an at_message. Maybe you asked because you think it isn't? Are `response` and `at_message` different concepts? I like that question! 🙃

aeatencio: Sorry! I was reading an old version of your code 🤦 . I'll review it again!

aeatencio: Something like this ?

```elixir
if response == "hangup" do
  QuestionnaireSimulator.stop_simulation(simulation, handle_app_reply)
else
  {simulation, reply} = handle_at_message(simulation, response)
  QuestionnaireSimulator.sync_simulation(simulation, reply, "ivr", handle_app_reply)
end
```

For me, yes, I read it easier. But it's your call, I think we shouldn't spend too much more time here. Both versions are good enough for me. Thanks for asking!

assets/js/components/simulation/VoiceWindow.jsx
aeatencio: > we must first locate & understand how are hangups handled in the Verboice channel

I'm not there yet 😞 . But I'm just wondering if this comment still applies 👇 

    // Simulates a phone hangup by sending a STOP message. This is kinda
    // hackish, we could probably send a proper hangup message that would be
    // properly handled by the simulator.

test/lib/runtime/questionnaire_simulator_test.exs
ysbaddaden: Oh, you're right, we're barely testing the IVR mode! We're not even testing the audio prompts (argh). I created #2074!

PR #2070, opened 2022-04-15
Fix: reset respondent sanitized phone number

### LINKED ISSUE #2055: Sanitized phone numbers not reset when changing to a channel without patterns
When we start contacting a respondent via a channel, we apply the channel's pattern to make the `sanitized_phone_number` match the one that the channel will use.

This is the relevant code:

https://github.com/instedd/surveda/blob/b6ec048fc804383efaf1ce3e656a901e5d0e5688/lib/ask/runtime/session.ex#L456-L466

When there's a multi-modal survey in which the first mode has a pattern and the second one does not, the `sanitized_phone_number` for the second mode remains unchanged (ie, with the pattern from the mode channel applied).

If there's no pattern, we should reset the `sanitized_phone_number` to the `canonical_phone_number`.

ysbaddaden:
Merely overwrites the sanitized phone number in the database instead of doing nothing. If another channel applied some patterns previously, then the new attempt with this channel will be reset to use the original phone number.

fixes #2055

### DIFF
--- a/lib/ask/runtime/session.ex
+++ b/lib/ask/runtime/session.ex
@@ -589,7 +589,7 @@ defmodule Ask.Runtime.Session do
   end
 
   defp run_flow(%{current_mode: current_mode, respondent: respondent} = session, persist) do
-    respondent = apply_patterns_if_match(current_mode.channel.patterns, respondent)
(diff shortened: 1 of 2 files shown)


matiasgarciaisaia (approved):
LGTM! 👍

aeatencio (approved):

PR #2075, opened 2022-05-02
Add daily success rate chart

### LINKED ISSUE #2054: Add chart for daily success rate in survey overview
The success rate of a survey can have a wide range of values during a survey, it will be helpful to visualize the changes in this rate throughout the survey. 
As the survey overview is quite crowded with charts currently we might want to reorganize them to get this issue done.
- An option can be to include this into the big line chart for quotas completion at the top of the screen.
[Mockup](https://projects.invisionapp.com/d/main#/console/8374235/241966468/preview)

ysbaddaden:
Calculates and reports the daily success rate to the frontend, and renders a line to the main survey chart.

I'm reporting a new `percentages` property to hold the success rate daily history, instead of adding it to the current `cumulative_percentages` property, because it's not a cumulative percentage (always grows over time) but a daily success rate (goes up and down).

I'm also not meddling with the `references` property because it describes the different cumulative percentages, and trying to add an `{ id: "success_rate", label: "Success rate" }` to it led the AskWeb.RespondentController to inject dummy values for it into `cumulative_percentages`. I thus hardcoded the addition of the line & label to the frontend, which seems acceptable because unlike the other references that are specific to the survey, this one is static and always displayed.

closes #2054

### DIFF
--- a/lib/ask/survey.ex
+++ b/lib/ask/survey.ex
@@ -584,6 +584,7 @@ defmodule Ask.Survey do
   def completion_rate(_, 0), do: 0.0
   def completion_rate(completed, respondents_target), do: completed / respondents_target
 
+  # Calculates the current success rate since the survey was launched.
   def get_success_rate(survey, respondents_by_disposition) do
     completed_respondents = get_completed_respondents(survey, respondents_by_disposition)
 
@@ -591,6 +592,42 @@ defmodule Ask.Survey do
       exhausted_respondents(respondents_by_disposition, survey.count_partial_results)
 
     success_rate(completed_respondents, exhausted_respondents)
+  end
+
+  # Calculates the daily sucess rate since the survey was launched until it
+  # ended or today if it's running.
+  def success_rate_history(survey) do
+    completed_dispositions = Respondent.completed_dispositions(survey.count_partial_results)
+    completed = survey |> respondents_history(completed_dispositions)
+
+    exhausted_dispositions = Respondent.metrics_final_dispositions(survey.count_partial_results)
+    exhausted = survey |> respondents_history(exhausted_dispositions)
+
+    running_date_range(survey)
+    |> Enum.map(fn date ->
+      {date, completion_rate(completed[date], exhausted[date]) * 100}
+    end)
+  end
+
+  # Counts respondents with specified dispositions with daily granularity.
+  defp respondents_history(survey, dispositions) do
+    survey
+    |> assoc(:respondents)
+    |> select({fragment("DATE(completed_at)"), fragment("COUNT(*)")})
+    |> group_by(fragment("DATE(completed_at)"))
+    |> where([r], r.disposition in ^dispositions)
+    |> Repo.all()
+    |> Enum.into(%{})
+  end
+
+  # Generates the actual running date range, from the day the survey started to
+  # the day it ended (or today if it's still running).
+  defp running_date_range(%Survey{started_at: started_at, ended_at: nil}) do
+    Date.range(started_at |> DateTime.to_date(), Date.utc_today())
+  end
+
+  defp running_date_range(%Survey{started_at: started_at, ended_at: ended_at}) do
+    Date.range(started_at |> DateTime.to_date(), ended_at |> DateTime.to_date())
   end
 
   def get_completed_respondents(survey, respondents_by_disposition) do
--- a/lib/ask/survey.ex
+++ b/lib/ask/survey.ex
@@ -582,6 +582,7 @@ defmodule Ask.Survey do
 
   def completion_rate(_, nil), do: 0.0
   def completion_rate(_, 0), do: 0.0
+  def completion_rate(nil, _), do: 0.0
   def completion_rate(completed, respondents_target), do: completed / respondents_target
 
   # Calculates the current success rate since the survey was launched.
@@ -596,6 +597,8 @@ defmodule Ask.Survey do
 
   # Calculates the daily sucess rate since the survey was launched until it
   # ended or today if it's running.
+  def success_rate_history(%{started_at: nil}), do: []
+
   def success_rate_history(survey) do
     completed_dispositions = Respondent.completed_dispositions(survey.count_partial_results)
     completed = survey |> respondents_history(completed_dispositions)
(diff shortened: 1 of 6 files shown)


lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -563,6 +564,38 @@ defmodule AskWeb.RespondentController do
...
+  end
+
+  # Cleanup repetitive values, by skipping any repetitive value except for the
+  # first and last, because they don't contribute anything to the graph
+  # (straight line).
+  #
+  # For example take A, B, C, D consecutive dates, each at value 5.0 then we
+  # only keep A and D and skip B and C entirely.
+  #
+  # TODO: consider moving to RespondentView.
```
ysbaddaden, line 582: Maybe this ought to be in RespondentView? The controller is doing the job in the controller, so I made it here.

lib/ask/survey.ex
```diff
@@ -593,6 +595,44 @@ defmodule Ask.Survey do
...
+      {date, completion_rate(completed[date], exhausted[date]) * 100}
+    end)
+  end
+
+  # Counts respondents with specified dispositions with daily granularity.
+  defp respondents_history(survey, dispositions) do
+    survey
+    |> assoc(:respondents)
+    |> select({fragment("DATE(COALESCE(completed_at, updated_at))"), fragment("COUNT(*)")})
+    |> group_by(fragment("DATE(COALESCE(completed_at, updated_at))"))
```
ysbaddaden, line 620: I started by using `completed_at` but it seems that it used to not be defined, and most of the test suite (written in 2016) don't define the column, and rely on `updated_at` instead. There is even a test that we don't crash when the disposition is completed but `completed_at` is nil!

test/ask/survey_test.exs
```diff
@@ -207,4 +207,44 @@ defmodule Ask.SurveyTest do
       assert created_survey.panel_survey_id == panel_survey.id
     end
   end
+
+  describe "success_rate_history" do
+    test "reports until today" do
+      started = DateTime.utc_now() |> Timex.shift(days: -5)
+      date = DateTime.to_date(started)
```
aeatencio, line 214: nitpick: maybe `start_date` is a better name for the `date` variable?

test/ask/survey_test.exs
```diff
@@ -207,4 +207,44 @@ defmodule Ask.SurveyTest do
...
+               {date |> Timex.shift(days: 3), 0.0},
+               {date |> Timex.shift(days: 4), 0.0},
+               {date |> Timex.shift(days: 5), 0.0}
+             ]
+    end
+
+    test "reports until survey ends date" do
+      started = DateTime.utc_now() |> Timex.shift(days: -10)
+      ended = started |> Timex.shift(days: +6)
+      date = DateTime.to_date(started)
```
aeatencio, line 233: same as above

test/ask/survey_test.exs
```diff
@@ -207,4 +207,44 @@ defmodule Ask.SurveyTest do
       assert created_survey.panel_survey_id == panel_survey.id
     end
   end
+
+  describe "success_rate_history" do
```
aeatencio, line 211: I think these tests would be more complete if they included examples where the values weren't always `0.0`. Do you agree?

lib/ask/survey.ex
aeatencio: Yes, this sounds odd but at the same time, it sounds familiar to me 🤷 . Maybe you can put this comment in the code, just to increase these lines' readability.

lib/ask_web/controllers/respondent_controller.ex
aeatencio: Please let me take advantage of your dilemma here. How would you decide this? Which considerations would you take into account to decide? Thanks!!!

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -563,6 +564,38 @@ defmodule AskWeb.RespondentController do
...
+  defp cleanup_repetitive_percentages(percentages, result \\ [])
```
aeatencio, line 583: Please take the following comment totally optative. I'm pretty sure this code is fine as it was written. But I'm wondering while reading it: Do we need to program these operations? I'm just not sure. Maybe we do, but I see them too far from our domain, and pretty generic. Maybe we can take a similar logic from an external library, or push this logic outside this module, something like `cleanup_repetitive_values(list)` or something like that? What do you think?

aeatencio:
The names `cumulative_percentages` and `percentages` aren't enough descriptive IMO. I'd use something like `completion_percentages` and `success_rate_percentages`, or `completion_cumulative_percentages` and `success_rate_historical_percentages`, depending on how much descriptive (and long!) we want to write them.

aeatencio (approved):
Nice job!!! 🎉 LGTM, I leave you some comments.

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: I didn't write this cleanup at first, and it was fine. But:

1. the methods for generating the cumulative percentages do such a cleanup, albeit a little differently (it does so while it sums the daily data, and only cares about straight lines);
2. the testsuite has _running_ surveys started in 2016, so we end up with over 2000 entries (:astonished:) in the JSON and it was impractical, so I did the cleanup :shrug:

I don't think we have other places where we cleanup historical chart data, so I think it's acceptable to have it here, especially since it's tied to how the data is structured (unless we decide to move it to the view).

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: I didn't move the cleanup because the controller already cleans up the similar `cumulative_percentages` to the view, because it does that while doing the sum (not as a separate step), so it felt proper to send it the cleaned up array, even though logically it should be in the view :shrug:

test/ask/survey_test.exs
ysbaddaden: Yes, but this is already tested at a higher level (the controller), so I'm a little lazy to generate actual data for these tests. In fact, I only needed them to write the model method, and maybe we don't need these tests anymore?

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: But I fully agree: it's not really at its place here. Maybe it ought to be moved to the View object.

test/ask/survey_test.exs
aeatencio: I'm not sure. I wouldn't remove them. I think I'd prefer to have at least a couple of simpler and more descriptive (unit) tests here. These tests don't pretend too much, but they explain part of the new behavior. The tests in `RespondentControllerTest` are so many (4893 lines) and so complex. I'd love to have some smaller, simpler, and more descriptive tests here rather than there, not only for coverage reasons (I guess the coverage is OK) but for readability purposes too.

I think we're OK regarding the scope of this issue. Maybe in the future, we can refactor/rethink our test suite regarding its coding style/design?

aeatencio:
As I said, this LGTM! Nicely done! Thanks for your comments, Julien!

PR #2079, opened 2022-05-11
Added enum type for survey state

### LINKED ISSUE #2035: Create unique list of valid SurveyState
Surveys have a state, and we could maybe use an enum instead of plain strings to limit the scope to a fixed set of possible values.

This is related to #1975 (see #1994).

leandroradusky:
Closes #2035 

Surveys' state type changed from string to Ecto.Enum. 

The main change is then:

```
field :state, :string, default: "not_ready"
```
to 

```
field :state, Ecto.Enum, values: [
   :not_ready,
   :ready,
   :pending,
   :running,
   :terminated,
   :cancelling,
 ], default: :not_ready
```

The rest of the changes are replacement of the strings to atoms where it corresponds. 

All test are passing, both for `mix test` and `yarn test`, manual testing on the website was done and it appears to work. 

DB migration files are also affected, now atoms are passed (strings where tested and they work as well). I.e: `/priv/repo/migrations/20160913180125_set_default_state_for_surveys.exs` which before look like this:

```
 def up do
   alter table(:surveys) do
     modify :state, :string, default: "pending"
   end
 
   flush()
 
   from(s in Ask.Survey, where: is_nil(s.state))
   |> Ask.Repo.update_all(set: [state: "pending"])
 end
```

and now looks like this:

```
 def up do
   alter table(:surveys) do
     modify :state, :string, default: :pending
   end
 
   flush()
 
   from(s in Ask.Survey, where: is_nil(s.state))
   |> Ask.Repo.update_all(set: [state: :pending])
 end
```


test/ask_web/controllers/folder_controller_test.exs
```diff
@@ -163,7 +163,7 @@ defmodule AskWeb.FolderControllerTest do
                    "name" => wave.name,
                    "description" => nil,
                    "project_id" => project.id,
-                   "state" => "not_ready",
+                   "state" => to_string(:not_ready),
```
ysbaddaden, line 166: I'd keep the string. Be them params sent to the controller or transformed to JSON, it's a mere `"not_ready"` string. It should just work. Same below, as well as in ShortLinkControllerTest and SurveyControllerTest.

lib/ask/channel.ex
```diff
@@ -49,8 +49,8 @@ defmodule Ask.Channel do
   end
 
   # Deletes a channel and:
-  # - marks related "ready" surveys as "not_ready"
-  # - marks related "running" surveys as "terminated" with exit code 3
+  # - marks related :ready" surveys as :not_ready
```
ysbaddaden, line 52:
```suggestion
  # - marks related :ready surveys as :not_ready
```

priv/repo/migrations/20160913180125_set_default_state_for_surveys.exs
```diff
@@ -4,13 +4,13 @@ defmodule Ask.Repo.Migrations.SetDefaultStateForSurveys do
 
   def up do
     alter table(:surveys) do
-      modify :state, :string, default: "pending"
+      modify :state, :string, default: :pending
```
aeatencio, line 7: Do we need to change the old migration files? Do the changes in the model force us to do it? If that's the case, OK, maybe it's our only option. In any case, I feel odd doing it. I'd always try first to add new migration files with the required changes.

ysbaddaden: If we want to run the migrations using current code, yes, we probably do.

For this specific case of a default SQL value, we probably could be using a string.

ysbaddaden: But just below we should be using the atom.

Issue #2088, opened 2022-06-01
Honor the channel limit added in channel settings


aeatencio:

manumoreira:
Work item deleted in Commit

PR #2089, opened 2022-06-02
Add capacity setting to channels

### LINKED ISSUE #2082: Add channel limit as a field in surveda channel settings
Currently in surveda>channel>settings we are showing a Verboice iframe that looks this way
image
This can be misleading and dangerous for surveda users
We should remove this iframe (or disable it at least) and **add 1 field for channel capacity.**
The textbox should have a hint saying: `Number of parallel contact attempts`
It should have a default value we can set at 100 for now.

aeatencio:
Back-end sibling: #2088

Fix #2082

channel capacity tab


ysbaddaden (approved):
That looks good to me!

Issue #356, opened 2016-11-30
Curate list of languages for language autocomplete


mverzilli:
The current list contains some non-sensical items and lots of unnecessary duplicates.

nditada:
We need to find a better curated list.

<img width="286" alt="Screen Shot 2019-09-17 at 12 35 46 PM" src="https://user-images.githubusercontent.com/1359189/65056691-b428f800-d947-11e9-9b52-5d15eaa232d7.png">

manumoreira:
If we filter by living in Language_Type and not blank in Part2B it generates a reasonable list of languages.
https://docs.google.com/spreadsheets/d/1CmoViP1SCiAZ9HtZQjUn8kj0bq2KIw-bAFf4cyLOMQg/edit?usp=sharing

waj:
Part2T is the code that matches the iso-639-3. We're using this library: https://github.com/wooorm/iso-639-3

Let's filter by: `type == "living" && iso6392T != null`

mmuller:
Working properly in version d7ff690

PR #2090, opened 2022-06-03
Update AddLanguage.jsx

manumoreira:
Delete filter for code set for 639-2/B not null to add more languages like Lugbara, Teso, Ganda, Nkore, and Nyankole

### DIFF
--- a/assets/js/components/questionnaires/AddLanguage.jsx
+++ b/assets/js/components/questionnaires/AddLanguage.jsx
@@ -33,7 +33,7 @@ class AddLanguage extends Component {
       lang.text.indexOf(value) !== -1 || lang.text.toLowerCase().indexOf(value) !== -1
 
     let languagesOptions = iso6393
-      .filter((lang) => lang.type == "living" && lang.iso6392T != null)
+      .filter((lang) => lang.type == "living")
       .map((lang) => ({ id: idForLanguange(lang), text: lang.name }))
 
     // Don't show languages that are already selected


matiasgarciaisaia (commented):
I was wandering if the null check was added for something code-wise, but it seems to have been a decision about how to filter-out languages ([comment](https://github.com/instedd/surveda/issues/356#issuecomment-540117379)).

It'd be nice to run the code to actually check nothing else breaks, though.

assets/js/components/questionnaires/AddLanguage.jsx
```diff
@@ -33,7 +33,7 @@ class AddLanguage extends Component {
       lang.text.indexOf(value) !== -1 || lang.text.toLowerCase().indexOf(value) !== -1
 
     let languagesOptions = iso6393
-      .filter((lang) => lang.type == "living" && lang.iso6392T != null)
+      .filter((lang) => lang.type == "living")
       .map((lang) => ({ id: idForLanguange(lang), text: lang.name }))
```
matiasgarciaisaia, line 37: From [the mapping function](https://github.com/instedd/surveda/blob/9b405132d00d5c849d5081055ebfcd09a0c61ea3/assets/js/components/questionnaires/AddLanguage.jsx#L31), it shows we're not using the `iso6392T` for anything else than trimming down the languages list (exactly what we're reverting with this PR), so the null-check doesn't change a thing in the app's behaviour.

matiasgarciaisaia (approved):
Looks good! 👍 :shipit:

aeatencio (approved):
Working OK in my local 👍

Issue #2087, opened 2022-05-30
Experiment with channelBroker as a proxy


manumoreira:
Add a broker between SurveyBroker and channels

PR #2092, opened 2022-06-08
Draft a channel broker as a proxy

aeatencio:
This changes just add a layer that mirrors the channel interface in the context of #2087. Here the experimentation phase is ending. We'll continue working on more specific tasks being defined in #2095.


aeatencio:
Hi @ysbaddaden! I have a couple of questions regarding this draft.

How do you see the direction this draft is taking? I started by adding a proxy between the channels and its client (the module that actually makes the callbacks to Verboice and Nuntium). Then I moved the proxy to the channels and finally, I put it between the session and the channels.

Is the proxy reasonably placed in the current (3rd) iteration?

Does it make sense to continue copying the entire channel interface (all the protocol, functions, and parameters) regarding the current experiment OR I'm missing building its own protocol (and functions) for the ChannelBroker.

Thanks for taking a look!

ysbaddaden:
From my minimal knowledge that should do it.

I'm wondering whether we really need to have a distinct implementation for each channel type? Do you foresee that we'll need this?

aeatencio:
It may be worth continuing this work for #2103

PR #2093, opened 2022-06-14
Feature/2081 enhance channel label in survey

### LINKED ISSUE #2081: Enhance channel label in survey
Currently, if a channel is shared among different users in Verboice each of them can add it to Surveda. Then in the channel selector at the survey they will all look the same (exact same name), this will confuse the user.
We should add the owner's username (email) to the channels list so each channel will look different.
Each channel will have a capacity defined and the channel broker will use that.
This solution only helps the user to identify the channels properly, every Verboice channel can be added as many times as users with the necessary permissions are.

image

leandroradusky:
Closes #2081.

On survey creation, the drop-down listing of the project channels now includes the email of the channel owner (in Surveda). This will help the users correctly identify the channels, as described in the issue, avoiding potential identical items within the list. 

The dropdown now looks like this:

Screensh
 
A channel with a very very long name was created to test how the elements display. As can be seen, they display in multiple lines properly.

To accomplish it, the `ChannelController` was modified. Now, when a channel is returned, it comes with an `userEmail` attribute. Test needed to be modified accordingly. 

A question arose: within `channel_view.ex` there is a mixture of camel case and snake case attributes. I went with camel case for `userEmail`. I can easily change it, not sure what is correct there.

### DIFF
--- a/assets/js/components/surveys/RespondentsList.jsx
+++ b/assets/js/components/surveys/RespondentsList.jsx
@@ -154,7 +154,7 @@ const NewRespondentChannel = (
           <option value="">{t("Select a channel...")}</option>
           {channels.map((channel) => (
             <option key={channel.id} value={channel.id}>
-              {channel.name + channelFriendlyName(channel)}
+              {channel.name + channelFriendlyName(channel) + " (" + channel.userEmail + ")"}
             </option>
           ))}
         </Input>
--- a/lib/ask_web/views/channel_view.ex
+++ b/lib/ask_web/views/channel_view.ex
@@ -21,7 +21,7 @@ defmodule AskWeb.ChannelView do
       settings: channel.settings,
       patterns: channel.patterns,
       status_info: channel.status,
-      userEmail: channel.user_email
+      user_email: channel.user.email
     }
   end
 end
--- a/lib/ask_web/views/channel_view.ex
+++ b/lib/ask_web/views/channel_view.ex
@@ -20,7 +20,8 @@ defmodule AskWeb.ChannelView do
       channelBaseUrl: channel.base_url,
       settings: channel.settings,
       patterns: channel.patterns,
-      status_info: channel.status
+      status_info: channel.status,
+      userEmail: channel.user_email
     }
   end
 end
--- a/lib/ask_web/controllers/channel_controller.ex
+++ b/lib/ask_web/controllers/channel_controller.ex
@@ -9,6 +9,9 @@ defmodule AskWeb.ChannelController do
       |> load_project(project_id)
       |> assoc(:channels)
       |> Repo.all()
+      |> Enum.map(fn channel ->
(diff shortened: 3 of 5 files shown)


lib/ask_web/views/channel_view.ex
```diff
@@ -20,7 +20,8 @@ defmodule AskWeb.ChannelView do
       channelBaseUrl: channel.base_url,
       settings: channel.settings,
       patterns: channel.patterns,
-      status_info: channel.status
+      status_info: channel.status,
+      userEmail: channel.user_email
```
aeatencio, line 24: I'd follow [this pattern](https://github.com/instedd/surveda/blob/01cd890c1f6031a0644044523a2311b43eea84b1/lib/ask_web/views/respondent_group_view.ex#L28) to get the user email from the channel:

```elixir
channel = Repo.get(Channel, 1) |> Repo.preload(:user)
```

With the suggested changes, you'll have 2 `Repo.preload`, one inside the query in the `channel_controller` and the other here. Here, the `Repo.preload` is suggested just for completeness, because the channel will be fetched with its user preloaded (because of the `Repo.preload` in the `channel_controller`. That's OK, it won't be loaded twice because by default Ecto won't preload associations that are already loaded. If this function is called from other place, and the user isn't preloaded by its caller, the line above will do it.

And then, you will be able to access the user email this way: `channel.user.email`

aeatencio, line 24: Regarding the casing, we normally follow [this snake_case convention](https://hexdocs.pm/elixir/main/naming-conventions.html#casing) in the backend in elixir files and then the standard javascript camelCase convention in the front-end. We use the `camelize` function of the `humps` library to convert the keys [here](https://github.com/instedd/surveda/blob/01cd890c1f6031a0644044523a2311b43eea84b1/assets/js/api.js#L617).

aeatencio (approved):
LGTM! 🎉

lib/ask_web/views/channel_view.ex
ysbaddaden: Elixir follows Ruby: `CamelCase` are constants (e.g. modules), `lower_cased` are functions and variables, and `snakeCase` should not exist on the backend (only in JavaScript).

assets/js/components/surveys/RespondentsList.jsx
```diff
@@ -154,7 +154,7 @@ const NewRespondentChannel = (
           <option value="">{t("Select a channel...")}</option>
           {channels.map((channel) => (
             <option key={channel.id} value={channel.id}>
-              {channel.name + channelFriendlyName(channel)}
+              {channel.name + channelFriendlyName(channel) + " (" + channel.userEmail + ")"}
```
ysbaddaden, line 157:
```suggestion
              {channel.name + channelFriendlyName(channel) + " (" + channel.user_email + ")"}
```

PR #2121, opened 2022-07-20
Verify that the Surveda build is broken

aeatencio:
The base Docker image was updated from Debian Stretch to Buster
https://github.com/erlang/docker-erlang-otp/commit/fd21a3b#diff-bba869fab39f9ce4816fb8f2ce9d4635751cae7c386ab70341ec08356a7dbe23

This upgrade breaks the Surveda build with the following error:
`E: Package 'mysql-client' has no installation candidate`
https://github.com/instedd/surveda/runs/7436387784?check_suite_focus=true#step:3:66

This implies replacing the MySQL client with the MariaDB client
https://stackoverflow.com/a/57049357

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN sed -i '/^mozilla\/DST_Root_CA_X3/s/^/!/' /etc/ca-certificates.conf && updat
 
 RUN \
   apt-get update && \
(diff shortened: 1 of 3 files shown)


matiasgarciaisaia (approved):
Looks good. Let's just remove the extra commit.

PR #2143, opened 2022-08-25
Honor channel capacity update on change

### LINKED ISSUE #2142: Update the ChannelBroker state capacity on end-user change
The UI changes the channel settings, but not the channel broker state.

Present in [b3c36903](https://github.com/instedd/surveda/commit/b3c36903)

Step to reproduce the bug:

1. Launch Survey A using Channel A. The current Channel A capacity is X. X is honored.
2. Modify the Channel A capacity. Set capacity Y. 
3. Launch Survey B using Channel A. The old X capacity is honored.

Expected behavior:

1. Launch Survey A using Channel A. The current Channel A capacity is X. X is honored.
2. Modify the Channel A capacity. Set capacity Y. 
3. Launch Survey B using Channel A. The new Y capacity is honored.

---

The above are just some arbitrary steps to reproduce the bug. There's no need to have different surveys running to reproduce it. If the channel is active, then the capacity won't be honored just after modifying it. It will eventually happen when the ChannelBroker process is started from scratch.

aeatencio:
Fix #2142

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -84,6 +84,13 @@ defmodule Ask.Runtime.ChannelBroker do
     check_status(0, respondent, respondent_state, provider)
   end
 
+  def on_channel_settings_change(channel_id, settings) do
+    call_gen_server(
+      channel_id,
+      {:on_channel_settings_change, settings}
+    )
+  end
+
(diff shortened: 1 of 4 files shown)


ysbaddaden (approved):

diegoliberman:
@aeatencio this can be merged, right?

PR #2160, opened 2022-10-10
Fix: can't import huge CSV of respondents

ysbaddaden:
Only fixes the initial load of respondents for a Survey using a CSV of phone numbers.

The following situations will still fail:
- loading a CSV of respondents (not phone numbers).
- adding the CSV of phone numbers twice (import once then add again).

### DIFF
--- a/lib/ask/runtime/respondent_group_action.ex
+++ b/lib/ask/runtime/respondent_group_action.ex
@@ -126,12 +126,28 @@ defmodule Ask.Runtime.RespondentGroupAction do
       |> Enum.chunk_every(64_000)
       |> Enum.flat_map(process_batch)
 
-    # And then remove them from phone_numbers (because they are duplicates)
-    # (no easier way to do this, plus we expect `existing_numbers` to
-    # be empty or near empty)
-    Enum.reject(phone_numbers, fn num ->
-      Respondent.canonicalize_phone_number(num) in existing_numbers
-    end)
+    if Enum.empty?(existing_numbers) do
+      # no duplicates
+      phone_numbers
+    else
+      # And then remove them from phone_numbers (because they are duplicates)
+
+      # the following is basically doing this, but optimized (to handle large
+      # lists of respondents):
+      # Enum.reject(phone_numbers, fn num ->
+      #   Respondent.canonicalize_phone_number(num) in existing_numbers
+      # end)
+
+      # zip into map %{ canonical_number => phone_number }
+      zip =
+        canonical_numbers
+        |> Enum.zip(phone_numbers)
+        |> List.foldl(%{}, fn ({k, v}, zip) -> Map.put(zip, k, v) end)
+
+      existing_numbers
+      |> List.foldl(zip, fn (n, zip) -> Map.delete(zip, n) end) # remove duplicates (using canonical number)
+      |> Map.values() # return the de-duplicated list of phone numbers
+    end
   end
 
   def take_sample(loaded_entries) do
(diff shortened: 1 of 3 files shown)


lib/ask/runtime/respondent_group_action.ex
```diff
@@ -291,18 +291,21 @@ defmodule Ask.Runtime.RespondentGroupAction do
...
-
-    # Restore the initial entries order
+    # keeps the initial entries array in order
+    # OPTIMIZE: is it required to keep the order? it would be magnitudes faster if we didn't have to
     Enum.map(entries, fn entry ->
-      Enum.find(loaded_entries, fn %{phone_number: phone_number} = loaded_entry ->
-        entry == phone_number or entry == Map.get(loaded_entry, :hashed_number)
-      end)
+      if String.starts_with?(entry, "r") do
+        Enum.find(phone_numbers_from_respondent_ids, fn %{
```
ysbaddaden, line 300: It could be interesting to use `Enum.delete` so on each iteration we'd have one less entry in the array... but I don't know how the Elixir runtime/Erlang VM react to inner list deletions?

ysbaddaden: Alternatively if we wouldn't need to retain the original order, the whole function would become:

```elixir
    respondent_ids = Enum.filter(entries, fn entry -> String.starts_with?(entry, "r") end)
    phone_numbers = Enum.filter(entries, fn entry -> not String.starts_with?(entry, "r") end)
    phone_numbers_from_respondent_ids = phone_numbers_from_respondent_ids(survey, respondent_ids)

    phone_numbers
    |> Enum.map(fn phone_number -> %{phone_number: phone_number} end)
    |> Enum.concat(phone_numbers_from_respondent_ids)
    |> Enum.filter(&(!is_nil(&1)))
    |> Enum.uniq_by(fn %{phone_number: phone_number} -> keep_digits.(phone_number) end)
```

aeatencio (approved):

ysbaddaden:
As explained in #2180 there is no need to retain the order of the original file, the validation doesn't even use it anymore to detect the lines that error, it merely iterates the original entries and verifies if we found the respondents.

ysbaddaden:
I still have an issue trying to re-upload the list of respondents. I'm searching where it's failing then.

ysbaddaden:
The hang issue on reupload is caused by the deduplication of phone numbers. There is a rogue `Enum.reject` with an inner call to `number in existing_numbers` which takes longer to process as the list of numbers grows bigger (and is eventually killed).

This is exactly the same issue that was initially fixed.

Using a Map instead of a List seems to make it snappy again.

test/ask_web/controllers/respondent_group_controller_test.exs
```diff
@@ -813,17 +813,16 @@ defmodule AskWeb.RespondentGroupControllerTest do
...
       respondents = Repo.all(from r in Respondent, where: r.survey_id == ^survey.id)
+      phone_numbers = respondents |> Enum.map(& &1.phone_number)
 
       assert length(respondents) == 15
 
       assert group
       assert group.respondents_count == 15
-      assert group.sample == respondents |> Enum.take(5) |> Enum.map(& &1.phone_number)
+      assert Enum.count(group.sample) == 5
+      assert Enum.count(group.sample -- phone_numbers) == 0
```
ysbaddaden, line 823: Elixir maps don't retain any order, so the original test is no longer working (we can't know which samples have been selected out of the 15). I'm thus checking that the samples come from the actual list of respondents.

TODO: we must still verify that the phone numbers have been masked!

leandroradusky (commented):
Looks good to me :)

PR #2186, opened 2023-02-10
Fix/2173 only accept int numeric answers

### LINKED ISSUE #2173: Only accept integers for numeric answers
Last year we've got some complicated answers on SMS mode.
[See thread](https://manas.slack.com/archives/C3KKQKHLG/p1667407241518149)
We realized that (for some reason) we are accepting non-integer answers for numeric steps. We are parsing the answers to make them "valid" and this is generating more problems than benefits.
We should only accept integers for numeric steps.

leandroradusky:
Close #2173.

In Elixir, 

```
iex(2)> Integer.parse("22.1")
{22, ".1"}
iex(3)> Float.parse("22.1")  
{22.1, ""}
```

Then, the methods `is_numeric_permissive` & `is_numeric` were accepting float values and passed them through. 

While before it was continuing with the survey, now:

image

**Note:** is easy to round the integer part and get it: I moved forward accepting integers strictly, and returning `false` otherwise (and not the number rounded to an integer). 

**Note 2:** this was happening only in SMS mode (it couldn't happen in phone channels ─an error is thrown upon responses containing `#` or `*`─, neither in the mobile web mode, where ─at least in the simulator─ the validation is performed well.

### DIFF
--- a/lib/ask/runtime/step.ex
+++ b/lib/ask/runtime/step.ex
@@ -279,9 +279,11 @@ defmodule Ask.Runtime.Step do
   end
 
   defp is_numeric_permissive(str, language, step) do
-    case Float.parse(String.trim(str)) do
-      {num, _} ->
-        if Float.parse(String.trim(str)) == Integer.parse(String.trim(str)) do
+    str = String.trim(str)
+
+    case Float.parse(str) do
+      {num, rest} ->
+        if Integer.parse(str) == {num, rest} do
           round(num)
         else
           false
--- a/lib/ask/runtime/step.ex
+++ b/lib/ask/runtime/step.ex
@@ -279,9 +279,13 @@ defmodule Ask.Runtime.Step do
   end
 
   defp is_numeric_permissive(str, language, step) do
-    case Integer.parse(String.trim(str)) do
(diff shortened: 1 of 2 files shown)


lib/ask/runtime/step.ex
```diff
@@ -281,10 +281,10 @@ defmodule Ask.Runtime.Step do
   defp is_numeric_permissive(str, language, step) do
     case Float.parse(String.trim(str)) do
       {num, _} ->
-        if round(num) == num do
+        if Float.parse(String.trim(str)) == Integer.parse(String.trim(str)) do
```
ysbaddaden, line 284: We can avoid calling `String.trim` and `Float.parse` twice times:

```elixir
str = String.trim(str)

case Float.parse(str) do
  {num, rest} ->
    if Integer.parse(str) == {num, rest}
```

ysbaddaden (commented):
Well found! Could you add tests in `test/ask/runtime/flow_test.exs`? There are a number of examples under `describe "numeric steps"`, but having a few tests around floats would help.

ysbaddaden (approved):

PR #2189, opened 2023-02-20
Fix: validate actual audio type + always transcode to MP3

### LINKED ISSUE #2175: Validate audio files in questionnaire editor
Currently the quex editor accepts any file with mp3 or wav extension. There's no validation of the file format
The wav to mp3 conversion works but without validation we usually endup having multiple issues with audio files. 
Steps:
1. Validate if the uploaded files are OK
2. Convert to mp3 if wav
3. Validate if conversion to mp3 was successful

ysbaddaden:
Relies on magic numbers to detect the actual file type of uploaded audio files, instead of relying on the file extensions (`.wav` and `.mp3`) that may be wrong. Validates the file types using the `file` command line tool, made to return the actual mime type (see `Ask.FileInfo`). Tells `sox` to identify the input file types using magic numbers (see `Ask.Sox`).

Transcodes all audio files to MP3 with a single channel at 44.1khz (even MP3 files) to make sure that the stored MP3 file is proper and not some weirdly encoded MP3 file. That should improve the portability of the audio files.

fixes #2175

### DIFF
--- a/lib/ask/audio.ex
+++ b/lib/ask/audio.ex
@@ -46,7 +46,6 @@ defmodule Ask.Audio do
 
     {%{}, types}
     |> cast(data, Map.keys(types))
-    #|> validate_change(:filename, &validate_file_type/2)
     |> validate_change(:mime_type, &validate_mime_type/2)
     |> validate_change(:size, &validate_size/2)
   end
--- a/lib/ask/audio.ex
+++ b/lib/ask/audio.ex
@@ -17,7 +17,10 @@ defmodule Ask.Audio do
     timestamps()
   end
 
+  @valid_extensions ~w(.mp3 .wav)
+  @valid_mime_types ~w(audio/mpeg audio/wave audio/wav audio/x-wav audio/x-pn-wav)
   @stored_audio_extension "mp3"
+
   def exported_audio_file_name(uuid), do: uuid <> ".#{@stored_audio_extension()}"
 
   @doc """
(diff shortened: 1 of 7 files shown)


lib/ask/audio.ex
```diff
@@ -17,7 +17,10 @@ defmodule Ask.Audio do
     timestamps()
   end
 
+  @valid_extensions ~w(.mp3 .wav)
+  @valid_mime_types ~w(audio/mpeg audio/wave audio/wav audio/x-wav audio/x-pn-wav)
```
matiasgarciaisaia, line 21: Do we care about these ones? As long as we get something that can get converted to a valid MP3, maybe we don't care what that something was?

lib/ask/file_info.ex
```diff
@@ -0,0 +1,20 @@
...
+  @doc """
+    Retrieves informations about one or multiple files, using the `file` command
+    line tool which is shipped with most linux distributions.
+  """
+  @spec get_info(Path.t() | [Path.t()]) :: %{Path.t() => String.t()}
+  def get_info(names)
+  def get_info(name) when is_binary(name), do: get_info([name])
+
+  def get_info(names) when is_list(names) do
+    {result, 0} = System.cmd("file", ["--mime-type" | names])
```
matiasgarciaisaia, line 11: I wouldn't add a dependency on yet another external tool (and parsing its output) if we are any way going to ask Sox to detect the type and convert it itself.

matiasgarciaisaia (commented):
Code looks OK. I'm hesitant about trying to detect file's MIME types, but if you still believe it's the best way - it's mergeable 👍

lib/ask/audio.ex
ysbaddaden: Yes, because there are mimetypes that `sox` doesn't support, and also because we probably don't want to support all possible audio formats.

lib/ask/file_info.ex
ysbaddaden: I noticed afterwards that we use `soxi` in specs, but it does exactly the same (relies on `libmagic`, and the `file` tool is more ubiquitous (I checked it on ubuntu/gnu and alpine/musl) and the output was identical.

ysbaddaden:
@matiasgarciaisaia Since we're dropping support for `sox` in #2190 in favor of `ffmpeg`, I replaced the use of `soxi` in AudioChecker to use `file --mime-type`.

FFmpeg will verify & validate the actual file content, but I think it still makes sense to verify it beforehand. We don't want to support whatever audio input formats.

matiasgarciaisaia (approved):
> We don't want to support whatever audio input formats.

I'm not really sure about that - what are we trying to prevent with that? Unless there's any kind of attack (I'm thinking something analog to a zip bomb) that could harm the system by allowing any extension (ie, that the accepted audio formats are not vulnerable to), it may make sense - but supporting any input format sounds nice otherwise.

Anyway, PR's good to go 👍

ysbaddaden:
@matiasgarciaisaia My concern is the future: having to support whatever audio formats as soon as someone relies on it. That may broke because we changed something on the server, like a patent issue and it's removed from Debian... having a fixed list seems preferable.

PR #2190, opened 2023-02-20
Add support for AAC audio files (ADTS, M4A, MP4)

### LINKED ISSUE #2179: Convert different audio formats to mp3
CDC has requested converting files from these formats
- mp4
- MPEG-4
- ADTS
To mp3

ysbaddaden:
Adds `ffmpeg` to add support for AAC because `sox` doesn't support this audio codec nor the associated file formats (M4A, ADTS) directly. It used to support them (through `ffmpeg` but support got broken and it was eventually removed).

We thus transcode twice: once to WAV using `ffmpeg` then to MP3 using `sox`.

closes #2179 
depends on #2189


lib/ask/sox.ex
```diff
@@ -23,7 +23,42 @@ defmodule Ask.Sox do
     end
   end
 
+  # Sox doesn't support AAC directly, we must first transcode to PCM using
+  # ffmpeg then transcode to MP3 using Sox. Maybe we could transcode to MP3
+  # directly from ffmpeg but I couldn't find the arguments to set encoding,
+  # channels or rate in ffmpeg.
```
matiasgarciaisaia, line 29: My thoughts when reading the PRs description. I guess it'd be easier to drop Sox in favour of ffmpeg? And do a single transcode.

I guess generating a mono audio with the desired bitrate/quality should be enough - I don't think we need to provide guarantees on the encoding.

matiasgarciaisaia:
```
ffmpeg -i "input_file.mp4" -ac 1 -ar 44100 -b:a 64k output_file.mp3
```

`-ac 1` is "mono", `-ar 44100` is sample rate at 44.1khz, `-b:a 64k` is 64kbps constant bitrate.

https://superuser.com/a/253470

ysbaddaden: Oooh, nice, I'll look into that!

ysbaddaden:
@matiasgarciaisaia now with ffmpeg only to transcode all input files into mono, 44.1khz but 128k CBR files which is what we use to encode to MP3 the captured audio on the frontend:

https://github.com/instedd/surveda/blob/240a9cc1ef1cb06a87988e8af897eda5fdac019f/assets/js/components/questionnaires/RecordAudio.jsx#L35-L37

matiasgarciaisaia (approved):
🚀 :shipit:

PR #2191, opened 2023-02-24
Feature/2172 admin pannel for project

### LINKED ISSUE #2172: Create an admin panel for project
The projects control questionnaires and surveys in Surveda, so it would be helpful from the users' perspective to control some variables at the project level.
- Default timezone (this will apply to all the surveys in the project)
- Initial success rate (this is now controlled by 3 environmental variables that work at the instance level) -     We should provide 2 ways for this
      - Input initial success rate
      - Input values for the 3 variables (eligibility rate, response rate, valid respondent rate)
- Name
- Color schema is not a priority, if it's complicated we can leave it for a future version

[Mockups](https://invis.io/M4135750P9HW)

leandroradusky:
Closes #2172.

New panel to define project settings. 

image

It differs from the [mockup](https://projects.invisionapp.com/share/M4135750P9HW#/screens/471694278) in the following:
- There is no button for deleting the project. Instead, the user can archive/unarchive the project (discussed with @manumoreira)
- The settings are a tab (it wasn't present in the mockup). 
- Also the option of changing color scheme was removed from the tabs since now is included in the project settings.

Functionality that was 

Some code comments:
- The `TimezoneAutocomplete` now has the `onChange` property that receives the callback function that should be called, this is in order to make it work with any other component which includes it, since before it was implemented only to work with surveys. 
- All the rates are `number` inputs that allow between 0 and 1 since they are rates.
- New fields for projects are set to `null`, and, unless the user saves settings for the project, the rates continue to be read from the config files (cc @manumoreira)
- There is a need to preload projects for surveys in some places in order to get the rates.

It was tested that:
- New surveys take project default timezone if defined, old behavior otherwise.
- Initial success rates are taken from the project if defined, from config files otherwise.

Extra notes:
- The test are failing in GitHub, but passing for me locally, is it a matter of connection to MySQL? Not sure what is happening. They say `Elixir.Ask.Swoosh.Adapters.Test does not exist` which is not something that I have touched. I should remove `structure.sql` from the commit? As far as I remember when migrations are included the new structure has to be part of the commit. 
- There aren't new tests since nothing was tested with respect to the rates in the tests before. 
- For project settings, I'm getting these two warnings and I don't know how the get rid of them, since I don't find that I'm doing what the warnings say and there is no link to the code:
image


ysbaddaden:
Thanks @leandroradusky! I'm taking over to check on the React warnings and the Elixir failure on CI, which is... odd.

> Elixir.Ask.Swoosh.Adapters.Test does not exist

Don't worry about that, it's a missing adapter for the mailer ([swoosh](https://hex.pm/packages/swoosh)), and I never got to fix it. When I tried... tests starting failing :shrug:

lib/ask/project.ex
```diff
@@ -25,7 +30,17 @@ defmodule Ask.Project do
...
+      :colour_scheme,
+      :timezone,
+      :archived,
+      :initial_success_rate,
+      :eligibility_rate,
+      :response_rate,
+      :valid_respondent_rate
+    ])
     |> validate_colour_scheme
   end
```
ysbaddaden, line 45: We should validate the rates (float between 0.0 and 1.0), and maybe the timezone (it should exist)?

assets/js/components/projects/ProjectSettings.jsx
```diff
@@ -0,0 +1,366 @@
...
+    this.setState({ detailedRates: !detailedRates })
+    this.configureRates(!detailedRates)
+  }
+
+  configureRates(detailedRates) {
+    if (detailedRates) {
+      document.getElementById("initialSuccessRate").getElementsByTagName("input")[0].disabled = true
+      document.getElementById("eligibilityRate").style.display = "block"
+      document.getElementById("responseRate").style.display = "block"
+      document.getElementById("validRespondentRate").style.display = "block"
```
ysbaddaden, line 77: You should never have to query the DOM. React handles the props & state and re-renders as needed what needs to be re-rendered.

Use `this.state.detailedRates` in `render()` directly to selectively render the 3 inputs for detailed rates. For example extract the rendering of these into a method, then in the main `render()`:

```jsx
{this.state.detailedRates && this.renderDetailedRates()}
```

 You already nullify the state values below, React will detect that thanks to `this.setState` and re-render.

assets/js/components/projects/ProjectSettings.jsx
```diff
@@ -0,0 +1,366 @@
...
+            <label className="gray-text">Initial success rate</label>
+            <input
+              type="number"
+              step="0.01"
+              min="0"
+              max="1"
+              value={initialSuccessRate}
+              disabled={readOnly}
+              onInput={(e) => this.setState({ initialSuccessRate: e.target.value })}
+              onChange={(e) => this.setState({ initialSuccessRate: e.target.value })}
```
ysbaddaden, line 222: You can parse as float immediately, so you state will already contains floats (not strings):
```js
this.setState({ initialSuccessRate: parseFloat(e.target.value, 10) })
```

Also specifying base 10 never hurts.

assets/js/components/projects/ProjectSettings.jsx
```diff
@@ -0,0 +1,366 @@
...
+              value={eligibilityRate}
+              disabled={readOnly}
+              onInput={(e) => {
+                this.setState({ eligibilityRate: e.target.value })
+                this.updateInitialSuccessRate()
+              }}
+              onChange={(e) => {
+                this.setState({ eligibilityRate: e.target.value })
+                this.updateInitialSuccessRate()
+              }}
```
ysbaddaden, line 241: You might want to extract methods to avoid repetitions?

assets/js/components/projects/ProjectSettings.jsx
```diff
@@ -0,0 +1,366 @@
...
+            }
+            onClick={() => this.saveProjectSettings()}
+          />
+        </div>
+        <div className="col">
+          <input
+            type="button"
+            value="Cancel"
+            className="btn-flat"
+            onClick={() => window.location.reload(false)}
```
ysbaddaden, line 319: We shall be able to optimize that, by restoring the initial state:

```js
this.setState(this.initialState)
this.configureRates(this.state.detailedRates)
```

ysbaddaden (commented):
Here is a first pass of review. Do you want me to handle them @leandroradusky?

I'll look into the warnings & CI failure.

assets/js/components/surveys/SurveyWizardScheduleStep.jsx
```diff
@@ -99,6 +99,10 @@ class SurveyWizardScheduleStep extends Component {
 
     const { startDate, endDate } = survey.schedule
 
+    const changeTimezone = (timezone) => {
+      dispatch(actions.setTimezone(timezone))
+    }
```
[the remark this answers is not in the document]
leandroradusky: This is the first that I've tryed... But, when doing:

```
  changeTimezone(timezone) {
    const { dispatch } = this.props
    dispatch(actions.setTimezone(timezone))
  }
```

I get 

image

Because I'm using it as a callback, but not directly calling the function... Any thoughts on this? Is something I'm doing wrong?

assets/js/components/timezones/TimezoneAutocomplete.jsx
```diff
@@ -124,6 +131,7 @@ class TimezoneAutocomplete extends Component {
               })}
             </span>
           )}
+          <input type="hidden" name="timezone_id" value={this.state.timezone.id} />
```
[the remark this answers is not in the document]
leandroradusky: Dead code, sorry :(

lib/ask/runtime/broker.ex
```diff
@@ -113,7 +113,10 @@ defmodule Ask.Runtime.Broker do
     Enum.filter(all_running_surveys, fn survey ->
       survey.id not in expired_surveys_ids and Schedule.intersect?(survey.schedule, now)
     end)
-    |> Enum.each(fn survey -> poll_survey(survey, now) end)
+    |> Enum.each(fn survey ->
+      survey = survey |> Repo.preload(:project)
+      poll_survey(survey, now)
+    end)
```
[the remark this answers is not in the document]
leandroradusky: It worked!

ysbaddaden:
@leandroradusky I found the issue with the null value property that react doesn't like, and cleaned a few things in. I also reproduce the `nil.project_id` error in specs on my local, I'm investigating. I'll push later.

Note: the project's name can't be changed (it's lacking an `onChange` event handler), I suppose it wasn't on purpose?

assets/js/components/projects/ProjectSettings.jsx
ysbaddaden: This actually doesn't work: we need to keep the actual value as a String otherwise we can't type `0.01` for example.

Sigh, React :face_exhaling:

assets/js/components/surveys/SurveyWizardScheduleStep.jsx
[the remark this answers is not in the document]
ysbaddaden: Yeah, I think this was true in React 0.14 and below but not longer with React 15 or something :shrug:

ysbaddaden:
The CI failure is because of `Survey.set_schedule_timezone()` called from `Survey.changeset()`. I'm not sure why it fails, but... the more I think about it, the more I'm convinced it's not the right place to do this.

Looking further I remembered that the schedule is actually an `Ask.Schedule` and I found `Ask.Schedule.default_timezone()` which is used in `AskWeb.SurveyController.create()` :tada:

leandroradusky:
I pulled the changes: It is true that what I've done wasn't working correctly, but now when I create a new survey for the project it still takes the browser timezone, not the project's property timezone. 

This is because in the API:

```
export const createSurvey = (projectId, folderId, generatesPanelSurvey = false) => {
  const timezone = getTimezone() // <- This take it from the browser
```

Maybe here we should fetch the project's timezone somehow?

ysbaddaden:
@leandroradusky Here come the changes to fix the react warnings, and leverage Redux to handle the "global state" that maintains previously fetched projects, moving API calls to the "actions" and having "reducers" to update the "global state" + checks in the component to react to those changes.

Here the benefits are that the router will fetch the project by itself in to the "global state" (unless it's already loaded) so we don't have to fetch it manually (a second time).

See the individual commits to understand better how it works.

assets/js/components/projects/ProjectSettings.jsx
```diff
@@ -0,0 +1,348 @@
...
+          ))}
+        </span>
+      )
+    } else {
+      return null
+    }
+  }
+
+  updateRate(newRate) {
+    this.setState(newRate, () => this.updateInitialSuccessRate())
```
ysbaddaden, line 100: `setState()` is asynchronous, so we must call `updateInitialSuccessRate()` in a callback to make sure that it runs after the state has been updated, otherwise it would run _before_ the state is updated (oops).

ysbaddaden (approved):

Issue #1818, opened 2020-12-18
Change x axis lenght in line chart at survey overview


manumoreira:
Currently, the min lenght for the x axis is 1 month https://github.com/instedd/surveda-d3-components/blob/master/src/Forecasts.js#L92-L93
<img width="1411" alt="survedaLinesChart" src="https://user-images.githubusercontent.com/31930782/102634847-3434a380-4131-11eb-974f-ca9917367c51.png">

As there are surveys that are completed in less than 1 week we should change 1 month to 1 week

PR #2201, opened 2023-03-08
Fix of linechart on survey card

### LINKED ISSUE #2199: Line chart not working on survey card
Currently, the line chart in the survey card is not working.

Screenshot 2023-03-06 at 10 53 05

This might be probably related to #2054

leandroradusky:
Close #2199 

Now the Survey Card will show the competition percentage and its evolution:

image

The problem was that the `RECEIVE_RESPONDENTS_STATS` wasn't transmitted into the component properties. To accomplish this I've created a `mapStateToProps` function as there is in many other places on the project. 

Note for @manumoreira: the line plot will have in the x-axis a minimum 3 months (or more if the survey takes longer) that made me suspect that wasn't fulfilled in surveys of 2-3 days... Just for you to consider if you want to do something different there.

### DIFF
--- a/assets/js/components/respondents/RespondentsChart.jsx
+++ b/assets/js/components/respondents/RespondentsChart.jsx
@@ -53,18 +53,18 @@ class RespondentsChart extends PureComponent {
     if (!cumulativePercentages || totalQuestionnaires < 1) {
       initialDate = new Date()
       lastDate = new Date()
-      lastDate.setDate(lastDate.getDate() + 90)
+      lastDate.setDate(lastDate.getDate() + 7)
     } else {
       // Uses random one because all questionnaires have the same range of dates.
       const randomQuestionnaireId = Object.keys(cumulativePercentages)[0]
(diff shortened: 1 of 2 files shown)


manumoreira:
A while ago we decided to use 1 week as the min value for the x-axis (see #1818), would it be possible to apply it to the card as well?

leandroradusky:
Done!

More than 1 week will look like that:

image

Less than one week will show one week:

image

ysbaddaden (approved):

manumoreira:
What does the label MAR refer to? If it is a month, can we get something more descriptive than a month for every week? 
Week numbers might look a bit too technical, maybe just the first day of the week (6-Mar, 13-Mar, 20-Mar).... I'm also open to new ideas

ysbaddaden:
Damn, I merged too fast :laughing: 

Let's have a followup @leandroradusky

leandroradusky:
Solved in [#2202](https://github.com/instedd/surveda/pull/2202)

PR #2203, opened 2023-03-10
Set default timezone to browsers timezone on project creation

### LINKED ISSUE #2194: Set project's default timezone to browser's timezone (on create)
When creating a project, we should set its default timezone to be that of the browser, and let the backend fallback to `Etc/UTC` if we can't get the browser's timezone.

On update the project's should only ever be changed manually from the new "Project Settings" page (see #2172).

leandroradusky:
Close #2194 

The `getTimeZone` function that was used to get the browsers' timezone was rescued to be used on project creation.

It was tested that:
- On creation it takes the browser timezone,
- that then this is used when created surveys, but if the survey's timezone is modified the modification remains,
- that modifying the project's timezone the original timezone taken from the browser was replaced and works properly.

If there is no browser timezone then `null` is passed and `UTC` will be used, but I wasn't able to test this. Any hint on how to do that is appreciated :)

### DIFF
--- a/assets/js/components/projects/ProjectIndex.jsx
+++ b/assets/js/components/projects/ProjectIndex.jsx
@@ -29,6 +29,14 @@ class ProjectIndex extends Component {
     this.fetchProjects()
   }
 
+  getTimezone() {
+    try {
+      return Intl.DateTimeFormat().resolvedOptions().timeZone
+    } catch (ex) {
+      return null
+    }
+  }
+
   newProject(e) {
     e.preventDefault()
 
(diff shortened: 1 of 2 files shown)


ysbaddaden (approved):
@leandroradusky you may want to write a test for the controller to assert that when we pass a timezone, the project is created with that timezone, but when we don't it's UTC?

leandroradusky:
Just doublechecking that what I did is good @ysbaddaden: Added just one test to check that timezone was settled... Otherwise will be just `nil` and then the `Schedule` behave like this, so no need to make it not `nil` in project if not settled. 

```
  def default_timezone(project \\ nil) do
    if project do
      project.timezone || "Etc/UTC"
    else
      "Etc/UTC"
    end
  end
```

ysbaddaden (approved):

PR #2219, opened 2023-03-16
Fix for callbacks bypassed by nuntium channel

### LINKED ISSUE #2150: Channel Broker bypassed in Nuntium Channel
Before the existence of the Channel Broker, the endpoint that receives AT messages from Nuntium ([this function](https://github.com/instedd/surveda/blob/289e36dab88b8c55153bb929375ffb8f5aca141f/lib/ask/runtime/nuntium_channel.ex#L95-L139)) answers the callback with a JSON that contains AO messages. Then those AO messages are immediately created on Nuntium.

While developing the Channel Broker, we weren't aware about this behavior. We thought `NuntiumChannel.ask` was the only function that delivered AO messages.

The main negative consequence of this bug is:

1. These AO messages aren't limited by the channel capacity (and its overloading state!). These messages should be sent to the Channel Broker with the maximum level of prioritization, and it should send them ASAP. But if the channel is overloaded they should wait until it's not.

Another negative consequence I'm observing is:

2. Every AO message sent to Nuntium will trigger an AO message status callback from Nuntium to Surveda, handled by [this function](https://github.com/instedd/surveda/blob/289e36dab88b8c55153bb929375ffb8f5aca141f/lib/ask/runtime/nuntium_channel.ex#L65-L93). The ChannelBroker will be noticed about this update and will try to clear a slot for a specific respondent. The ChannelBroker won't know for now (maybe this can be fixed in the future) to which AO message this status update is specifically related. Then, the slots it will try to clear will be more than it previously created when sending AO messages. I don't expect this to be a big problem, because this Channel Broker behavior is wrapped by respondent. This is why I expect the ChannelBroker will support having more status updates than expected without much trouble. But it wasn't for sure what was planned while developing it.

leandroradusky:
close #2150.

In the `callback` function in `NuntiumChannel` if there is an answer upon calling `reply_to_messages` they will be sent bypassing the `ChannelBroker`. 
To avoid this, a new method `force_active_respondent` in the `ChannelBroker` put the respondent within the `activeContacts`. 

Tests were performed to see that this is working, and also that they are added and later deactivated. A small survey of 5 respondents was run. 

In the following plots, you will see in the x-axis the time when a `Logger` event was logged, in the y-axis you will see the number of active contacts, and in the tooltip, you will see the function executed to obtain this result (plot not available in this branch, but in the console output you should be able to see the same):

1) At the beginning, the contacts are normally activated through the `activate_contacts` method, reaching a contact from the queue:

image

2) When a callback is received from the method indicated in the issue, now the respondent is forced to be active (here called `activate_contact_by_force`):

image

Other callback, another forced activation:

image

3) Contacts are then normally deactivated (as if they were activated with the `activate_contacts` function of **1)**)

image

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -102,10 +102,10 @@ defmodule Ask.Runtime.ChannelBroker do
     )
   end
 
-  def force_activate_respondent(channel_id, respondent_id, provider) do
+  def force_activate_respondent(channel_id, respondent_id) do
     call_gen_server(
       channel_id,
-      {:force_activate_respondent, respondent_id, provider}
+      {:force_activate_respondent, respondent_id}
     )
   end
 
@@ -834,7 +834,7 @@ defmodule Ask.Runtime.ChannelBroker do
 
   @impl true
   def handle_call(
-        {:force_activate_respondent, respondent_id, provider},
+        {:force_activate_respondent, respondent_id},
         _from,
         %{
           active_contacts: active_contacts,
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -102,6 +102,13 @@ defmodule Ask.Runtime.ChannelBroker do
     )
   end
 
+  def force_active_respondent(channel_id, respondent, provider) do
+    call_gen_server(
+      channel_id,
+      {:force_active_respondent, respondent, provider}
+    )
+  end
+
   defp call_gen_server(channel_id, message) do
     {channel_id, _} =
       if is_integer(channel_id), do: {channel_id, nil}, else: Integer.parse(channel_id)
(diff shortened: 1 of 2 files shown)


lib/ask/runtime/nuntium_channel.ex
```diff
@@ -192,6 +192,19 @@ defmodule Ask.Runtime.NuntiumChannel do
...
     json_reply = reply_to_messages(reply, from, respondent_id, channel_id)
+
+    case json_reply do
+      [] ->
+        :ok
+
+      _ ->
+        ChannelBroker.force_active_respondent(
+          channel_id,
+          Repo.get(Respondent, respondent_id),
```
ysbaddaden, line 203: Reading `ChanneBroker.activate_respondent/2` I see it only needs `respondent.id`; we could spare a query (we already request the respondent twice above) by passing `respondent_id` directly.

lib/ask/runtime/channel_broker.ex
```diff
@@ -825,6 +866,25 @@ defmodule Ask.Runtime.ChannelBroker do
...
+        _from,
+        %{config: config, channel_id: channel_id} = state
+      ) do
+    Logger.debug("CHN_BRK force_active_respondent: #{inspect(binding())}")
+
+    end_state =
+      case provider do
+        "nuntium" ->
+          activate_respondent(state, respondent)
+      end
```
ysbaddaden, line 881: I wouldn't restrict the function to Nuntium. We only call it from Nuntium but that's the channel's choice/requirement to use the feature, or not.

ysbaddaden (commented):
The channel broker's state is still nebulous to me. I trust that you did the correct thing.

I see you're using two different namings `force_active_respondent` and `activate_respondent`, maybe there should be only one? Either one, or a mix of both, for example `force_activate_respondent` to outline that this isn't the normal behavior, yet keep the `activate` naming that is already used for this action (i.e. `activate_contact`).

leandroradusky:
I've deleted `activate_respondent` and left only `force_activate_respondent` since it was only called from there and was the only instruction given that we've eliminated the provider case.

ysbaddaden (approved):

PR #2221, opened 2023-03-17
Added batch_limit_per_minute to project settings

### LINKED ISSUE #2206: Add batch_limit_per_minute on admin console
`batch_limit_per_minute` currently is an environment variable that operates at the instance level to control the amount of max number of respondents that are set to active in the surveys.
We should add it to the project admin console as we did with `eligibility_rate`, `response_rate`, etc
adminPanelLimitPerMinute

leandroradusky:
Close #2206 

Project settings now include a new field, `batch_limit_per_minute`. It's added to the settings tab following the issue's mockup:

image

In order to make it functional, the `batch_limit_per_minute\0` was modified to `batch_limit_per_minute\1` taking the parameter and returning the global setting if it's `nil` on the project.

The two functions using `batch_limit_per_minute` are `retry_respondents` & `start_some`. In the latter, the `survey` is a parameter, then we can easily access the project. For `retry_respondents`, all projects where exists respondents to retry are iterated and they are retried respecting the project's `batch_limit_per_minute`.

### DIFF
--- a/lib/ask/runtime/broker.ex
+++ b/lib/ask/runtime/broker.ex
@@ -7,6 +7,7 @@ defmodule Ask.Runtime.Broker do
     Repo,
     Logger,
     Survey,
+    Project,
     Respondent,
     RespondentGroup,
     QuotaBucket,
@@ -36,11 +37,7 @@ defmodule Ask.Runtime.Broker do
   def init(_args) do
     :timer.send_after(1000, :poll)
 
-    Logger.info(
-      "Broker started. Default batch size=#{default_batch_size()}. Limit per minute=#{
-        batch_limit_per_minute()
-      }."
-    )
+    Logger.info("Broker started. Default batch size=#{default_batch_size()}.")
 
     {:ok, nil}
   end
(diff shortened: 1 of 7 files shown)


priv/repo/structure.sql
ysbaddaden, line 1: There are too many changes here. Your database still has the channel broker tables. To file shall be cleaned up before merge.

ysbaddaden (commented):
Hey, nothing much to say :muscle: 

Just some notes to simplify a few things or improve readability, otherwise :+1:

lib/ask/runtime/broker.ex
```diff
@@ -226,14 +223,34 @@ defmodule Ask.Runtime.Broker do
...
+          exists(
+            from survey in Survey,
+              join: r in Respondent,
+              on: r.survey_id == survey.id,
+              where:
+                parent_as(:project).id == survey.project_id and r.state == :active and
+                  r.timeout_at <= ^now,
+              select: 1,
+              limit: 1
+          )
```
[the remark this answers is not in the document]
leandroradusky: I did it, but just to clarify: 

I still need to load the project's struct in order to pass it to the `batch_limit_per_minute\1` function. 

I could pass there the `project_id` but then I will need to load the project struct to get its `batch_limit_per_minute` and in the other call of this function I have the project already, I will pass the id and reload the project when there is no need. 

I did that (complicated, I admit it :P) query to have the project struct. I think now is clearer but still not sure is the best option.

priv/repo/structure.sql
leandroradusky: True!! 

I cleaned up manually... There's a more rational way to do it? Run a revert/migrate with an N old enough?

ysbaddaden: yes, you can rollback the migrations in the branch they're defined to cleanup your database. Then you can dump or re-run the migration in the new branch to have a cleaned `structure.sql`.

lib/ask/runtime/broker.ex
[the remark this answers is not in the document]
ysbaddaden: Urgh, of course you need the projects *and* it's better to query them _once_ than one by one :facepalm: 

I got confused with the ecto_sql query that looks more complex than it really is, and using `EXISTS()` was a much better solution. I'm sorry I said :hankey:

ysbaddaden (approved):

PR #2224, opened 2023-03-20
Showing running surveys count in folder card

### LINKED ISSUE #2196: Show running surveys count in surveys folders
Surveda's Projects screen show every project with the amount of currently running surveys.

Screenshot 2023-03-03 at 12 12 48

Before introducing survey folders, when the user selected a project, the Surveys index would quickly show which surveys were running - since they stay at the top.

But now there's a chance that the running surveys are inside a folder, so the user gets into a project thinking "there are 2 running surveys", but then get into the project - and have no clue where those two running surveys are.

Screenshot 2023-03-03 at 12 12 57

It'd be nice to display a counter in each folder, stating how many active surveys there are.

Something in this spirit - but that doesn't hurt anybody's eyes.

Screenshot 2023-03-03 at 12 12 57

leandroradusky:
Close #2196 & #2222.

Running surveys count in folders (if any). 

Awaiting for designs @manumoreira, its functionally implemented and looks like this:

image

### DIFF
--- a/lib/ask_web/controllers/folder_controller.ex
+++ b/lib/ask_web/controllers/folder_controller.ex
@@ -49,23 +49,15 @@ defmodule AskWeb.FolderController do
   def count_running_surveys(folder_id) do
     surveys =
       from(s in Survey,
-        where: s.folder_id == ^folder_id,
-        where: s.state == ^:running,
-        select: count(s.id)
-      )
-      |> Repo.one()
-
-    pannel_surveys =
-      from(s in Survey,
-        join: ps in PanelSurvey,
+        left_join: ps in PanelSurvey,
         on: s.panel_survey_id == ps.id,
-        where: ps.folder_id == ^folder_id,
+        where: (ps.folder_id == ^folder_id or s.folder_id == ^folder_id),
         where: s.state == ^:running,
         select: count(s.id)
       )
       |> Repo.one()
 
-    surveys + pannel_surveys
+    surveys
   end
 
   def show(conn, %{"project_id" => project_id, "id" => folder_id}) do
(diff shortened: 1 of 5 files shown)


lib/ask_web/controllers/folder_controller.ex
```diff
@@ -41,6 +41,7 @@ defmodule AskWeb.FolderController do
         where: f.project_id == ^project.id
       )
       |> Repo.all()
+      |> Repo.preload(:surveys)
```
ysbaddaden, line 44: We actually did the exact opposite 1.5 years ago to fix a loading issue (loading too much is fine when you have a dozen surveys, but it's starting to take far too long when you start to have hundreds of them).

We might want to report how many running surveys a folder has, with a quick SQL query.

Also: a Folder can contain PanelSurveys that themselves can contain running Surveys, so you'll probably want to left join them too in the SQL query.

ysbaddaden (changes requested):

lib/ask_web/controllers/folder_controller.ex
leandroradusky: `PanelSurvey` ids are stored in the `Survey` then just querying the surveys table should make it no? There is no `state` in the panel surveys to match `running` with, this is on the surveys.

leandroradusky:
Design is fixed according to the [mockup](https://projects.invisionapp.com/d/main#/console/8374235/180693793/preview):

image

And the changes requested are applied.

ysbaddaden (commented):
Looks good to me!

Did you try the edge case where we have a survey running in a panel survey in a folder? Is it counted? I believe a survey running in a panel survey won't have a folder id (it's not directly under the folder, it's only through the panel survey).

PR #2232, opened 2023-04-03
Add batch-size-to-project-settings

### LINKED ISSUE #2229: Add default_batch_size to project console
The env variable batch_size is used when no target is defined to limit the size of the batch of respondents.
Adding it to the admin console can be useful

leandroradusky:
Closes #2229.

A new setting was added to project settings, `batch_size`. In the form, it was added in a "new line" and not in the same line as `batch_limit_per_minute`, not sure if this should be changed (cc @manumoreira).

image

Functionally, `batch_size` is used in the `Broker` and, if `completed_respondents_needed_by(survey)` returns `:all` then the `default_batch_size` given by the environment was used, and this was modified to use the `batch_size` of the project if it is defined. Otherwise, the `batch_size` is computed based using other rates, and then this was not modified.

### DIFF
--- a/lib/ask/runtime/broker.ex
+++ b/lib/ask/runtime/broker.ex
@@ -326,7 +326,7 @@ defmodule Ask.Runtime.Broker do
   defp batch_size(survey, respondents_by_state) do
     case Survey.completed_respondents_needed_by(survey) do
       :all ->
-        default_batch_size()
+        survey.project.batch_limit_per_minute || default_batch_size()
 
       respondents_target when is_integer(respondents_target) ->
         successful = Survey.completed_state_respondents(survey, respondents_by_state)
--- a/priv/repo/structure.sql
+++ b/priv/repo/structure.sql
@@ -289,6 +289,7 @@ CREATE TABLE `projects` (
   `response_rate` double DEFAULT NULL,
   `valid_respondent_rate` double DEFAULT NULL,
   `batch_limit_per_minute` int(11) DEFAULT NULL,
+  `batch_size` int(11) DEFAULT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `id` (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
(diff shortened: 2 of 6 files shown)


lib/ask/runtime/broker.ex
```diff
@@ -326,7 +326,7 @@ defmodule Ask.Runtime.Broker do
   defp batch_size(survey, respondents_by_state) do
     case Survey.completed_respondents_needed_by(survey) do
       :all ->
-        default_batch_size()
+        survey.project.batch_limit_per_minute || default_batch_size()
```
ysbaddaden, line 329: I think you meant:
```suggestion
        survey.project.batch_size || default_batch_size()
```

priv/repo/structure.sql
ysbaddaden, line 1: The schema was updated, but I don't see the actual migration file?

manumoreira:
I agree with the layout

lib/ask/runtime/broker.ex
leandroradusky: :face_with_spiral_eyes:

priv/repo/migrations/20230317094712_add_batch_limit_per_minute_to_projects.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Ask.Repo.Migrations.AddSettingsToProjects do
+defmodule Ask.Repo.Migrations.AddBatchLimitPerMinutToProjects do
```
leandroradusky, line 1: This was a mistake on a previous PR, solving it now.

priv/repo/migrations/20230317094712_add_batch_limit_per_minute_to_projects.exs
```diff
@@ -1,4 +1,4 @@
-defmodule Ask.Repo.Migrations.AddSettingsToProjects do
+defmodule Ask.Repo.Migrations.AddBatchLimitPerMinuteToProjects do
```
leandroradusky, line 1: This was a mistake on a previous PR, solving it now.

ysbaddaden (approved):

PR #2234, opened 2023-04-04
Fix/2233 unique project user pair

### LINKED ISSUE #2233: Project memberships must be unique for each project/user pair
Copy of #1721

leandroradusky:
Close #2233.

After studying `Invite` and `ProjectMembership` and trying many things, this is the one that worked better, no 100% sure this is the way to go though.
Because I'm using `get_change` and not `get_field` the validation function added will allow for updating the role of the users in the project with no problem. 
When inviting a user already in the project, the modal will close and no error will be displayed; the user will be listed in the collaborators' index already, then I took it as correct behavior. A more complex one would be to validate inputted email to be in the project on typing and inform that in the modal but I don't think it deserved to be done.

### DIFF
--- a/priv/repo/migrations/20230405111657_remove_project_memberships_duplicates.exs
+++ b/priv/repo/migrations/20230405111657_remove_project_memberships_duplicates.exs
@@ -1,0 +1,32 @@
+defmodule Ask.Repo.Migrations.RemoveProjectMembershipsDuplicates do
+  use Ecto.Migration
+
+  def up do
+    # Remove duplicates and keep max level in column level for each user and project in project_membnerships
+    execute("""
+      DELETE t1
+      FROM project_memberships t1
+      INNER JOIN (
+          SELECT user_id, project_id, MAX(
+              CASE level 
+                  WHEN 'owner' THEN 4 
+                  WHEN 'admin' THEN 3 
+                  WHEN 'editor' THEN 2 
+                  WHEN 'reader' THEN 1 
+                  ELSE 0 
+              END) AS max_level
+          FROM project_memberships
+          GROUP BY user_id, project_id
+      ) t2 
+      ON t1.user_id = t2.user_id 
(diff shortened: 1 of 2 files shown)


priv/repo/migrations/20230405111657_remove_project_memberships_duplicates.exs
```diff
@@ -0,0 +1,32 @@
...
+      AND t1.project_id = t2.project_id 
+      AND CASE t1.level 
+              WHEN 'owner' THEN 4 
+              WHEN 'admin' THEN 3 
+              WHEN 'editor' THEN 2 
+              WHEN 'reader' THEN 1 
+              ELSE 0 
+          END < t2.max_level;
+    """)
+  end
```
leandroradusky, line 31: Had to do this, which is kind of verbose because: 

- When trying to go with a solution using 
`SELECT id, ROW_NUMBER() OVER (PARTITION BY user_id, project_id ORDER BY level DESC) AS row_number FROM project_memberships;` 
I get a syntax error, I've checked and re-checked and I cannot find the error there.
- My local server has enabled `sql_mode=only_full_group_by` which does not allow me to make a simpler version, and probably the DB in stg-prod are configured the same way.

priv/repo/migrations/20230405111657_remove_project_memberships_duplicates.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Ask.Repo.Migrations.RemoveProjectMembershipsDuplicates do
+  use Ecto.Migration
+
+  def up do
+    # Remove duplicates and keep max level in column level for each user and project in project_membnerships
+    execute("""
+      DELETE t1
```
ysbaddaden, line 7: Are you sure about `DELETE t1`? Shouldn't it be `DELETE FROM project_memberships t1`?

PR #2237, opened 2023-04-06
Migrate Poison to Jason

### LINKED ISSUE #2046: Migrate from Poison to Jason (Phoenix 1.4)
Phoenix 1.4+ can use either [Poison](https://hex.pm/packages/poison) or [Jason](https://hex.pm/packages/jason) for its own parse/dump JSON (be it internal or implicit). Phoenix now recommends Jason, and it's supposed to be faster, but there seemed be some issues, so I'm sticking to Poison for now.

We also use Poison explicitly to [parse/dump JSON in Surveda](https://github.com/instedd/surveda/search?q=Poison). We could probably migrate those cases to use Jason.

Refs #2019

leandroradusky:
Close #2046.

Changed all references to `Poison` into references to `Jason`.

- For some structs, `@derive Jason.Encoder` has to be added since this needs to be explicit for Jason. 
- The hardcoded return values of some tests needs to be modified since now the keys of the JSON returned are sorted alphabetically.

**Just one replacement is missing:**

```
  #lib/ask/o_auth_token.ex:37
  def access_token(token) do
    Poison.Decode.decode(token.access_token, as: %OAuth2.AccessToken{})
  end
``` 

Don't know why but changing here `Poison.Decode.decode` to `Jason.decode` would break lots of tests, don't know if this has to do something with the external wiring of the project, or if this schema should have something similar to the `@derive` statement mentioned above.
Because of that, `Poison` was not removed from the mix file. `Jason` version (1.0) was kept in the mix file, while there are newer versions available.

### DIFF
--- a/lib/ask/ecto_types/day_of_week.ex
+++ b/lib/ask/ecto_types/day_of_week.ex
@@ -1,6 +1,7 @@
 defmodule Ask.DayOfWeek do
   alias __MODULE__
 
+  @derive Jason.Encoder
   defstruct [:sun, :mon, :tue, :wed, :thu, :fri, :sat]
 
   def cast(%DayOfWeek{} = day_of_week) do
--- a/lib/ask/ecto_types/json.ex
+++ b/lib/ask/ecto_types/json.ex
@@ -2,6 +2,6 @@ defmodule Ask.Ecto.Type.JSON do
   use Ecto.Type
   def type, do: :longtext
   def cast(any), do: {:ok, any}
-  def load(string) when is_binary(string), do: Poison.decode(string)
-  def dump(json), do: Poison.encode(json)
+  def load(string) when is_binary(string), do: Jason.decode(string)
+  def dump(json), do: Jason.encode(json)
 end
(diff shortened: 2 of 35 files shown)


ysbaddaden:
Ah, that one was meant for @nthiad but she'll may be able to look at the issue you reached.

leandroradusky:
Ohh Sorry!! I've been doing some peer programming with Caspian today and I chose from the board what seems doable to show him some code! Sorry @nthiad!!

leandroradusky:
I applied the change proposed by @nthiad and it worked neatly :) Thanks!

Despite that, I couldn't remove `Poison` from the mix file because of [this](https://hexdocs.pm/oauth2/changelog.html#improvements-6). I've left a comment there to remove it in the future.

@ysbaddaden, I @ you to re-ask for your review :)

ysbaddaden (commented):
@leandroradusky That looks good to me, but I notice that `oauth2` is already v0.7.0 in `mix.exs` :thinking:

leandroradusky:
Exactly, in v0.7.0 Poison is only required as a test dependency, if I remove `Poison` tests fail... There's something I'm misinterpreting? Effectively, if I remove `Poison` from `mix.exs` lots of tests fail.

ysbaddaden:
Yeah, same here, I'll try.

PR #2239, opened 2023-04-06
Fix: channel broker verboice callbacks

### LINKED ISSUE #2238: Verboice status callback crashes on "completed" completion (& channel broker is never notified)
There are 2 callbacks from Verboice:

1. _Call progress_: notifies about the current disposition of the call, reports the digits typed (or a timeout?) and Surveda responds with the next action to continue the call (or hangup).

2. _Call status_: notifies of status change to the call (ringing, in-progress, completed, busy, failed, expired) and we update the respondent accordingly.

The status callback (2) now tries to notify the Channel Broker whereas the call status (1) doesn't. Looking at Surveda logs, we sometimes reply to a status callback (2) with a 500 internal server error.

After some ~~bad old printf~~ debugging, I noticed that it's always failing for the `completed` status.

In order to notify the Channel Broker, we need to know which Channel is currently in use for the respondent, we search for respondent -> current_mode -> session -> channel, but when we received the `completed` call status... the channel is `nil` yet we try to access `channel.id` => :boom:

ysbaddaden:
*The bug*

Verboice has two callbacks: one to make progress in the call and another for statuses. In the case where the progress callback ends the call, the respondent won't have a session anymore when we receive the actual status:completed callback.

This resulted in the channel broker getting stuck respondents when they completed the survey (oops). This patch fixes the issue by immediately notifying the channel broker with the completed state from the progress callback.

Also a tiny performance improvement (only load the current mode, not the whole session, `Ask.Runtime.Session.load_current_mode/1`) and avoid relying on exceptions when trying to reach the respondent's channel.

*About tests*:

I don't like adding dependencies, but `Mox` seemed too complex: it would have meant creating a swap-able implementation of `Ask.Runtime.ChannelBroker`, and I also couldn't figure out how to set the mock only for some tests (or does it only mock when you `expect` something? or it should allow fallback from mock to real implementation?).

`Mock` was much easier (once you understand how to match) and didn't need to change anything.

closes #2238


ysbaddaden:
Grmbl, the test suites uses a `SurveyStub` module and a `TestChannel` module that doesn't have a `channel_id` attribute, so we must handle these test-specific cases in the application's runtime :face_exhaling:

This seems to be the whole channel id `nil` to `0` mapping in the channel broker.

ysbaddaden:
Allright, now I broke some channel broker tests. That was expected.

lib/ask/runtime/verboice_channel.ex
```diff
@@ -408,6 +395,28 @@ defmodule Ask.Runtime.VerboiceChannel do
...
+      _ ->
+        :error
+    end
+  end
+
+  defp respondent_channel(%{"current_mode" => %{"mode" => "ivr", "channel_id" => _}} = state) do
+    Ask.Runtime.Session.load_current_mode(state).channel
+  end
+
+  defp respondent_channel(%{"current_mode" => %{"mode" => "ivr"}}), do: 0
```
ysbaddaden, line 417: We should only allow this during test where we often don't have a `channel_id` (and use `0` as a fallback). Something like (to be verified):

```suggestion
  defp respondent_channel(%{"current_mode" => %{"mode" => "ivr"}}) when Mix.env == "test", do: 0
```

ysbaddaden: We could do the same in `Ask.Runtime.ChannelBroker` and simplify all the duplicate methods that map `nil` to `0`. I'll push another PR on Tuesday.

mix.exs
```diff
@@ -100,7 +101,7 @@ defmodule Ask.Mixfile do
       # {:telemetry, "~> 0.4.3"},
 
       # held back because of warnings & errors with newer versions
-      {:swoosh, "~> 0.17.0"},
+      {:swoosh, "~> 0.17.0"}
```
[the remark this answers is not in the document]
ysbaddaden: Yeah, it did. I think some rule changed between Elixir 1.8 and Elixir 1.10 :shrug:

lib/ask/runtime/verboice_channel.ex
ysbaddaden: @leandroradusky I pushed the commit.

PR #2244, opened 2023-04-12
Moves build tasks from build.sh to Dockerfile (#2223)

nthiad:
Here is a first attempt at #2223.

I tried to make this minimal but I have some questions:

* Should any of these tasks run as a non-root user?
* Is yarnpkg from apt ok or should it use a more recent non-apt source?
* Should `mix deps.get` use `--only prod`? I took it out to match `build.sh` but perhaps that is wrong here.

I didn't remove build.sh entirely in this patch since it has other tasks like `dockerBuildAndPush` but perhaps this could be removed if somebody who knows more about this can comment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,12 @@ FROM elixir:1.10 AS dev
 RUN sed -i '/^mozilla\/DST_Root_CA_X3/s/^/!/' /etc/ca-certificates.conf && update-ca-certificates -f
 
 RUN apt -q update && \
-    apt -q install -y default-mysql-client inotify-tools festival yarnpkg && \
+    apt -q install -y default-mysql-client inotify-tools festival && \
     apt -q install -y --no-install-recommends ffmpeg libaacs0 && \
     apt -q clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 RUN mix local.hex --force
 RUN mix local.rebar --force
-
-FROM dev AS release
 
 ENV MIX_ENV=prod
 
@@ -26,10 +24,23 @@ ADD . /app
 RUN mix compile
 RUN mix phx.digest
 
-RUN yarnpkg install --no-progress
-RUN yarnpkg deploy
+FROM node:10 as js
 
+COPY --from=dev /deps /deps
+ADD . /app
+WORKDIR /app
+
+RUN yarn install --no-progress
+RUN yarn deploy
+
+FROM dev AS release
+
+COPY --from=js /app/assets /app/
+
+ENV MIX_ENV=prod
 ENV PORT=80
+
+WORKDIR /app
 EXPOSE 80
 
 CMD elixir --sname server -S mix phx.server


ysbaddaden (commented):
_To answer your questions:_

The task can run as root. Docker makes it a nightmare to try to use another user. Still, a good practice is to try to run the eventual command as non-root... but we don't do it (yet, probably something to consider).

Using system packages is fine :+1: 

Yes, we'd like to keep `mix deps.get --only prod`. Looking at `mix.exs` and `package.json` it seem that the elixir dependencies with an impact on the frontend in production are installed.

_To further improve_:

Ideally, we'd like to avoid having `yarnpkg` and more importantly the `node_modules` in the resulting image, we only need the built assets, and `node_modules` grows very big.

Sadly we can't just delete them, because a Docker image is made of all the intermediary layers; it's not a copy of the end image state. Each command in the `Dockerfile` is a layer, and each command is responsible to cleanup, hence the `RUN apt update && apt install && apt clean && rm` command for example, solely to avoid artifacts.

This is where multistage images are useful. In this Dockerfile, we take advantage of it to have a single `Dockerfile` to build a `dev` image (as requested by `target: dev` in docker compose), with just the bare system, enough for local development, that we segue into a `release` image to build the actual image to deploy.

The advantage of multistage is that we can build something, then start again from another image, and copy the bits we want from it in the final image. There is an example in [Citta's Dockerfile](https://github.com/manastech/citta/blob/master/Dockerfile): we compile using a crystal image, then start the actual release image using a bare alpine system and copy the built binaries.

I think you can intertwine multiple stages?

Sorry for the long description, but you said you're not very familiar with Docker, so I went in verbose mode :bow: 

NOTE: now that I think about it, I'm wondering if we _need_ the elixir `deps` or even the Elixir image at all after we compile? and... [apparently we don't](https://staknine.com/build-an-elixir-release-with-docker-to-deploy-anywhere/) but let's save that for later.

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM elixir:1.10 AS dev
 RUN sed -i '/^mozilla\/DST_Root_CA_X3/s/^/!/' /etc/ca-certificates.conf && update-ca-certificates -f
 
 RUN apt -q update && \
-    apt -q install -y default-mysql-client inotify-tools festival && \
+    apt -q install -y default-mysql-client inotify-tools festival yarnpkg && \
```
matiasgarciaisaia, line 7: We're not using `yarn` in the development image (since we're running a `webpack` service in the `docker-compose.yml`). And we definitely not need it in the production one.

Dockerfile
```diff
@@ -19,13 +19,16 @@ ADD mix.exs mix.lock /app/
...
-RUN mix deps.get --only prod
+RUN mix deps.get
 RUN mix deps.compile
 
 ADD . /app
 RUN mix compile
 RUN mix phx.digest
 
+RUN yarnpkg install --no-progress
+RUN yarnpkg deploy
```
matiasgarciaisaia, line 30: We should add yet another stage in the file, using the same image we're running for the `webpack` service right now, and build the UI as we're now doing in the `build.sh` file - but in this Dockerfile, so we can then `COPY --from=webpack` those resulting files.

ysbaddaden (commented):
@nthiad Nice, we're almost there!

The only issue is setting `MIX_ENV=prod` for the `dev` build. Maybe we could squeeze a `deps` stage? For example:

    dev -> deps -> js
                -> release

So that `dev` would only have the bare erlang image (where we can install, update or remove deps at will), but allow js and release to have the erlang deps installed once.

And a nitpick: maybe rename `js` as `frontend`?

PR #2253, opened 2023-04-13
Speed up generation of respondent CSV files

### LINKED ISSUE #2205: Manage large csv files download
Surveda allows users to download 4 different CSV files for each survey

* Results
* Disposition
* Incentives
* Interactions

When running large surveys (> 30000 respondents and > 1000 completes/partials) the download process for these files is usually aborted due to timeout settings.
A while ago @matiasgarciaisaia created this elegant workaround
https://github.com/instedd/surveda/commit/80a6cb3884e5f04fbe1fa41fcd7e4d46d907fb64
Currently, every time a user has problems downloading CSVs we disable the timeouts and enable them back a few days after.
We should find a long-term solution for exporting CSV files from Surveda.
The solution might require some research on how the files are generated and then analyzing the potential solutions. 
Let's use this issue for initial research.

ysbaddaden:
The main bottleneck was the manipulation of datetimes to shift to the survey scheduled timezone, as well as formatting those datetimes, or repeatedly accessing `survey.schedule.timezone` when we could cache the values for the whole HTTP request.

That may not be the only bottleneck or optimization possible, but that's a first step that actually allows the CSV files to be generated faster. I don't have enough data to have a significant benchmark for the results and incentives CSV, and the disposition history CSV was already optimized, but the interactions CSV is now magnitudes faster:

- `results.csv` (60% faster):
  - main: 5s at 168 KB/s
  - patch: 2s at 431 KB/s

- `disposition_history.csv` (no improvement):
  - main: 2s at 1084 KB/s
  - patch: 2s at 1023 KB/s

- `incentives.csv` (66% faster):
  - main: 3s at 155 KB/s
  - patch: 1s 486 KB/s

- `interactions.csv` (93% faster):
  - main: 31s at 261 KB/s
  - patch: 2s at 1023 KB/s

@matiasgarciaisaia As far as I understand, SQL queries are negligible, the bottleneck is the Elixir/Erlang code that manipulates the data to generate the CSV files. The impact is _very_ noticeable.

I tried to replace the ugly serialization of DateTime to String, to use more readable functions now available in Elixir 1.10 such as `DateTime.add/4` or the legacy `Timex.format!/3` but they had too much of an impact (at least 2 or 3 times as slower). Newer versions of Elixir have a datetime formatter, maybe it will be faster, and it has been optimized since...

closes #2205

### DIFF
--- a/lib/ask/time_util.ex
+++ b/lib/ask/time_util.ex
@@ -1,16 +1,41 @@
 defmodule Ask.TimeUtil do
+  @months %{
+    1 => "Jan",
+    2 => "Feb",
+    3 => "Mar",
+    4 => "Apr",
+    5 => "May",
+    6 => "Jun",
+    7 => "Jul",
+    8 => "Aug",
+    9 => "Sep",
+    10 => "Oct",
+    11 => "Nov",
+    12 => "Dec",
+  }
+
   def format(ecto_timestamp, offset_seconds, offset_label) do
+    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
+    "#{year}-#{pad2(month)}-#{pad2(day)} #{pad2(hour)}:#{pad2(min)}:#{pad2(sec)} #{offset_label}"
+  end
+
+  # Applies offset and formats datetime as `Timex.format!("%b %e, %Y %H:%M #{tz_offset}", :strftime)`.
+  # There are no reasons to not use Timex.format unless you need raw performance and make lots of
+  # repeating calls in a tight loop (eg. exporting a CSV) which is sole reason that method exists.
+  def format2(ecto_timestamp, offset_seconds, offset_label) do
+    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
+    "#{Map.get(@months, month)} #{pad2(day, " ")}, #{year} #{pad2(hour)}:#{pad2(min)} #{offset_label}"
+  end
+
+  defp to_ymdhms(ecto_timestamp, offset_seconds) do
     timestamp_in_seconds =
       ecto_timestamp
       |> Timex.to_erl()
       |> :calendar.datetime_to_gregorian_seconds()
 
-    {{year, month, day}, {hour, min, sec}} =
-      (timestamp_in_seconds + offset_seconds)
-      |> :calendar.gregorian_seconds_to_datetime()
-
-    "#{year}-#{pad2(month)}-#{pad2(day)} #{pad2(hour)}:#{pad2(min)}:#{pad2(sec)} #{offset_label}"
+    (timestamp_in_seconds + offset_seconds)
+    |> :calendar.gregorian_seconds_to_datetime()
   end
 
-  defp pad2(s), do: Integer.to_string(s) |> String.pad_leading(2, "0")
+  defp pad2(s, char \\ "0"), do: Integer.to_string(s) |> String.pad_leading(2, char)
 end
--- a/lib/ask/time_util.ex
+++ b/lib/ask/time_util.ex
@@ -23,7 +23,7 @@ defmodule Ask.TimeUtil do
   # There are no reasons to not use Timex.format unless you need raw performance and make lots of
   # repeating calls in a tight loop (eg. exporting a CSV) which is sole reason that method exists.
   def format2(ecto_timestamp, offset_seconds, offset_label) do
-    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
+    {{year, month, day}, {hour, min, _sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
     "#{Map.get(@months, month)} #{pad2(day, " ")}, #{year} #{pad2(hour)}:#{pad2(min)} #{offset_label}"
   end
 
--- a/lib/ask_web/controllers/respondent_controller.ex
+++ b/lib/ask_web/controllers/respondent_controller.ex
@@ -925,19 +925,14 @@ defmodule AskWeb.RespondentController do
         row = row ++ [respondent_group]
 
         questionnaire_id = respondent.questionnaire_id
+        questionnaire = questionnaires |> Enum.find(fn q -> q.id == questionnaire_id end)
         mode = respondent.mode
 
         row =
           if has_comparisons do
             variant =
-              if questionnaire_id && mode do
-                questionnaire = questionnaires |> Enum.find(fn q -> q.id == questionnaire_id end)
-
-                if questionnaire do
-                  experiment_name(questionnaire, mode)
-                else
-                  "-"
-                end
+              if questionnaire && mode do
+                experiment_name(questionnaire, mode)
               else
                 "-"
               end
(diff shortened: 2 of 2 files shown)


lib/ask/time_util.ex
```diff
@@ -1,16 +1,41 @@
...
+    "#{year}-#{pad2(month)}-#{pad2(day)} #{pad2(hour)}:#{pad2(min)}:#{pad2(sec)} #{offset_label}"
+  end
+
+  # Applies offset and formats datetime as `Timex.format!("%b %e, %Y %H:%M #{tz_offset}", :strftime)`.
+  # There are no reasons to not use Timex.format unless you need raw performance and make lots of
+  # repeating calls in a tight loop (eg. exporting a CSV) which is sole reason that method exists.
+  def format2(ecto_timestamp, offset_seconds, offset_label) do
+    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
+    "#{Map.get(@months, month)} #{pad2(day, " ")}, #{year} #{pad2(hour)}:#{pad2(min)} #{offset_label}"
+  end
```
ysbaddaden, line 28: It's awful to replace a couple readable calls into this, but the performance improvement is real when generating results: from over 5s to 2.1s when downloading a file on my local.

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -953,7 +956,7 @@ defmodule AskWeb.RespondentController do
             row
           end
 
-        # We traverse all fields and see if there's a response for this respondent
+        # # We traverse all fields and see if there's a response for this respondent
```
ysbaddaden, line 959:
```suggestion
        # We traverse all fields and see if there's a response for this respondent
```

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -1106,6 +1116,8 @@ defmodule AskWeb.RespondentController do
     project = load_project_for_owner(conn, project_id)
     survey = load_survey(project, survey_id)
 
+    channels = survey_channel_names(survey)
```
ysbaddaden, line 1119: It wasn't necessary per-se to load the list of channels _once_, but it allows to spare a preload request for each batch, leading to halve the number of SQL requests we make, which is always nice (and there are usually just a single digit number of channels, so we don't hog memory).

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -874,6 +874,10 @@ defmodule AskWeb.RespondentController do
       end)
       |> Enum.uniq()
 
+    # timezone = survey.schedule.timezone
+    tz_offset_in_seconds = Survey.timezone_offset_in_seconds(survey)
+    tz_offset = Survey.timezone_offset(survey)
```
ysbaddaden, line 879:
```suggestion
    tz_offset_in_seconds = Survey.timezone_offset_in_seconds(survey)
```

lib/ask/time_util.ex
```diff
@@ -1,16 +1,41 @@
...
   def format(ecto_timestamp, offset_seconds, offset_label) do
+    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
+    "#{year}-#{pad2(month)}-#{pad2(day)} #{pad2(hour)}:#{pad2(min)}:#{pad2(sec)} #{offset_label}"
+  end
+
+  # Applies offset and formats datetime as `Timex.format!("%b %e, %Y %H:%M #{tz_offset}", :strftime)`.
+  # There are no reasons to not use Timex.format unless you need raw performance and make lots of
+  # repeating calls in a tight loop (eg. exporting a CSV) which is sole reason that method exists.
+  def format2(ecto_timestamp, offset_seconds, offset_label) do
+    {{year, month, day}, {hour, min, sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
```
ysbaddaden, line 26:
```suggestion
    {{year, month, day}, {hour, min, _sec}} = to_ymdhms(ecto_timestamp, offset_seconds)
```

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -1130,29 +1139,21 @@ defmodule AskWeb.RespondentController do
...
-          end
-
-        disposition = disposition_label(e.disposition)
-        action_type = action_type_label(e.action_type)
-        timestamp = csv_datetime(e.timestamp, survey)
-
         [
           Integer.to_string(e.id),
           e.respondent_hashed_number,
           interactions_mode_label(e.mode),
```
leandroradusky, line 1151: Just a suggestion here... Since you're saying that SQL goes fast and the problem is the processing... could `interactions_mode_label`, `disposition_label`, `action_type_label` be converted in a `CASE ... WHEN` in the query and report some speeding up? The query will be a little bit uglier, but maybe...

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: We could probably spare preloading the `questionnaire` for each respondent batch in `results` too: we already load the questionnaires for the survey, and have available.

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -1117,8 +1127,7 @@ defmodule AskWeb.RespondentController do
                   ((e.respondent_hashed_number == ^last_hash and e.id > ^last_id) or
                      e.respondent_hashed_number > ^last_hash),
               order_by: [e.respondent_hashed_number, e.id],
-              limit: 1000,
-              preload: :channel
+              limit: 1000
```
leandroradusky, line 1130: Another question, more for my own learning than for a review ... Could this batching size change something if we increase it, i.e., to 2000?

leandroradusky (approved):
LGTM, 

Just left a few questions, more out-loud thinking than actual reviews :)

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: It will load 2000 respondents in memory on each batch, meaning that it would halve the number of SQL requests, but double the number of respondents to return at once, and also double the memory usage. It's all about tradeoff and fine tuning. You can try it out, and see what's the impact.

I often see batches at 1000. It's neither too big (limited memory usage) nor too small (too many queries).

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: I tried to inline the calls, but I couldn't get any noticeable speedup. I also tried to be more idiomatic Erlang, with multiple definitions and using argument matchers, but it also didn't bring any improvement.

```elixir
  defp interactions_mode_label("mobileweb"), do: "Mobile Web"
  defp interactions_mode_label(mode), do: String.upcase(mode)

  defp action_type_label(nil), do: nil
  defp action_type_label("contact"), do: "Contact attempt"
  defp action_type_label(action), do: String.capitalize(action)

  defp disposition_label(nil), do: nil
  defp disposition_label(disposition), do: String.capitalize(disposition)
```

lib/ask_web/controllers/respondent_controller.ex
ysbaddaden: Answering myself: one results request was preloading the respondent questionnaires but it never used it, and actually used the questionnaires loaded from the survey. I also replaced

They don't bring noticeable speed improvements (returns just a few rows, always the same, already cached in MySQL), but it spares loooots of repeated requests.

ysbaddaden:
@leandroradusky I pushed a couple more commits to reduce the number of SQL requests, if you could give 'em a look :bow:

leandroradusky (approved):

matiasgarciaisaia:
Great! Can confirm 2.5x speedup, with same results, for interactions file 👍

Issue #1266, opened 2018-03-02
Texts should be left-aligned in Survey's Overview.


mmuller:
Version 6a459a3

The texts `Dispositions` and `3 weeks to complete the survey` are not left-aligned.

Design proposal:
https://projects.invisionapp.com/share/3U8BARPXJ#/screens/266351230_Cockpit_-_Simple

Current stg version:
![image](https://user-images.githubusercontent.com/518152/36918956-5aa28f72-1e3a-11e8-84d8-6b52d24b853c.png)

sardar-usman:
@nthiad It is working fine as expected.

PR #2254, opened 2023-04-18
Left-align for dispositions label and arrows in survey overview

nthiad:
This patch adjusts the alignment of arrows and the dispositions label from #1266.

However I couldn't find the time element to align, I think perhaps it was removed since the issue was posted in 2018.

Here's how the overview table looks with this patch:

dispositions

### DIFF
--- a/assets/js/components/surveys/SurveyShow.jsx
+++ b/assets/js/components/surveys/SurveyShow.jsx
@@ -527,17 +527,19 @@ class SurveyShow extends Component<any, State> {
     }
     const groupRow = (
       <tr key={group}>
-        <td>{dispositionGroupLabel(group)}</td>
+        <td className="expand-column">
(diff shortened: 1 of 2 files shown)


ysbaddaden (approved):
Thank you @nthiad 
I pushed a tiny errata and it's good to merge.

assets/js/components/surveys/SurveyShow.jsx
```diff
@@ -587,7 +588,7 @@ class SurveyShow extends Component<any, State> {
     return (
       <div className="card overflow">
         <div className="card-table-title">
-          <div>{t("Dispositions")}</div>
+          <div className="left-align">{t("Dispositions")}</div>
```
ysbaddaden, line 591:
```suggestion
          <div>{t("Dispositions")}</div>
```

PR #2262, opened 2023-04-20
Feature/2240 pause surveys

### LINKED ISSUE #2240: Evaluate feasibility of pausing surveys
Currently surveys cannot be paused, only stopped (cancelled)
We want to do a bit of research and see how we can implement a pause and resume actions in a survey.
We should evaluate some border cases like:
- What to do with active respondents that are answering once the survey is paused
- What to do with queued respondents that have contact attempts on channels control once the survey is pauses
- etc

leandroradusky:
Close #2240 

ccing @manumoreira & @matiasgarciaisaia here to check functionality is as expected and to have your opinion about the design and redirect to Jony if needed.

Now the surveys can be paused. Functionally, that means that the `Broker` won't poll new respondents. Polled respondents will continue with their normal behavior as if the survey is running.

First, some screenshots: 

- Paused surveys are displayed as paused within the survey detail:
image

- And in the survey cards:
image

- Paused surveys won't be counted as running surveys in the project index:
image

- The pause/play button was added next to the stop button (should it be green? Switch from green to orange depending on the state?), and will be shown and locked/unlocked with the same permission criteria:
image
image
image

I've tested the following:
- Pausing won't poll new respondents.
- Resuming works properly (with the simulators) 
- Resuming the day after works properly (with the simulators)
- Respondents completing paused surveys will follow its flow as if the survey is running.

**Important notes:**
- For the `ChannelStatusServer`, paused surveys are considered as part of the `Survey.running_channels()` set, the only thing we are preventing is polling new respondents, but things are still happening with the channels so paused surveys should have live channels to interact with. I've tested it and completed surveys ongoing for some respondents without any problem.
- **There is something that needs to be fixed**, please see the `autosave.js` modification that I had to introduce to get an idea: when I'm pausing or resuming a survey, it remains in `saving` state, product of the autosave, unless I do this trick. This is due to some wiring that I'm doing wrongly in the React code, but I couldn't devise what the problem is. I need your input here to help me understand what I'm doing differently with respect for example to `stop()` or `lock()` methods. The `pause` and `resume` functions in the `SurveyController` can be written better but I've left this verbose/redundant version to have the step-by-step very readable because they will need modifications to solve this issue. 
- I have scanned test files and have modified and introduced several tests. I think this covers the new functionality but please let me know if there is something you think I'm missing.

### DIFF
--- a/assets/js/components/surveys/SurveyShow.jsx
+++ b/assets/js/components/surveys/SurveyShow.jsx
@@ -8,7 +8,7 @@ import * as panelSurveysActions from "../../actions/panelSurveys"
 import SurveyStatus from "./SurveyStatus"
 import * as routes from "../../routes"
 import { Tooltip, Modal, dispositionGroupLabel, dispositionLabel } from "../ui"
-import { stopSurvey } from "../../api"
+import { stopSurvey, resumeSurvey, pauseSurvey } from "../../api"
 import sum from "lodash/sum"
 import { modeLabel } from "../../questionnaire.mode"
 import {
@@ -31,6 +31,7 @@ type State = {
   contacted: boolean,
   uncontacted: boolean,
   stopUnderstood: boolean,
+  paused: boolean,
 }
 
 class SurveyShow extends Component<any, State> {
@@ -70,12 +71,15 @@ class SurveyShow extends Component<any, State> {
       contacted: false,
       uncontacted: false,
       stopUnderstood: false,
+      paused: false,
     }
   }
 
   componentWillMount() {
-    const { dispatch, projectId, surveyId } = this.props
-    dispatch(actions.fetchSurveyIfNeeded(projectId, surveyId))
+    const { dispatch, projectId, surveyId, survey } = this.props
+    dispatch(actions.fetchSurveyIfNeeded(projectId, surveyId)).then((response) => {
+      this.setState({ paused: response.state == "paused" })
+    })
     dispatch(respondentActions.fetchRespondentsStats(projectId, surveyId))
     dispatch(actions.fetchSurveyStats(projectId, surveyId))
   }
@@ -89,7 +93,22 @@ class SurveyShow extends Component<any, State> {
 
   showHistograms() {
     const { survey } = this.props
-    return survey && survey.state == "running"
+    return survey && (survey.state == "running" || survey.state == "paused")
+  }
+
+  pauseOrResumeSurvey() {
+    const { dispatch, survey, projectId, surveyId, router } = this.props
+    if (survey && survey.state == "running") {
+      pauseSurvey(projectId, surveyId).then(() => {
+        dispatch(actions.pauseSurvey())
+        this.setState({ paused: true })
+      })
+    } else if (survey && survey.state == "paused") {
+      resumeSurvey(projectId, surveyId).then(() => {
+        dispatch(actions.resumeSurvey())
+        this.setState({ paused: false })
+      })
+    }
   }
 
   stopSurvey() {
@@ -191,7 +210,7 @@ class SurveyShow extends Component<any, State> {
       additionalCompletes,
       additionalRespondents,
     } = this.props
-    const { stopUnderstood } = this.state
+    const { stopUnderstood, paused } = this.state
 
     if (!survey || !cumulativePercentages || !questionnaires) {
       return <p>{t("Loading...")}</p>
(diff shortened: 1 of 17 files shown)


ysbaddaden:
I only did a quick overview. It looks great.

The `autosave` thingie is indeed weird. It feels like we don't want actual actions for pause & resume and only rely on autosave. It might be interesting to see how `start a survey` is working? But maybe start & stop redirect to another page, so they're not affected?

Maybe we don't need the `state.paused` attribute? but without it the UI may not be properly updated.

jkicillof:
The dates and status should be flex containers with `align-items: center` so the icon and text are aligned.
image

We should never have 2 FAB buttons, and the FAB button should be aligned like [here](https://projects.invisionapp.com/d/main#/console/8374235/266351230/preview). What's the difference between pause and stop? If stop can't be resumed we could have another button less accesible with a label that warns about what's going to happen and ask for confirmation after click. The only FAB should be pause in orange that toggles to play in green.

assets/js/components/surveys/SurveyShow.jsx
```diff
@@ -70,12 +71,15 @@ class SurveyShow extends Component<any, State> {
...
       stopUnderstood: false,
+      paused: false,
     }
   }
 
   componentWillMount() {
     const { dispatch, projectId, surveyId } = this.props
-    dispatch(actions.fetchSurveyIfNeeded(projectId, surveyId))
+    dispatch(actions.fetchSurveyIfNeeded(projectId, surveyId)).then((response) => {
+      this.setState({ paused: response.state == "paused" })
```
matiasgarciaisaia, line 81: Does it make sense to have a function that checks the survey's state instead of adding a new variable to the state? So you don't have to maintain it separately.

assets/js/components/surveys/SurveyShow.jsx
```diff
@@ -89,7 +93,22 @@ class SurveyShow extends Component<any, State> {
...
-    return survey && survey.state == "running"
+    return survey && (survey.state == "running" || survey.state == "paused")
+  }
+
+  pauseOrResumeSurvey() {
+    const { dispatch, survey, projectId, surveyId } = this.props
+    if (survey && survey.state == "running") {
+      pauseSurvey(projectId, surveyId).then(() => {
+        dispatch(actions.pauseSurvey())
+        this.setState({ paused: true })
```
matiasgarciaisaia, line 104: Definitely have to ditch the `paused` variable in the state and make it a function of the survey's state. You don't want to have to remember to update the `this.setState` every time you invoke an action that changes the survey state.

assets/js/components/surveys/SurveyWizardRespondentsStep.jsx
```diff
@@ -154,8 +154,8 @@ class SurveyWizardRespondentsStep extends Component {
 
     const { survey, t } = this.props
 
-    // If the survey is running, the only option is to add more respondents
-    if (survey.state == "running") {
+    // If the survey is running or paused, the only option is to add more respondents
```
matiasgarciaisaia, line 157: Would thinknig of an "active" state ("ogoing", or any other name) simplify things? Instead of constantly refering to "running or paused".

lib/ask/survey.ex
```diff
@@ -460,7 +462,7 @@ defmodule Ask.Survey do
   end
 
   def has_floip_package?(survey) do
-    survey.state == :running || survey.state == :terminated
+    survey.state == :paused || survey.state == :running || survey.state == :terminated
```
matiasgarciaisaia, line 465: Isn't this `launched?(survey)`?

lib/ask_web/controllers/survey_controller.ex
```diff
@@ -381,6 +381,68 @@ defmodule AskWeb.SurveyController do
...
+
+        _ ->
+          conn
+          |> render_with_links(survey)
+      end
+    else
+      conn
+      |> render_with_links(survey)
+    end
+  end
```
matiasgarciaisaia, line 444: These two methods are effectively the same one - with different values.

test/ask/runtime/broker_test.exs
```diff
@@ -1151,6 +1151,19 @@ defmodule Ask.Runtime.BrokerTest do
...
+      survey1 = insert(:survey, %{schedule: Schedule.always(), state: :running})
+
+      survey2 = insert(:survey, %{schedule: Schedule.always(), state: :paused})
+
+      Broker.handle_info(:poll, nil)
+
+      survey1 = Repo.get(Ask.Survey, survey1.id)
+      survey2 = Repo.get(Ask.Survey, survey2.id)
+      assert Ask.Survey.completed?(survey1)
+      assert survey2.state == :paused
```
matiasgarciaisaia, line 1165: Does this actually assert that the survey wasn't polled?

matiasgarciaisaia (commented):
Haven't run the code to see what's going on with the autosaving feature - but we can pair it to check it out if needed.

anaPerezGhiglia:
Closing this PR since it became stale and there we are not working on this issue

PR #2263, opened 2023-04-27
Rework GC of ChannelBroker

### LINKED ISSUE #2252: Regular check call/message state (to fix desynchronizations)
**Context**

I got the channel broker to be completely desynchronized from Verboice's queue. Surveda thought that the queue was filled, while the queue in Verboice got emptied. There may be various reasons, possibly non standard, but if they can happen in development, they may happen in production, and a survey going to a halt because of the channel broker is _very_ bad.

The active call GC won't help. It's meant to handle a few calls in desync, that won't affect the channel much, but here we're talking about a full desync (100 active vs 0 in reality) hanging up the whole channel :fearful:

**Proposal**

I'm proposing to implement a regular check of the real status of active contacts (calls/messages) to verify that the state in Surveda is still correct:

- yes: the active contact is really active/queued in the actual channel (allright);
- no: the active contact is no longer active/queued in the actual channel (oops, deactivate the contact).

Maybe we don't need to check _all_ active contacts, but those that have been pushed to the channel for some time (e.g. over 5 or 10 minutes, configurable). Hopefully, and after some fine tuning (channel capacity, active lifetime), it might never be called, except to fix a situation gone wrong.

We could also report when we fix a desync, reporting which calls needed to be corrected, so we could investigate from logs.

**Notes**

Once implemented, we can probably drop the active call GC for the Channel Broker. We can let the equivalent active call GC in Verboice stop calls when they take too long (likely the result of a bug), that will notify Surveda _or_ Surveda will find out during the next state check (at worst).

**Verboice** has an API endpoint to check the actual state of a call `/api/calls/:call_id/state`. We could enhance it to support reporting the state of multiple calls at once (to group http requests).

**Nuntium** doesn't have an API endpoint to check the state of AO messages, but we could add it. Same it could support checking multiple AO messages at once.

Both Verboice and Nuntium are old Ruby on Rails applications. It shouldn't be hard to enhance / add these endpoints.

TODO:
- [x] CHNLBRK: save the `call_id` in channel broker when making an IVR call to Verboice;
- [x] CHNLBRK: save the AO message token (`x-nuntium-token`) in channel broker when sending SMS messages to Nuntium;
- [ ] Verboice: add or modify endpoint to get the state of multiple calls at once (optional);
- [x] Nuntium: add an endpoint to get the state of one (and/or multiple) AO message(s);
- [x] CHNLBRK: configuration to tell how often we must verify the active contacts, and how long a contact can stay idle;
- [x] CHNLBRK: new process to regularly check the state of "active contacts" that have been idle for too long:
  - [x] Verboice: check the state of the call;
  - [x] Nuntium: check the state of the set messages;
    - [x] Bug: check the state for _each_ message in the set (and unlock all slots at once);
- [x] Verboice progress callback: notify channel broker of in-progress (to touch the last updated time);
- [x] CHNLBRK: ~~drop the active call GC~~ revamp the GC to check for idle time.
- [x] Tests:
  - [x] Verboice
  - [x] Nuntium

Where to set the envvar? project settings? channel settings?

ysbaddaden:
Revamps the GC to regularly check for the actual state of idle active contacts on the external channel (Verboice or Nuntium).

Also extracts the `Ask.RunTime.ChannelBrokerState` out of the `Ask.Runtime.ChannelBroker` to separate the state operations from the channel broker logic.

Depends on https://github.com/instedd/nuntium/pull/98
Closes #2252

### DIFF
--- a/lib/ask/config.ex
+++ b/lib/ask/config.ex
@@ -66,6 +66,7 @@ defmodule Ask.Config do
       # lost.
       # To disable saving to DB, set this value to 0.
       to_db_operations: env_to_int("CHNL_BKR_TO_DB_OPERATIONS", 0),
+
       # How many minutes the ChannelBroker garbage collector will wait between rounds. Default: 10
       # minutes.
       # Two situations brings the need of running the GC frequently:
@@ -76,7 +77,8 @@ defmodule Ask.Config do
       # High values: the GC will run less frequently. A too high value could lead to unnecessary
       # stops of the survey by reaching the channel capacity because of failed contacts or
       # callback losses.
-      gc_interval_minutes: env_to_int("CHNL_BKR_GC_INTERVAL_MINUTES", 10),
+      gc_interval_minutes: env_to_int("CHNL_BKR_GC_INTERVAL_MINUTES", 2),
+
       # How many hours after the last contact the ChannelBroker waits until a contact is discarded
       # because it's considered outdated. Default: 24 h.
       # Low values: the GC will discard younger contacts. A too low value could lead to discard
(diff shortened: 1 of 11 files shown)


leandroradusky:
Also, I wanted to review the test not passing that still needs to be fixed.

ysbaddaden:
@leandroradusky This is indeed really weird. We should get a batch of respondents added to the pqueue every minute. I'll have a look.

Issue #2133, opened 2022-08-18
Simplify ChannelBroker state persistence


aeatencio:
Save to DB on change and recover from DB on init
Reason: early optimization with difficult testing

ysbaddaden:
**Alternative**: disable the state persistence. We'll see if we eventually rework the state to live in the database (so we can have multiple servers handling the queues). Maybe we'd only have to return the state when the channel broker crashes and feed it back when we restart it (full memory).

leandroradusky:
The actual implementation saves to DB only after a series of operations (actually, every 100 operations & on timeout).
The supervisor recovers the state (`:one_for_one` strategy) from memory if it is there, from DB otherwise, else from scratch. 

After some testing, this appears to be working ok. We can play with the numbers to make this work as desired (i.e. never save to DB if that is what we want) but following this thread appears that what we have is OK.

ysbaddaden:
Current action: let's disable saving for now. I had issues on local when stopping surveda while a survey is running and the channel broker would crash on restart (taking surveda down).

PR #2273, opened 2023-05-15
Fix: recover channel broker state from agent after crash

### LINKED ISSUE #2266: ChannelBroker crashes resets the internal state
When a ChannelBroker crashes, its internal state is lost and the new instance will be started with an empty state, but the state should have been saved by ChannelBrokerAgent in memory and should be reusable.

We must verify what's happening:
- maybe disabling the save to DB disabled the save to memory?
- maybe a ChannelBroker crash causes the whole ChannelBrokerSupervisor to restart?

ysbaddaden:
See individual commits for more details.

The *first commit* removes the code to save the channel broker state to the database (as we don't need it now), the associated database table and Ecto model. Related to #2133.

The *second commit* fixes #2266 by moving the state recovery to `ChannelBroker.init`. After testing with a chaos monkey, the `DynamicSupervisor.start_link` method isn't invoked when the channel broker crashes (maybe when the supervisor itself crashes), but the `ChannelBroker.init` is always called.

The *last two commits* try to fix race conditions when running the test suite on CI.

Example **chaos monkey**, to place anywhere you want to have a random crash:

```elixir
if :rand.uniform(5) == 1 do
  raise "chaos"
end
```


lib/ask/config.ex
```diff
@@ -77,7 +66,7 @@ defmodule Ask.Config do
       # High values: the GC will run less frequently. A too high value could lead to unnecessary
       # stops of the survey by reaching the channel capacity because of failed contacts or
       # callback losses.
-      gc_interval_minutes: env_to_int("CHNL_BKR_GC_INTERVAL_MINUTES", 2),
+      gc_interval_minutes: env_to_int("CHNL_BKR_GC_INTERVAL_MINUTES", 1),
```
ysbaddaden, line 69:
```suggestion
      gc_interval_minutes: env_to_int("CHNL_BKR_GC_INTERVAL_MINUTES", 2),
```

lib/ask/config.ex
```diff
@@ -91,7 +80,7 @@ defmodule Ask.Config do
 
       # How long after a call/message has been left idle (no callbacks received)
       # before the GC starts polling the remote channel for its actual state.
-      gc_active_idle_minutes: env_to_int("CHNL_BKR_GC_IDLE_MINUTES", 2)
+      gc_active_idle_minutes: env_to_int("CHNL_BKR_GC_IDLE_MINUTES", 1)
```
ysbaddaden, line 83:
```suggestion
      gc_active_idle_minutes: env_to_int("CHNL_BKR_GC_IDLE_MINUTES", 5)
```

lib/ask/runtime/channel_broker.ex
```diff
@@ -348,14 +342,12 @@ defmodule Ask.Runtime.ChannelBroker do
...
   @impl true
-  def handle_info(
-        :timeout,
-        %{channel_id: channel_id, config: %{to_db_operations: to_db_operations}} = state
-      ) do
-    info("handle_info[timeout]", channel_id: channel_id, to_db_operations: to_db_operations)
+  def handle_info(:timeout, %{channel_id: channel_id} = state) do
+    info("handle_info[timeout]", channel_id: channel_id)
+
+    # TODO: delete state from agent if both pqueue & active_contacts are empty!
```
leandroradusky, line 348: This `TODO` seems easily solvable, `pqueue` and `active_contacts` can be checked here and then call `ChannelBrokerAgent.clear()`. Following the code the recovery should still work ok, there's something I'm missing?

leandroradusky (approved):
Looks good to me! I just left a small comment.

lib/ask/runtime/channel_broker.ex
ysbaddaden: Yes. I'm wondering if there could be any race condition, but we only ever read/update/save the state from the channel broker process which should never run in parallel: each message should be handled sequentially.

We should however only delete the state for the channel, not clear the agent :)

PR #2280, opened 2023-05-26
Fix: dont cache channel in queue (channel broker)

### LINKED ISSUE #2279: Don't save runtime channel for queued contacts
We save the runtime channel into the channel-broker state when we queue the contact, which in turn means that we save the channel client with the oauth token, but the token is only refreshed when we create the runtime channel (it requests the access token from Ask.OAuthTokenServer that will load the token from the database & refresh if it's about to expire).

It was fine until the channel-broker, because the runtime channel & client were created and then the calls were pushed within the 15 minutes that the token is valid. Subsequent calls usually re-create the runtime channel on the fly.

**Quick fix:**

- [x] remove the channel from the `ChannelBroker.setup/7` and `ChannelBroker.ask/6` calls (not needed).
- [x] don't save the channel when we queue a contact (not needed);
- [x] create a runtime channel in `ChannelBroker.activate_next_queued_contact/1` before each call to `Ask.Runtime.Channel.ivr_call/6` and `Ask.Runtime.Channel.ask/4`.

**Ideally**, we'd introduce an `Ask.Runtime.ChannelAgent` process to create/save the runtime channels directly, and that would take care of refreshing the oauth token when needed. We'd then have `Ask.Channel.runtime_channel/1` interact with this agent, which should significantly improve performance.

ysbaddaden:
The individual commits show some of my failed attempts to fix the issue _and_ keep the test suite happy. Some attempts are missing (e.g. database cleaner approach).

I had to fully remove any reference to Ask.Runtime.Channel from outside of Ask.Runtime.ChannelBroker to get the test suite happy without introducing race conditions. The channel broker is now fully responsible for creating Ask.Runtime.Channel instances and refreshing the oauth token before it expires.

The Ask.Runtime.Channel instance is also loaded _before_ spawning the ChannelBroker processes to keep the test suite happy and avoid a deadlock situation with Ecto.Adapters.SQL.Sandbox shared mode. For example: Ask.Runtime.Session starts a DB transaction then a nested function starts a channel broker in a new process that can't checkout the DB connection because it's already been checkout by the initial process (deadlock). This is a test-only bug.

I took on the opportunity to change some calls in the channel broker to be asynchronous (`GenServer.cast`) instead of synchronous (`GenServer.call`). They don't expect any reply, and I thought at some point that it would help with my deadlock struggles (it didn't). I kept the change as I think it's good.

I leave the PR as `draft` because I didn't run a manual test to verify that the oauth token (I'll see that after the weekend). The PR can nonetheless be reviewed.

fixes #2279


lib/ask/o_auth_token.ex
```diff
@@ -37,4 +37,15 @@ defmodule Ask.OAuthToken do
   def access_token(token) do
     Poison.Decode.decode(token.access_token, as: %OAuth2.AccessToken{})
   end
+
+  def about_to_expire?(token) do
```
leandroradusky, line 41: Just asking here... Why to change the `Timex` implementation of 2 lines for this one?

lib/ask/runtime/channel_broker_agent.ex
```diff
@@ -27,7 +27,7 @@ defmodule Ask.Runtime.ChannelBrokerAgent do
 
   def save_state(%State{channel_id: channel_id} = state) do
     Agent.update(__MODULE__, fn agent ->
-      Map.put(agent, channel_id, state)
+      Map.put(agent, channel_id, Map.delete(state, :runtime_channel))
```
leandroradusky, line 30: Asking here again because is not clear to me: this new implementation does not do the same as before and it is not clear to me why we want to remove the runtime channel now (and not before).

See for example `ChannelBroker.refresh_runtime_channel(state)`

We return (when channel is about to expire):

```
state
      |> Map.put(:runtime_channel, new_runtime_channel)
      |> Agent.save_state()
```

We put the runtime channel in the state and then we remove it?

leandroradusky (approved):
Left a few questions, besides that, looks good to me :)

lib/ask/o_auth_token.ex
ysbaddaden: The Timex library isn't deprecated, but Elixir added lots of support for dates, times and timezones right into the stdlib, and Timex mostly delegates to these types, now. The [library maintainers](https://github.com/bitwalker/timex#migrating-to-timex-3x) state that themselves.

lib/ask/runtime/channel_broker_agent.ex
ysbaddaden: Urgh, you got a bug! I only want to remove it from the saved state in the agent, so that when we restore the state, we create a new one.

But the `ChannelBrokerAgent.save_state/1` method should return the pristine `state` object, indeed. Right now we probably reload the runtime_channel _every_ call to the ChannelBroker :facepalm:

ysbaddaden: @leandroradusky Ah no, that's OK: the method returns the original `state`. So I answered your question above: no need to save the `runtime_channel` it's not important, and in case of a channel-broker crash we only need to restore the active/queued contacts, and we can reset the rest.

PR #2297, opened 2023-08-28
Keep ChannelBrokerState in the database

ysbaddaden:
This allows to resume operations as if (almost) nothing happened when Surveda restarts.

The state is now saved in the database on each operation to the ChannelBrokerState, instead of fully kept in memory. If Surveda crashes or needs to be restarted, we don't lose anything, and don't cause surveys to hang because the survey broker activated enough respondents, but we lost the queue because of a restart. That should also reduce the memory usages, at the expense of more interaction with MySQL.

Improves the test suite with tests for the Ask.Runtime.ChannelBrokerState module to ascertain its individual behavior. The behavior got changed in a few places to accommodate the new design yet still be valid if we change the data store (again).

I replaced the default Ecto.Adapters.SQL.Sandbox pool used in the test suite (Ecto default) with a DatabaseCleaner solution. The sandbox only allows a single connection to be checkout for transactional tests, but it was breaking the test suite: there are long transactions in SurveyBroker and Session that call into the ChannelBroker process that got blocked because the process failed to checkout the unique connection (already checkout by another process in a blocking manner). The DatabaseCleaner solution is only marginally slower, and could be faster by using the MyXQL driver directly.

:warning: I wanted to have only one row per channel & respondent (composite primary key of `channel_id` and `respondent_id`), but that failed quickly in the test suite that tries to queue the same respondent multiple times (:sweat:). I fixed the test suite by adopting an UPSERT strategy: whenever there is a conflict, the new contact will deactivate and replace the existing contact in the queue.

- In theory this may be incorrect: replacing a queued contact for another queued contact seems fine (forget about the previous attempt) but deactivating an active contact may be overboard?

- In practice this may not be an issue: the runtime first asks the channel-broker whether a message is already queued (`has_queued_message?`) or expired (``) before timing out an attempt (see `Ask.Runtime.SurveyBroker.retry_respondent/1` and `Ask.Runtime.Session.timemout/1`).


lib/ask/channel_broker_queue.ex
```diff
@@ -0,0 +1,92 @@
...
+    ])
+    |> validate_required([
+      :channel_id,
+      :respondent_id,
+      :queued_at,
+      :priority,
+      :size,
+      :token
+    ])
+    # |> assoc_constraint(:channel, :respondent)
```
ysbaddaden, line 66: I can't validate the channel or the respondent because they're often live objects (not stored in the database) in the test suite.

lib/ask/runtime/channel_broker_state.ex
```diff
@@ -80,201 +58,234 @@ defmodule Ask.Runtime.ChannelBrokerState do
...
+  # OPTIMIZE: Cache the number of active contacts into the state to heavily
+  #           reduce the number of COUNT(*) queries that tend to be slow.
+  #
+  #           on load: count from the database & cache the value
+  #           on activate: increment the number (by size)
+  #           on increment/decrement/deactivate/reenqueue: update the count
   defp under_capacity?(state) do
-    count_active_contacts(state) < state.capacity
+    Queue.count_active_contacts(state.channel_id) < state.capacity
   end
```
ysbaddaden, line 281: This method may have a negative impact on performance: we're counting how many contacts are active for the channel right from the database, and we count every time we try to activate a contact.

We could keep a counter in `state` for the number of active contacts, and manipulate it as we activate / increment / decrement / deactivate / reenqueue, instead of counting all the time. We'd only count from the database on `init/1`.

lib/ask/runtime/channel_broker_state.ex
```diff
@@ -284,45 +295,64 @@ defmodule Ask.Runtime.ChannelBrokerState do
...
+
+    Repo.all(query)
+    |> Enum.each(fn active_contact ->
+      if Ask.Runtime.Channel.message_inactive?(
+           state.runtime_channel,
+           active_contact.channel_state
+         ) do
+        Repo.delete(active_contact)
+      end
+    end)
```
ysbaddaden, line 335: Note: we can't stream from the database as Ecto streams require to live inside a transaction, and we don't want to keep another long transaction open while we make external calls to verboice/nuntium, and we don't want to revert all the deletions because of a network error.

We could however load in batches from the database.

lib/ask/runtime/channel_broker_state.ex
```diff
@@ -284,45 +295,64 @@ defmodule Ask.Runtime.ChannelBrokerState do
...
-          end
-        end,
-        state.active_contacts
-      )
+    last_contact = SystemTime.time().now |> DateTime.add(-idle_time, :second)
+
+    query =
+      from q in Queue,
+        select: [:channel_id, :respondent_id, :channel_state],
+        where: q.channel_id == ^state.channel_id and q.last_contact < ^last_contact
```
ysbaddaden, line 325: Hum, maybe we could `ORDER BY last_contact ASC` to process the oldest idle contacts first?

test/ask/runtime/channel_broker_state_test.exs
```diff
@@ -1,49 +1,436 @@
...
+    {:ok, state: State.new(0, "ivr", %{"capacity" => 4})}
+  end
+
+  describe ".inactive?" do
+    test "returns true when empty", %{state: state} do
+      assert state
+             |> State.inactive?()
+    end
+
+    test "returns true if any pending contact", %{state: state} do
```
matiasgarciaisaia, line 19:
```suggestion
    test "returns false if any pending contact", %{state: state} do
```

test/ask/runtime/channel_broker_state_test.exs
```diff
@@ -1,49 +1,436 @@
...
+      refute state
+             |> State.queue_contact(new_contact(2), 1)
+             |> State.inactive?()
     end
 
-    {:ok, state: State.new(0, %{}), mock_queued_contact: mock_queued_contact}
+    test "returns true if any active contact", %{state: state} do
```
matiasgarciaisaia, line 25:
```suggestion
    test "returns false if any active contact", %{state: state} do
```

test/ask/runtime/channel_broker_state_test.exs
```diff
@@ -1,49 +1,436 @@
...
+             |> State.queue_contact(new_contact(2), 1)
+             |> State.activate_next_in_queue()
+             |> elem(0)
+             |> State.queued_or_active?(2)
+    end
 
-    {_, [queud_size, queued_contact]} = Ask.PQueue.pop(q)
-    assert queued_contact == {%{id: respondent_id, disposition: disposition}, params}
-    assert queud_size == size
+    test "returns false otherwise", %{state: state} do
```
matiasgarciaisaia, line 49:
```suggestion
    test "returns false when empty", %{state: state} do
```

matiasgarciaisaia (commented):
I still have to finish reviewing the `channel_broker_state_test.exs` file.

lib/ask/runtime/channel_broker.ex
```diff
@@ -355,27 +349,27 @@ defmodule Ask.Runtime.ChannelBroker do
...
 
-    active_respondents =
-      from(r in "respondents",
-        where: r.id in ^Map.keys(state.active_contacts) and r.state == "active",
-        select: r.id
+    active_respondent_ids =
+      Repo.all(
+        from r in "respondents",
+          select: r.id,
+          where: r.id in ^State.active_respondent_ids(state) and r.state == "active"
```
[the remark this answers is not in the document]
ysbaddaden: This is already limited to the channel capacity (e.g. 200) not the batch size (e.g. 20000) since the query restricts to respondents that the channel believes are active.

Maybe we could reverse the logic to return inactive respondents instead of active ones? We expect there won't be less active than we think there are. Then we could paginate and clean-up the respondents from the channel-broker-queue in multiple passes. :thinking:

lib/ask/runtime/channel_broker_state.ex
```diff
@@ -80,201 +58,234 @@ defmodule Ask.Runtime.ChannelBrokerState do
...
+
+    state
+  end
+
+  def put_channel_state(state, respondent_id, channel_state) do
+    query =
+      from q in Queue,
+        where: q.channel_id == ^state.channel_id and q.respondent_id == ^respondent_id
+
+    Repo.update_all(query, set: [channel_state: channel_state])
```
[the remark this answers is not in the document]
ysbaddaden: Hum? This is already a single UPDATE. `query` is just the pending SQL statement.

Or do you mean the overall logic:

1. queue (upsert)
2. activate (update × 1) -> before request to verboice/nuntium
3. put channel state (update × 2) -> after request to verboice/nuntium
4. each callback (update × n)
5. final callback (delete)

lib/ask/runtime/channel_broker_state.ex
ysbaddaden: Yes, there are a couple indexes:

- PRIMARY KEY (channel_id, respondent_id)
- INDEX (last_contact)

I tried to avoid aggregate methods everywhere, preferring the quick `Repo.exists?`, but the capacity check is the only request where I needed an actual value.

I didn't implement the optimization because keeping a cache is easy, but keeping it correctly updated is another matter.

lib/ask/runtime/channel_broker_state.ex
```diff
@@ -284,45 +295,64 @@ defmodule Ask.Runtime.ChannelBrokerState do
...
   end
 
-  # For leftover active contacts, we ask the remote channel for the actual IVR
-  # call or SMS message state. Keep only the contacts that are active or queued.
+  # For idle contacts, we ask the remote channel for the actual IVR call or SMS
+  # message state. Keep only the contacts that are active or queued.
+  #
+  # FIXME: this may take a while, and during that time the channel broker
+  #        won't process its mailbox, if it ever becomes a problem, we might
+  #        consider:
```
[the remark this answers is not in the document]
ysbaddaden: We could spawn a Task to run the checks in parallel and have a new message to handle each response individually (e.g. `:remove_inactive_respondent`) :thinking:

priv/repo/structure.sql
```diff
@@ -1204,4 +1228,6 @@ INSERT INTO `schema_migrations` (version) VALUES (20220131103226);
 INSERT INTO `schema_migrations` (version) VALUES (20230217143550);
 INSERT INTO `schema_migrations` (version) VALUES (20230317094712);
 INSERT INTO `schema_migrations` (version) VALUES (20230402091100);
+INSERT INTO `schema_migrations` (version) VALUES (20230405111657);
```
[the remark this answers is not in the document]
ysbaddaden: Most likely because it didn't change the schema (only a data migration):

https://github.com/instedd/surveda/blob/main/priv/repo/migrations/20230405111657_remove_project_memberships_duplicates.exs

lib/ask/runtime/channel_broker_state.ex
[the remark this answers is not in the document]
matiasgarciaisaia: I was refering to the `Repo.update_all` instead of doing a `Repo.update` or adding a `LIMIT 1`.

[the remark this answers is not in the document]
ysbaddaden: Oh! I use `Repo.update_all()` for a direct SQL request, because `Repo.all` requires a changeset, and thus require to load -> changeset -> update. I could had `LIMIT 1` but we're already updating by the primary key (composite channel_id, respondent_id), so there will only be one row at most.

matiasgarciaisaia (approved):
I think we could improve the naming of some models, and there are some rough edges to polish, but overall the PR is OK 👍

PR #2298, opened 2023-08-29
Fix: channel broker must report expiration to survey

ysbaddaden:
The channel broker no longer sends all messages to the external service, so when contacts expire, we'll no longer receive an expiration notice from the service.

The channel broker thus needs to report the expiration internally, so the respondents can be properly rescheduled properly.

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -426,6 +426,7 @@ defmodule Ask.Runtime.ChannelBroker do
       {respondent, token, not_before, not_after} ->
         cond do
           expired_call?(not_after) ->
+            Ask.Runtime.Survey.contact_attempt_expired(respondent)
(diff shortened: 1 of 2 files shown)


lib/ask/runtime/channel_broker.ex
```diff
@@ -426,6 +426,7 @@ defmodule Ask.Runtime.ChannelBroker do
       {respondent, token, not_before, not_after} ->
         cond do
           expired_call?(not_after) ->
+            Ask.Runtime.Survey.contact_attempt_expired(respondent)
```
ysbaddaden, line 429: I'm just calling here what `Ask.Runtime.VerboiceChannel.callback/2` calls when a message has expired.

matiasgarciaisaia (approved):
Great one!

As we discussed, this whole "let the survey broker that the call expired to extend its expiration date" seems like a bit of a hack, but it's the way we're currently doing it 👍

PR #2314, opened 2024-03-19
Upgrade ngrok image in docker-compose

anaPerezGhiglia:
### Context
Current ngrok accounts are not allowing old versions of ngrok. When trying to run the old ngrok with a new token the app wouldn't start and fail with
> Your ngrok-agent version "2.3.40" is too old. The minimum supported agent version for your account is "3.2.0". Please update to a newer version

### Solution
Replace old `wernight/ngrok` image with `ngrok/ngrok:3.8.0-alpine` image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,9 @@ erl_crash.dump
 # variables.
 /config/prod.secret.exs
 
+# Local configuration
 /config/local.exs
+/docker-compose.override.yml
 
 .vagrant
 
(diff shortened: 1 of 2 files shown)


ysbaddaden (approved):
Exactly the things we didn't notice before because we used a fully local setup.

matiasgarciaisaia (approved):
👍 :shipit:

Issue #2300, opened 2023-08-29
Keep the ChannelBroker state in the database


ysbaddaden:
We have issues with the channel broker in production where the SurveyBroker already activated enough respondents, but the ChannelBroker queue was lost because of a restart (or multiple restarts): the survey hangs up (until the contact eventually expire).

We should save the state in the database in order for the channel broker operations to be resumed (almost) as if nothing happened on restarts.

PR #2315, opened 2024-03-22
ChannelBroker: small improvements

### LINKED ISSUE #2289: Channel Broker: re-enqueue :low contacts as per the respondent's disposition
The reasoning for the `:low` level was to avoid to dequeue just to reenqueue the contact because `not_before` isn't reached, which would impact performance if we have another survey using the the channel (on different schedules).

But when the survey is resumed on the following day, it'll push new contacts as `:normal` and they'd take precedence over the `:low` messages, whatever the respondent's disposition. Hum :thinking:

The channel-broker-gc now runs often (every 2 minutes by default) and it queues contacts 10 minutes before `not_before`, so contacts should get queued before the survey-broker starts pushing new contacts if the channel isn't used by another survey, but if the channel is used by another survey, they'll stay in `:low`. Damn.

Maybe we could add a pass to the channel-broker-gc to dequeue/re-enqueue messages as per the respondent's disposition when `not_before` is reachable?

anaPerezGhiglia:
- Remove `:low` priority since it doesn't make sense anymore. The piece of code where it was being considered is now unreachable after #2300 
   ```
      future_call?(not_before) ->
            State.reenqueue_contact(new_state, respondent.id, :low)

        defp future_call?(not_before) do
    
            DateTime.compare(not_before, Ask.SystemTime.time().now |> DateTime.add(60, :second)) == :gt
        end
    ```
    
    ☝️ The thing is that now (with the queue stored in the db) we only activate respondents which contact 
    ```
    not_before <= not_before = SystemTime.time().now |> DateTime.add(60, :second)
    ```
    See [`ChannelBrokerState.activate_next_in_queue`](https://github.com/instedd/surveda/blob/683a050b2aeec1a6bfafaa65f481f1f5e1c8ab72/lib/ask/runtime/channel_broker_state.ex#L166)
    
    this closes #2289 
    
 - Invert calls to `Survey` and `ChannelBrokerState` when the `ChannelBroker` tries to activate the next queued contact and it's expired
    - See [this slack thread](https://manas.slack.com/archives/C1L7CJMNU/p1710878095079079) for more context 

## Observations 👀
I had to do a workaround for inverting the calls. I had to keep the `state` in a variable and return that after the call to `Survey.contact_attempt_expired`. 
I'm not fan of this solution, but I _think_ I prefer this to the previous version, but I read your opinions!

### DIFF
--- a/lib/ask/channel_broker_queue.ex
+++ b/lib/ask/channel_broker_queue.ex
@@ -12,7 +12,7 @@ defmodule Ask.ChannelBrokerQueue do
 
     # queued (pending):
     field :queued_at, :utc_datetime
-    field :priority, Ecto.Enum, values: [low: 2, normal: 1, high: 0]
+    field :priority, Ecto.Enum, values: [normal: 1, high: 0]
     # number of outgoing messages (ivr=1, sms=1+)
     field :size, :integer
     field :token, :string
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -420,8 +420,8 @@ defmodule Ask.Runtime.ChannelBroker do
       {respondent, token, not_before, not_after} ->
         cond do
           expired_call?(not_after) ->
-            Ask.Runtime.Survey.contact_attempt_expired(respondent)
             State.deactivate_contact(new_state, respondent.id)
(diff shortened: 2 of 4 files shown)


ysbaddaden (approved):

matiasgarciaisaia (approved):
👍

PR #2316, opened 2024-03-27
#2306: prioritize contacted respondents

### LINKED ISSUE #2306: ChannelBroker: add a new case to prioritize contact attempts
## The new case
Currently, we prioritize the contact attempts of the respondents with `disposition != queued`.
This is enough for SMS but not for IVR,  to fully guarantee the retry window we should add another case based on the number of previous attempts. If previous attempts > 0 then it should be prioritized.

## Things to consider
~This might affect the size of the prioritized contact attempts queue in the channel broker impacting negatively in other respondents that should have a higher priority (like the SMS respondents that have just replied to a question), let's keeo an eye on this when testing.~

anaPerezGhiglia:
## Changes

* move responsibility of defining a contact priority from `ChannelBrokerState` to `ChannelBroker`
   * `ChannelBrokerState` should not be deciding how to queue respondents. It's something that's outside it responsibility. The one that _knows_ about respondents is `ChannelBroker`
* Enqueue contacts with `priority: :high` if surveda has already _bothered_ that respondent
   * We are checking `Stats.attempts :full` because 
      * we want to count any mode attempt 
      * we want to consider ivr pending calls as well
   * When `runtime/Session` starts a respondent, already adds the mode attempt. This happens before the respondent gets queued to the `ChannelBrokerQueue` - see [`runtime/Session.run_flow`](https://github.com/instedd/surveda/blob/c2b39f6eb1c2c12c79fac2f843b7acdce97b4e84/lib/ask/runtime/session.ex#L595). This means that for distinguishing respondents that are being queued for the first time or not, we have to check for `attempts > 1`

closes #2306

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -501,10 +501,13 @@ defmodule Ask.Runtime.ChannelBroker do
     priority = cond do
       # Respondent is participating in the survey
       prioritized_disposition?(respondent.disposition) -> :high
+
       # We have already 'bothered' this respondent
-      # attempts > 1 since when queuing the respondent the first time
-      # it already has `attempts: 1`
+      # we check that attempts > 1 since
+      # when queuing the respondent the first time
+      # already has `attempts: 1` (see Session.run_flow)
       Stats.attempts(respondent.stats, :full) > 1 -> :high
+
       true -> :normal
     end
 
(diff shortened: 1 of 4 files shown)


ysbaddaden (approved):
LGTM!

matiasgarciaisaia (commented):
👍

PR #2317, opened 2024-04-05
Log batch limit when activating respondents

matiasgarciaisaia:
If the SurveyBroker wants to activate more respodents than allowed by the batch limit per minute, log the fact that it was capped.

Before this commit, the limit was applied without logging it, making it harder to understand what was going on.

### DIFF
--- a/lib/ask/runtime/survey_broker.ex
+++ b/lib/ask/runtime/survey_broker.ex
@@ -173,7 +173,6 @@ defmodule Ask.Runtime.SurveyBroker do
 
         active < batch_size && pending > 0 && !survey_completed ->
           count = batch_size - active
-          Logger.info("Survey #{survey.id}. Starting up to #{count} respondents.")
           start_some(survey, count)
 
         true ->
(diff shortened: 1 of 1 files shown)


lib/ask/runtime/survey_broker.ex
```diff
@@ -372,7 +371,13 @@ defmodule Ask.Runtime.SurveyBroker do
   end
 
   defp start_some(survey, count) do
-    count = Enum.min([batch_limit_per_minute(survey.project), count])
+    count_limit = batch_limit_per_minute(survey.project)
+    if count > count_limit do
+      Logger.info("Survey #{survey.id}. Starting up to #{count_limit} respondents due to batch limit (wanted to start #{count}).")
```
anaPerezGhiglia, line 376: do we want to log the limit as well to have all the info in one log?

lib/ask/runtime/survey_broker.ex
```diff
@@ -372,7 +371,13 @@ defmodule Ask.Runtime.SurveyBroker do
...
+      count = count_limit
```
anaPerezGhiglia, line 377: the compiler is complaining about this. I think that we should do something like this instead 
```elixir
count = if count > count_limit do
...
end
```

anaPerezGhiglia (approved):
LGTM (besides the little warning)

matiasgarciaisaia:
I've now tested it and the logs end up like this:

```
2024-04-17T19:55:39.331Z [info] Polling survey 2 (active=2, pending=2, completed=0, batch_size=100)
2024-04-17T19:55:39.331Z [info] Survey 2. Starting up to 2 respondents (wanted to start 98, batch limit at 2).
```

PR #2319, opened 2024-04-10
#2318: make numeric ranges quotas exhaustive

### LINKED ISSUE #2318: Integer variable quota buckets should be exhaustive
We've seen #1471 happen time and time again: when using quotas, if a respondent doesn't match any of the existing quotas, the survey progress calculation gets broken.

When setting the quota buckets (as of 0.37.2), Surveda asks the survey editor for the buckets limits of integer variables, and creates ranges between them. For example, for an `Age` variable with range `0-110`, the delimiters `18,30,45,100` yield the buckets `18-29`, `30-44`, `45-100`. **The values <18 and >100 don't belong to any bucket.**

Instead, Surveda should guarantee every value belongs to a bucket, creating extra buckets for the values below the minimum delimiter, and above the maximum. So, in this case, the resulting buckets should be `<18`, `18-29`, `30-44`, `45-100`, `>100`.

We should probably avoid making the implicit boundaries of the intervals explicit (ie, make the first bucket `0-17`) because they can change in the quex before the survey is started (and, also, because the user doesn't care about them - if they would care, they would include that value as a delimiter).

anaPerezGhiglia:
## Changes
Create upperbound and lowerbound ranges to make numeric quotas exhaustive.

### Examples
image

### Evidence

https://github.com/instedd/surveda/assets/13237343/72c403e1-10d6-4aa4-a1b2-7e0b4b446de4

image

closes #2318

### DIFF
--- a/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
+++ b/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
@@ -77,9 +77,9 @@ class SurveyWizardCutoffStep extends Component {
       bucket.condition.map(({ store, value }) => {
         if(Array.isArray(value) && value.length == 2){
           if(value[0] == null){
-            value = `<= ${value[1]}`
+            value = `< ${value[1] + 1}`
           } else if(value[1] == null){
-            value = `>= ${value[0]}`
+            value = `${value[0]}+`
           } else {
             value = join(value, " - ")
           }
--- a/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
+++ b/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
@@ -75,8 +75,14 @@ class SurveyWizardCutoffStep extends Component {
   bucketLabel(bucket) {
     return join(
       bucket.condition.map(({ store, value }) => {
-        if (typeof value == "object") {
-          value = join(value, " - ")
+        if(Array.isArray(value) && value.length == 2){
+          if(value[0] == null){
+            value = `<= ${value[1]}`
+          } else if(value[1] == null){
+            value = `>= ${value[0]}`
+          } else {
+            value = join(value, " - ")
+          }
         }
         return `${store}: ${value}`
       }),
--- a/lib/ask/quota_bucket.ex
+++ b/lib/ask/quota_bucket.ex
@@ -50,6 +50,28 @@ defmodule Ask.QuotaBucket do
     end
   end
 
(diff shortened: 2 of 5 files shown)


lib/ask/quota_bucket.ex
```diff
@@ -50,6 +50,28 @@ defmodule Ask.QuotaBucket do
     end
   end
 
+  def matches_condition?(value, [from, nil]) do
+    from = if is_integer(from), do: from, else: String.to_integer(from)
```
ysbaddaden, line 54: **nitpick:** maybe leverage guards?

```elixir
def matches_condition?(value, [from, nil]) when is_integer(from) do
  # ...
end

def matches_condition?(value, [from, nil]) do
  matches_condition?(value, [String.to_integer(from), nil])
end
```

ysbaddaden (approved):

lib/ask/quota_bucket.ex
anaPerezGhiglia: I though about it, but I decided not to. 
IMHO in this particular case is clearer to have it this way. The code needed for converting the param to int is "small" and does not pollute the main logic. I think is more succinct to have the guards to identify _actual_ logic differences in this case. 
But I would love to read you if you have a strong opinion why not to 🤓

ysbaddaden: No strong opinion, it's a nitpick (that has zero impact on performance), but it feels more elixir/erlang-y that way, and... the function itself already does just that :shrug:

```elixir
def matches_condition?(value, [from, to]) when is_integer(from) and is_integer(to) do
  # ...
end

def matches_condition?(value, [from, to]) do
  matches_condition?(value, [String.to_integer(from), String.to_integer(to)])
end
```

matiasgarciaisaia (commented):
👍

assets/js/components/surveys/SurveyWizardCutoffStep.jsx
```diff
@@ -75,8 +75,14 @@ class SurveyWizardCutoffStep extends Component {
...
-        if (typeof value == "object") {
-          value = join(value, " - ")
+        if(Array.isArray(value) && value.length == 2){
+          if(value[0] == null){
+            value = `< ${value[1] + 1}`
+          } else if(value[1] == null){
+            value = `${value[0]}+`
+          } else {
+            value = join(value, " - ")
+          }
```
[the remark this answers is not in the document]
anaPerezGhiglia: #2325

PR #2323, opened 2024-04-17
Make pre-commit git hook opt-in

matiasgarciaisaia:
Instead of having it as a default. The pre-commit script is slow to run, making it harder to commit often.


anaPerezGhiglia (approved):
✨  Lovely 🙌

ysbaddaden:
Sure. I go with `--no-verify` when I need to commit quickly myself.

PR #2324, opened 2024-04-19
Fix randomly failing channel broker test

matiasgarciaisaia:
The test was randomly failing, regardless of the seed that was used.

(( **FIXME**: update this description explaining the solution once we found it))


anaPerezGhiglia:
@matiasgarciaisaia I think the test will be fixed once we merge #2334. Now the test uses [`assert_receive`](https://hexdocs.pm/ex_unit/1.12.3/ExUnit.Assertions.html#assert_receive/3) that 
> "asserts that a message matching pattern **was or is going to be received** within the timeout period, specified in milliseconds."

anaPerezGhiglia:
closing this PR since ce03a67 should have fixed it

PR #2325, opened 2024-04-22
QuotaBucket: Show single value when single-value-range

anaPerezGhiglia:
for #2318

### DIFF
--- a/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
+++ b/assets/js/components/surveys/SurveyWizardCutoffStep.jsx
@@ -80,6 +80,8 @@ class SurveyWizardCutoffStep extends Component {
             value = `< ${value[1] + 1}`
           } else if(value[1] == null){
             value = `${value[0]}+`
+          } else if(value[0] == value[1]) {
+            value = `${value[0]}`
           } else {
             value = join(value, " - ")
           }


matiasgarciaisaia (changes requested):
Looks like that wasn't enough.

See this example for `18,50,50,80` (note the `Age: 50-49` bucket):
Screenshot 2024-04-22 at 3 42 44 PM

Not sure if it's worth implementing this, given it's not as trivial as it looked.

anaPerezGhiglia:
@matiasgarciaisaia the thing is that the definition should have been `18,50,51,80`

The "inverted" bucket is still an issue, but the single-value bucket seems to work ok
image

matiasgarciaisaia:
Inverted would be solved by sorting the input before processing, right?

We should write a ton of test cases for this, though - should be easy.

anaPerezGhiglia:
> Inverted would be solved by sorting the input before processing, right?
 
Not quite. The values are being sorted already. The thing is that the input has two cutoffs that are the same (`50, 50`).
The thing is how ranges are being defined. The upper bound of the ranges is being defined as `next_lower_bound - 1` 
So, the range "between 50 and 50" turns out to be `50 - 49`. I think that the way of solving this is removing duplicated cutoffs before generating the ranges

PR #2326, opened 2024-04-22
Upgrade ecto & elixir versions

anaPerezGhiglia:
## Changes

* Upgrade Ecto to version that includes fix to elixir-ecto/ecto_sql#539 
  * So we don't have to edit the dump file by-hand for the file to make sense. This is _"blocking"_ new migrations (it's actually not a strict blocker, but definitively a big pain)
* Upgrade elixir because Ecto new version requires it

## Considerations 👀 
* When deploying this to stg we should remove the `--log-sql` parameter from `db-migrate` container since it's no longer a valid parameter
* This actually may apply to other instances deployments (production does not have that parameter configured)


ysbaddaden:
No deprecations to fix in Elixir 1.11? Nice.

PR #2334, opened 2024-04-25
#2330: lock respondents in ChannelBroker

### LINKED ISSUE #2330: ChannelBroker: Lock respondents
The `ChannelBroker` should use the `Respondent.with_lock` function when activating a respondent (as it's being done in VerboiceChannel, NuntiumChannel, SurveyBroker - all modules that belong to the "Triggering layer" - [see old diagram](https://github.com/instedd/surveda/blob/main/docs/surveda-layers-and-respondents_lock.md))

We should remove the `preload(:respondent) from ChannelBrokerState and use the respondent given in `Respondent.with_lock`

anaPerezGhiglia:
## Main changes
- Lock respondent when activating next queued contact in `ChannelBroker`
   - This is necessary since some of the actions triggered when activating a respondent may end up modifying it
- Edit tests that were asserting received messages (either by the Channel or the ChannelBroker). Replace `assert_received` with `assert_receive` which _"asserts that a message matching pattern was or **is going to be received** within the timeout period"_  (see [doc](https://hexdocs.pm/ex_unit/1.12.3/ExUnit.Assertions.html#assert_receive/3)) 
   - This was necessary since the ChannelBroker process now may need to wait for the respondent lock to be granted

### Bonus track
 - Update "Surveda Layers" docs to reflect current state

closes #2330


ysbaddaden:
@anaPerezGhiglia Is this meant to ensure that another actor won't modify the respondent while we're trying to contact them? Otherwise only the expired callback is supposed to update the respondent (ask/setup shouldn't).

anaPerezGhiglia:
> Is this meant to ensure that another actor won't modify the respondent while we're trying to contact them? Otherwise only the expired callback is supposed to update the respondent (ask/setup shouldn't).

@ysbaddaden yes, it's true that `nuntium_channel` and `verboice_channel` do not end up modifying the respondent, but actually the channel_broker doesn't know that... that's why I decided to leave the `Respondent.with_lock` for both scenarios (ivr and sms)

ysbaddaden:
I'd consider it surprising for the channels to _modify_ the respondent, I'd expect them to try to push the contact and failures to bubble back to the channel broker.

Just in case: expiring the contact doesn't also try to lock the respondent (or it's reentrant)?

ysbaddaden (approved):

docs/surveda-layers-and-respondents_lock.md
```diff
@@ -49,3 +51,9 @@ Every time a respondent interacts with Surveda
 * `sync_step`
 
 See [this commit](https://github.com/instedd/surveda/pull/1667/commits/d2f3aa59f3fe21b869009c72bb369a4775b7b133) for further details
+
+### ChannelBroker
+
+Every time the ChannelBroker unqueues a respondent and activates their contact
```
matiasgarciaisaia, line 57:
```suggestion
Every time the ChannelBroker dequeues a respondent and activates their contact
```

matiasgarciaisaia (approved):

PR #2335, opened 2024-04-26
#2328: validate sms parts

### LINKED ISSUE #2328: Questionnaires marked as complete with empty SMS parts
As of 0.37.2, a questionnaire with an SMS step consisting of two empty parts is marked as a correct/completed questionnaire - but it should be marked as incomplete if any SMS part is empty.

Both SMS parts of a step are empty, but the questionnaire is validated as correct

anaPerezGhiglia:
## Identifying the bug
 - The questionnaire validator was validating whether the `sms.prompt` was empty
 - As sms parts are being defined by the presence of `smsSplitSeparator` character in the `sms.prompt`, the prompt wasn't empty when there are parts

## Solution
 - When validating the `sms.prompt` consider the presence of `smsSplitSeparator and analyze each part individually
 - For propagating the errors for each part I had to add an index into the error message so the component in charge of rendering them knows which part is having issues

### Evidence

**on staging - without the changes**

https://github.com/instedd/surveda/assets/13237343/624bdb78-31a0-4a2c-8d8d-c055eb875d51

**locally - after the changes**

https://github.com/instedd/surveda/assets/13237343/a9aa75f6-049f-4047-8b57-49f65180bc19

closes #2328

### DIFF
--- a/assets/js/components/questionnaires/SmsPrompt.jsx
+++ b/assets/js/components/questionnaires/SmsPrompt.jsx
@@ -109,7 +109,7 @@ class SmsPrompt extends Component {
         />
       </div>
     )
-    if (index == 0) {
+    if (first) {
       return (
         <div className="row" key={index}>
           <div className="col s12">{inputComponent}</div>
--- a/assets/js/components/questionnaires/SmsPrompt.jsx
+++ b/assets/js/components/questionnaires/SmsPrompt.jsx
@@ -50,24 +50,36 @@ class SmsPrompt extends Component {
     return joinSmsPieces(this.allInputs.map((x) => x.getText() || ""))
   }
 
+  filterErrorsForPart(errors, index) {
+    const partErrorPrefix = `[${index}]`
+    return errors &&
+      errors
+      .flat()
+      // We are interested only in the errors that
+      // match with the sms part index
+      .filter(error => error.startsWith(partErrorPrefix))
+      .map(error => {
+        // We remove the part prefix
+        const errorMessage = error.split(partErrorPrefix)[1]
+        return [errorMessage]
+      })
+  }
+
   renderSmsInput(total, index, value) {
     let { inputErrors, readOnly, fixedEndLength, label, t } = this.props
-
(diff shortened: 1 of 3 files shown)


assets/js/components/questionnaires/SmsPrompt.jsx
```diff
@@ -97,7 +109,7 @@ class SmsPrompt extends Component {
         />
       </div>
     )
-    if (total <= 1 || index == 0) {
+    if (index == 0) {
```
ysbaddaden, line 112: `if (first)` ?

ysbaddaden (approved):
I even comes with tests :heart:

assets/js/reducers/questionnaire.validation.js
```diff
@@ -204,6 +204,12 @@ const validateSmsLangPrompt = (
     addError(context, `${path}.sms`, k("SMS prompt must not be blank"), lang, "sms")
     return
   }
+  const smsParts = prompt.sms && prompt.sms.trim().split(smsSplitSeparator) || []
+  smsParts.forEach((part, idx) => {
+    if(isBlank(part)){
+      addError(context, `${path}.sms`, k(`[${idx}]SMS prompt must not be blank`), lang, "sms")
```
matiasgarciaisaia, line 210: The `k` function is there to tag strings that are translatable. In this case, given that the `[${idx}]` part is a prefix for the code itself, I wouldn't show it to translators (they could choose to move the preffix to be a suffix, or whatever.

If it was a parameter, we should parameterize it like we do below:
https://github.com/instedd/surveda/blob/c046534ac3857275576fee9d031239557f0aadc3/assets/js/reducers/questionnaire.validation.js#L306

But in this case, I think we should leave the index part out of the `k` key:

```javascript
      addError(context, `${path}.sms`, `[${idx}]${k("SMS prompt must not be blank")}`, lang, "sms")
```

It'd be even better if we didn't abuse the notation with a prefix (and, instead, make it an object or an array). But that'd be even more complicated.

PR #2336, opened 2024-04-29
#2308: Fix SuccessRateLine chart

### LINKED ISSUE #2308: Success Rate chart has issues with the x-axis scale
After a two week of running a survey the x-axis of the success rate chart changed its scale, making it really difficutl to read.

image

This might be related to the quotas line-chart and its completion estimation

You can check an example [here](https://surveda-lk.org/projects/42/surveys/484
)

anaPerezGhiglia:
## Changes
* Remove the forecast for success rate values. We were trying to forecast when will the success rate percentage hit the `100%` ceil. This didn't (and doesn't) make any sense. It was this "forecasted" date the one that was messing around the x-axis, since we were selecting this date as the end-date of the chart.
* Simplify `SuccessRateLine` component by reciving a single data object instead of a list of them. This seems to be a leftover from copying Forecasts component
* Simplify `SurveyShow` component by avoid putting together quota bucket forecasts and success rate values
* Change chart title and subtitle since it was the same as the cockpit chart

### Evidence
**Before**
image

**After**

image

closes #2308

### DIFF
--- a/assets/js/components/charts/SuccessRateLine.jsx
+++ b/assets/js/components/charts/SuccessRateLine.jsx
@@ -99,7 +99,7 @@ export default class SuccessRateLine extends Component<Props> {
           .attr("r", "3px")
           .style("fill", data.color)
           .style("stroke", data.color)
-          .style("opacity", 0.2)
+          .style("opacity", 0.1)
           .on("mouseover", (d) => {
             tooltip
               .text(percentFormat(d.value / 100))
--- a/assets/js/components/charts/SuccessRateLine.jsx
+++ b/assets/js/components/charts/SuccessRateLine.jsx
@@ -53,33 +53,28 @@ export default class SuccessRateLine extends Component<Props> {
   renderD3(initial = false) {
     const { width, height, data } = this.state
 
-    let srData = data.filter((d) => d.label === "Success rate")
-    if (!srData[0].values.length) {
+    if (!data){
       return
     }
-    srData[0].values.unshift({ time: srData[0].values[0].time, value: 0 })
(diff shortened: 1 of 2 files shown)


assets/js/components/charts/SuccessRateLine.jsx
```diff
@@ -53,33 +53,28 @@ export default class SuccessRateLine extends Component<Props> {
...
     } else {
       initialTime = d3.timeHour.offset(
-        d3.min(flatten, (d) => d.time),
+        d3.min(data.values, (d) => d.time),
         -1
       )
-      lastTime = d3.timeMonth.offset(initialTime, 1)
-      lastTime = d3.max([d3.max(flatten, (d) => d.time), lastTime])
+      const oneMonthFromStart = d3.timeMonth.offset(initialTime, 1)
+      lastTime = d3.max([d3.max(data.values, (d) => d.time), oneMonthFromStart])
```
matiasgarciaisaia, line 71: We could eventually add a few days of leeway for still-running surveys, to hint that it hasn't finished yet.

matiasgarciaisaia (approved):
👍 LGTM!

assets/js/components/charts/SuccessRateLine.jsx
anaPerezGhiglia: it's common to have surveys that last longer than one month?

matiasgarciaisaia: I think most full-scale surveys are around one month long now, but we used to have larger ones (a bit over two months).

anaPerezGhiglia: follow up on #2338

PR #2340, opened 2024-05-07
Fix 'Stop MO' tests

anaPerezGhiglia:
## Context
Stop MO tests have been failing erratically.
They have been failing even before changing the `ChannelBroker` to use the `Respondent.with_lock` mutex. [This](https://github.com/instedd/surveda/actions/runs/8757591555/job/24036559070) is an evidence of that

## Discovery
The tests were not filtering by `respondent_id`, so they could be using another records than the intended ones for the assertions.
The tests were using plain `Repo.one!` for fetching `Respondent`, `SurveyLogEntry` and `RespondentDispositionHistory` models. The thing is that Mix runs tests of different modules in parallel and other tests insert records in those tables, so there is no guarantee from the `stop mo` tests that they will be testing the same respondent activity throughout the test

## Observations
I could not replicate the failures locally, so I actually have no guarantee that these changes fix them. However I find this a plausible cause of the failures
Leaving here some other examples of the failures in GHA
 - https://github.com/instedd/surveda/actions/runs/8972247623/job/24641885219
 - https://github.com/instedd/surveda/actions/runs/8819089373/job/24209452986

### DIFF
--- a/test/ask/runtime/survey_test.exs
+++ b/test/ask/runtime/survey_test.exs
@@ -3673,13 +3673,13 @@ defmodule Ask.Runtime.SurveyTest do
   end
 
   defp assert_respondent(respondent_id, %{
-    current_state: current_state,
-    previous_disposition: previous_disposition,
-    current_disposition: current_disposition,
-    user_stopped: user_stopped
-    }) do
+         current_state: current_state,
+         previous_disposition: previous_disposition,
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia (approved):
👍 Looks good!

If there are more fixes to do, we'll add them as we found them.

Issue #2329, opened 2024-04-24
Activity log is recording internal Docker IPs


matiasgarciaisaia:
In the country instances we're using, Surveda's Activity Log is recording internal Docker IPs (`10.42.133.167`) instead of the user's actual public IP address.

matiasgarciaisaia:
I've done some quick local tests, and [here](https://github.com/instedd/surveda/blob/012c741dd8ba0022eeab2e4b47a96dba105c3460/lib/ask/activity_log.ex#L113) this `conn.remote_ip` is the reverse proxy's one - but we have the `X-Forwarded-For` header present in the `conn`, so we should use that instead:

<details>
<summary>Local request (using dockerdev)</summary>
<pre>
app-1      | 2024-04-24T19:31:29.898Z request_id=F8lOLVYWVPy2524AAB_D [warn] Here's a conn we received:
app-1      | %Plug.Conn{
app-1      |   adapter: {Plug.Cowboy.Conn, :...},
app-1      |   assigns: %{
app-1      |     current_user: %Ask.User{
app-1      |       __meta__: #Ecto.Schema.Metadata<:loaded, "users">,
app-1      |       channels: #Ecto.Association.NotLoaded<association :channels is not loaded>,
app-1      |       confirmation_sent_at: nil,
app-1      |       confirmation_token: nil,
app-1      |       confirmed_at: nil,
app-1      |       current_password: nil,
app-1      |       email: "mgarcia@manas.com.ar",
app-1      |       id: 1,
app-1      |       inserted_at: ~U[2024-04-24 19:31:15Z],
app-1      |       name: "",
app-1      |       oauth_tokens: #Ecto.Association.NotLoaded<association :oauth_tokens is not loaded>,
app-1      |       password: nil,
app-1      |       password_confirmation: nil,
app-1      |       password_hash: "$2b$12$F/aI3s4KuhUg.IVHZ5WR1ehUK.mAzO85dFEwwKqBssN/DKN20XJC2",
app-1      |       project_memberships: #Ecto.Association.NotLoaded<association :project_memberships is not loaded>,
app-1      |       projects: #Ecto.Association.NotLoaded<association :projects is not loaded>,
app-1      |       remember_created_at: nil,
app-1      |       reset_password_sent_at: nil,
app-1      |       reset_password_token: nil,
app-1      |       settings: %{},
app-1      |       updated_at: ~U[2024-04-24 19:31:15Z]
app-1      |     }
app-1      |   },
[...]

PR #2342, opened 2024-05-11
Use proxy-forwarded IP address in ActivityLog

matiasgarciaisaia:
There's no fail-safe way of detecting it[0], but we were logging Rancher internal IPs so far - so it won't be worse than that.

See #2329

[0]: https://adam-p.ca/blog/2022/03/x-forwarded-for/

### DIFF
--- a/lib/ask/activity_log.ex
+++ b/lib/ask/activity_log.ex
@@ -103,24 +103,31 @@ defmodule Ask.ActivityLog do
   defp typeof(%Questionnaire{}), do: "questionnaire"
   defp typeof(%Folder{}), do: "folder"
 
+  defp remote_ip(nil), do: "0.0.0.0"
+
+  defp remote_ip(%{remote_ip: remote_ip, req_headers: headers}) do
+    case Enum.find(headers, fn {key, _value} -> key == "x-forwarded-for" end) do
+      {"x-forwarded-for", forwarded_ip} -> forwarded_ip
+      nil -> to_string(:inet_parse.ntoa(remote_ip))
+    end
+  end
+
   defp create(action, %Project{id: project_id}, conn, entity, metadata) do
     create(action, project_id, conn, entity, metadata)
   end
 
   defp create(action, project_id, conn, entity, metadata) do
-    {user_id, remote_ip} =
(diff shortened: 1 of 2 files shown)


lib/ask/activity_log.ex
```diff
@@ -103,21 +103,28 @@ defmodule Ask.ActivityLog do
...
-            user -> {user.id, remote_ip}
-          end
+    user_id =
+      if conn do
+        case current_user(conn) do
+          nil -> nil
+          user -> user.id
+        end
+      else
+        nil
```
anaPerezGhiglia, line 123: wouldn't it make more sense to add a guard for the `current_user` that handles the a `nil` parameter?
```elixir
  def current_user(nil) do
    nil
  end
```

and here just do
```elixir
user_id = case current_user(conn) do
           nil -> nil
           user -> user.id
          end
```

lib/ask/activity_log.ex
```diff
@@ -103,21 +103,28 @@ defmodule Ask.ActivityLog do
   defp typeof(%Questionnaire{}), do: "questionnaire"
   defp typeof(%Folder{}), do: "folder"
 
+  defp remote_ip(nil), do: "0.0.0.0"
+
+  defp remote_ip(%{remote_ip: remote_ip, req_headers: headers}) do
+    case Enum.find(headers, fn {key, _value} -> key == "x-forwarded-for" end) do
```
anaPerezGhiglia, line 109: shouldn't we make this condition case insensitive?

anaPerezGhiglia (approved):
LGTM

lib/ask/activity_log.ex
matiasgarciaisaia: Plug states that ["request headers are normalized to lowercase"](https://hexdocs.pm/plug/1.12.1/Plug.Conn.html)

lib/ask/activity_log.ex
matiasgarciaisaia: I'm not 100% convinced that we should _not_ crash if we get a `nil` `conn`, but I've implemented the change anyways 👍

PR #2346, opened 2024-05-31
Improve Interactions file generation startup time

matiasgarciaisaia:
Index `survey_log_entries` (which can grow really large) by `channel_id` to speed up a query checking for DISTINCT channel_id's when starting to generate the interactions file.

In a 10M-row table, the query goes down from 25s to less than 1s with the index.


matiasgarciaisaia:
I still have to measure if/how this impacts performance while running the surveys themselves.

I wouldn't expect a lot of overhead, since there's usually just a single channel (maybe two, if doing multi-mode), but I haven't tested it yet.

PR #2347, opened 2024-06-07
Fix SuccessRateLine chart

anaPerezGhiglia:
## Problem
The colored area of the chart was weird
 
Screenshot 2024-06-07 at 11 40 02

## Changes
- use all values (as a leftover of old component we were discarting the first and last element)
- build the area comound by the circle points "by-hand" (instead of just using the "fill" between the line and the points)

### Evidence 
Screenshot 2024-06-07 at 12 47 49

related to #2308


matiasgarciaisaia (approved):
LGTM! 👍

PR #2348, opened 2024-06-10
[workaround] Skip canceller tests

anaPerezGhiglia:
However, there seems to be something odd with the canceller since the tests print these non-test-breaking errors

It seems that we are missing a `handle_cast` implementation on it.
Were these errors always there and we never noticed them? 🤔 

```
2024-06-10T18:10:43.356Z [error] GenServer #PID<0.554.0> terminating
** (FunctionClauseError) no function clause matching in Ecto.Changeset.cast/4
    (ecto 3.9.4) lib/ecto/changeset.ex:497: Ecto.Changeset.cast(nil, %{state: "terminated"}, [:name, :description, :project_id, :folder_id, :mode, :state, :locked, :exit_code, :exit_message, :cutoff, :schedule, :sms_retry_configuration, :ivr_retry_configuration, :mobileweb_retry_configuration, :fallback_delay, :started_at, :quotas, :quota_vars, :comparisons, :count_partial_results, :simulation, :ended_at, :incentives_enabled, :first_window_started_at, :last_window_ends_at, :panel_survey_id, :generates_panel_survey], [])
    (ask 0.31.0) lib/ask/survey.ex:139: Ask.Survey.changeset/2
    (ask 0.31.0) lib/ask/runtime/survey_canceller.ex:44: Ask.Runtime.SurveyCanceller.terminate_survey/1
    (ask 0.31.0) lib/ask/runtime/survey_canceller.ex:83: Ask.Runtime.SurveyCanceller.handle_info/2
    (stdlib 3.12.1.2) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib 3.12.1.2) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib 3.12.1.2) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: :cancel
```

### DIFF
--- a/lib/ask.ex
+++ b/lib/ask.ex
@@ -20,7 +20,6 @@ defmodule Ask do
       supervisor(Registry, [:unique, :channel_broker_registry]),
       worker(Ask.Runtime.ChannelBrokerAgent, []),
       supervisor(Ask.Runtime.ChannelBrokerSupervisor, []),
-      supervisor(Ask.Runtime.SurveyCancellerSupervisor, []),
       {Mutex, name: Ask.Mutex}
       # Start your own worker by calling: Ask.Worker.start_link(arg1, arg2, arg3)
       # worker(Ask.Worker, [arg1, arg2, arg3]),
@@ -45,7 +44,10 @@ defmodule Ask do
             worker(Ask.Config, []),
             worker(Ask.Runtime.QuestionnaireSimulatorStore, [])
             | children
-          ]
+          ] ++ [
+              # SurveyCancellerSupervisor depends on Ask.Repo, so must be started (and declared!) after it
+              supervisor(Ask.Runtime.SurveyCancellerSupervisor, [])
+            ]
 
         true ->
           children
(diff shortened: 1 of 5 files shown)


lib/ask.ex
```diff
@@ -45,7 +44,10 @@ defmodule Ask do
             worker(Ask.Config, []),
             worker(Ask.Runtime.QuestionnaireSimulatorStore, [])
             | children
-          ]
+          ] ++ [
+              # SurveyCancellerSupervisor depends on Ask.Repo, so must be started (and declared!) after it
+              supervisor(Ask.Runtime.SurveyCancellerSupervisor, [])
+            ]
```
matiasgarciaisaia, line 50: This looks really ugly, but the ordering is relevant.

matiasgarciaisaia (approved):
I think I've fixed the tests instead of skipping them.

I also tested running the app and cancelling surveys (both while the app is running, and by stopping the server with a cancelling survey and starting it again), and it worked.

PR #2354, opened 2024-07-16
Duplicate survey

### LINKED ISSUE #2351: Survey settings duplication
There are several scenario in which the users create similar surveys.
When creating a survey for each MNO.
When canceling a survey for an issue and then creating it again to resume data collection
Etc.
It will be useful to have a way to duplicate the survey settings.
The duplicated aspects will be: 
	- Questionnaire (the current version not the saved one)
	- Modes
	- Schedule (including start, end and block dates, timezone, contact window)
	- Retry delays per mode and fallback delay
	- Quotas definition or completes
	- Count partial as completes

Sample, channel and quota values will not be copied.

matiasgarciaisaia:
Allows users to duplicate non-panel surveys.

Survey card showing the button to duplicate it

Duplicated surveys don't copy the source's sample nor bucket's quotas. Surveys can be copied before, during, and after they run.

Fixes #2351

### DIFF
--- a/assets/js/components/surveys/SurveyCard.jsx
+++ b/assets/js/components/surveys/SurveyCard.jsx
@@ -23,7 +23,7 @@ class _SurveyCard extends Component<any> {
     readOnly: boolean,
   }
 
-  duplicatingSurvey: boolean // FIXME: this should be declared at the Index level (shared between cards)
+  duplicatingSurvey: boolean
 
   constructor(props) {
     super(props)
@@ -52,7 +52,6 @@ class _SurveyCard extends Component<any> {
     const { dispatch, survey } = this.props
     dispatch(surveyActions.duplicateSurvey(survey)).then((newSurvey) => {
       this.duplicatingSurvey = false
-      // FIXME: navigate using the router? it's not available here
       window.location = routes.survey(newSurvey.projectId, newSurvey.id)
     })
   }
--- a/assets/js/components/surveys/SurveyCard.jsx
+++ b/assets/js/components/surveys/SurveyCard.jsx
@@ -23,6 +23,8 @@ class _SurveyCard extends Component<any> {
     readOnly: boolean,
   }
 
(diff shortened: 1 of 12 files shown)


test/ask_web/controllers/survey_controller_test.exs
```diff
@@ -1271,6 +1271,155 @@ defmodule AskWeb.SurveyControllerTest do
...
+      project = create_project_for_user(user)
+      panel_survey = insert(:panel_survey, project: project)
+      survey = insert(:survey, project: project, panel_survey: panel_survey)
+
+      assert_error_sent :conflict, fn ->
+        post conn, project_survey_survey_path(conn, :duplicate, project, survey)
+      end
+    end
+
+    test "running survey copies use editable questionnaire", %{ conn: conn, user: user } do
```
anaPerezGhiglia, line 1388: it would be nice to add a test to assert that the duplicated survey of a running survey uses the **original** questionnaire instead of the original survey id

assets/js/components/surveys/SurveyCard.jsx
```diff
@@ -36,10 +38,25 @@ class _SurveyCard extends Component<any> {
...
+
+    this.duplicatingSurvey = false
   }
 
   changeFolder = (folderId) => this.setState({ folderId })
 
+  duplicateSurvey = () => {
+    // Prevent duplicating multiple surveys at once
+    if (this.duplicatingSurvey) return
+    this.duplicatingSurvey = true
```
anaPerezGhiglia, line 50: what's the scenario for this?

anaPerezGhiglia (approved):
Looking good!

assets/js/components/surveys/SurveyCard.jsx
matiasgarciaisaia: This is mostly to block the `Duplicate` button while we're duplicating a survey - if the server gets slow and takes a while to reply.

anaPerezGhiglia (approved):
✨  :shipit: ✨

PR #2355, opened 2024-07-17
#2343: Log in interaction files on ChannelBroker

### LINKED ISSUE #2343: Enhance interactions file logging (contact enqueueing)
The current enqueuing call no longer matches when surveda enqueues the call on verboice since we have the channel broker in the middle.
A quick&easy solution would be to remove the log_contact("Enqueueing call") from runtime/session and add it to the ChannelBroker when it really setups the call in the verboice channel

anaPerezGhiglia:
## Changes
Move the responsibility of logging into the interaction fril from the `runtime/session` to `runtime/channel_broker` since is the "last" component of the chain, is who's responsible of actually enqueuing the contacts in the runtime channels

## Collateral changes
 - with these new changes, we are logging into `SurveyLog` the "enqueuing" message both for IVR, SMS and WebMobile. Before it was only recorded for IVR
 - Included the usage of `SurveyLogEntry.changeset()` to validate and normalize the entries
 - As logging it is now the `ChannelBroker` responsibility, which is a different process than the `SurveyBroker`, the entries generated by the SurveyBroker "compete" with the ones generate by the `ChannelBroker`. So, it could happen that two entries get "swapped". 
   - I had to change how we were testing these things to avoid having erratic tests about it
   - 
## Considerations 👀 
~Mobile web surveys contacts will be registered as `sms` mode since the channel being used is an sms one
@manumoreira is this ok? or do we want to have `mobile web` mode listed there?~
Was able to get the real mode by looking into the `session.flow` instead on the channel

closes #2343

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -439,7 +439,7 @@ defmodule Ask.Runtime.ChannelBroker do
     DateTime.compare(not_after, Ask.SystemTime.time().now) != :gt
   end
 
-  defp log_contact(status, state, respondent) do #channel, mode, respondent, disposition \\ nil) do
+  defp log_contact(status, respondent) do
     session = respondent.session |> Ask.Runtime.Session.load()
     SurveyLogger.log(
       respondent.survey_id,
@@ -454,7 +454,7 @@ defmodule Ask.Runtime.ChannelBroker do
   end
 
   defp ivr_call(state, respondent, token, not_before, not_after) do
-    log_contact("Enqueueing call", state, respondent)
+    log_contact("Enqueueing call", respondent)
 
     response =
       state.runtime_channel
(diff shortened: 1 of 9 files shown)


lib/ask/runtime/channel_broker.ex
```diff
@@ -439,7 +439,23 @@ defmodule Ask.Runtime.ChannelBroker do
     DateTime.compare(not_after, Ask.SystemTime.time().now) != :gt
   end
 
+  defp log_contact(status, respondent) do
+    session = respondent.session |> Ask.Runtime.Session.load()
+    SurveyLogger.log(
+      respondent.survey_id,
+      session.current_mode.channel.type,
```
matiasgarciaisaia, line 446: I'm surprised we still don't have a function that receives a mode and returns its label (`SMS`/`IVR`/`Mobile Web`). We should use that here to avoid breaking the current interaction file format.

matiasgarciaisaia (changes requested):
Let's fix the mode we log to avoid breaking the interaction files' format.

Other than that, the PR's OK 👍

anaPerezGhiglia:
@matiasgarciaisaia this should be ready 🙂

matiasgarciaisaia (approved):
Great! :shipit:

PR #2358, opened 2024-07-30
Fix & improve survey logger tests

anaPerezGhiglia:
* remove inconsistent checks of SurveyLoggerEntries in tests that want to assert respondent
* add SurveyLoggerEntry normalize mode tests
* add SurveyLogger tests

### DIFF
--- a/test/ask/survey_log_entry_test.exs
+++ b/test/ask/survey_log_entry_test.exs
@@ -19,4 +19,27 @@ defmodule Ask.SurveyLogEntryTest do
     changeset = SurveyLogEntry.changeset(%SurveyLogEntry{}, @valid_attrs)
     assert changeset.valid?
   end
+
+  test "changeset normalizes sms mode" do
+    changes = %{@valid_attrs | mode: "sms"}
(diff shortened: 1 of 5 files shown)


test/ask/survey_log_entry_test.exs
```diff
@@ -19,4 +19,27 @@ defmodule Ask.SurveyLogEntryTest do
...
+
+  test "changeset normalizes mobileweb mode" do
+    changes = %{@valid_attrs | mode: "mobileWEB"}
+    changeset = SurveyLogEntry.changeset(%SurveyLogEntry{}, changes)
+    assert changeset.changes.mode == "Mobile Web"
+  end
+
+  test "changeset leaves unrecognized mode as it is" do
+    changeset = SurveyLogEntry.changeset(%SurveyLogEntry{}, @valid_attrs)
+    assert changeset.changes.mode == "some content"
```
matiasgarciaisaia, line 43: It's a bit weird to have `"some content"` as the mode for a `@valid_attrs` given that it isn't _that_ valid, I guess.

But not something to fix now.

matiasgarciaisaia (approved):
Great one!

The way in which this tests the processes finally starts to make sense.

Issue #2350, opened 2024-07-02
CSV files download should be async


manumoreira:
In the last years CSV download has been generating several problems to the users. On medium to big surveys it ends up crashing the server frequently. 
A solution for this problem can be uncoupling file creation from download.
This will imply a change in the user interaction. 
Initially the user will click to create the file, the creation will be processed and once it is ready the user will be see a button to download it.
We will need mockups for this

![Image](https://github.com/user-attachments/assets/9092dcc0-3d8d-45c0-a288-9c9a7755bc1e)

([source](https://manas.slack.com/archives/C1L7CJMNU/p1722964757776989?thread_ts=1722546440.413409&cid=C1L7CJMNU))

Questions:
* Will it be usefull to have an option to trigger the file creation for all the files of the survey?

matiasgarciaisaia:
Let's note that the file which usually gives issues is the Interactions file (I don't remember having had issues with any other CSV file - but please correct me, @manumoreira).

But it may make a lot of sense to share the same approach between different files.

We could also generate the task (to create the files) and then notifiy the requester via email.

manumoreira:
In some cases we've seen issues with the results file in large surveys. 
I'd prefer not to add an email service for this, just to keep it simple. 
A circular progress bar might be enough.

manumoreira:
After a few tries I realized that the interaction in this modal can be enhanced ![Image](https://github.com/user-attachments/assets/ee1a646a-424b-4e14-8a50-d44c260b0bb7)
When clicking on the refresh icon it takes ~30'' to provide some feedback to the user, sometimes it shows a legend of "Generating file" sometimes (when the file is smaller) it just starts counting the time sindce the last generated file.
providing some immediate feedback seems important, otherwise we will have the user clicknig the icon too many times.

PR #2360, opened 2024-08-01
Increase DB chunk size for respondent files

matiasgarciaisaia:
Respondent files are usually large (Interactions files can grow up to 1M rows), and the "low" limit in queries made the DB work much more than needed (we've observed 99% CPU usage in the mysqld process when generating a 1M-rows interactions file with 1000 rows per query).

Increasing this limit makes the app generate less queries to the DB, effectively driving the CPU usage down to about 30% instead.

There's probably more room for improvement (the generation of the file is still CPU-bound instead of network-bound), but that's on the app itself - we should profile the app's code to further improve the performance.

See #2350
See #2359

### DIFF
--- a/lib/ask_web/controllers/respondent_controller.ex
+++ b/lib/ask_web/controllers/respondent_controller.ex
@@ -15,6 +15,8 @@ defmodule AskWeb.RespondentController do
     RespondentsFilter
   }
 
+  @db_chunk_limit 10_000
+
   def index(conn, %{"project_id" => project_id, "survey_id" => survey_id} = params) do
     limit = Map.get(params, "limit", "")
     page = Map.get(params, "page", "")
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
We suspect the CSV encoder may be suboptimal.

We've also tested with 100k and 50k limits for the interactions files, but the performance was a slightly worse (startup took a bit longer, and the overall speed didn't improve further).

matiasgarciaisaia:
CC: @ggiraldez in case you want to add anything else.

ggiraldez:
> CC: @ggiraldez in case you want to add anything else.

I know very little of Elixir or Ecto, but you may also want to explore streaming directly from the database. The MySQL driver supports it, although it requires a transaction which may be a deal breaker perhaps? Anyway, see https://hexdocs.pm/ecto/Ecto.Repo.html#c:stream/2

matiasgarciaisaia:
I tried changing the queries to use Ecto's `Repo.stream` (instead of manually doing `Stream.resource` and paginating from the app) but `preloads are not supported on streams` 🫠 

There may still be room for doing the CSV streaming straight from the database (ie, make MySQL output CSV) as @ggiraldez suggested me, but I'm not sure if that'll work or not - I'll leave this as is, we might explore that optimization if we eventually need it. Given we're about to make the file generation async (in #2350) improving the times won't be _that_ important, either.

PR #2362, opened 2024-08-05
#2350: Generate survey files async

anaPerezGhiglia:
## Changes
- create new `SurveyResults` GenServer
- move responsibility of generating files from `RespondentController` to new `SurveyResults` actor 
- change files endpoints to "common" POST (without `csv` format)
- change UI to make the POST call instead of a redirect to a new location

## Out of scope
- Keep track of generated files in the db
- Files cleanup policy
- Download existing files

for #2350 

Screenshot 2024-11-28 at 5 42 54 PM


matiasgarciaisaia:
First of all, sorry for the _huge_ PR. Things got wild.

Here's my list of pending things, which we may want to move to a different issue/PR.

- [ ] Check Survey state against pre-generated files (eg, don't offer to download a file from when the survey was running if it's already finished, maybe prevent regenerating files from finished surveys)
- [ ] Only return status of files that the user can access (we're leaking some harmless info there - files can't be downloaded by unauthorized users)
- [ ] Avoid concurrent `fetchRespondentsFilesStatus` calls
- [ ] Fix skipped tests

I'll try to address some of these while we review the PR.

matiasgarciaisaia:
- [ ] Surveda won't manage old survey files (they'll pile up in the FS)
- [ ] We should probably Docker-mount another volume/FS in the directory where the files are going to be generated in
- [ ] Filtered results should honour the current filter (we're serving _whichever_ filter we last generated). For this to work, filtered results should include filter in filename (beware special characters like `:` and ` `).

lib/ask/survey_results.ex
```diff
@@ -353,56 +349,102 @@ defmodule Ask.SurveyResults do
...
 
-  defp do_generate_file(file_type, _, _),
-    do: Logger.warn("No function for generating #{file_type} files")
-
-  def file_path(survey, file_type) do
-    filename = csv_filename(survey, file_prefix(file_type))
-    "#{@target_dir}/#{filename}"
+  def file_path(survey, file_type, target_dir \\ @target_dir) do
+    # FIXME: as a first iteration, generate a stable name and have a single file per type
+    # but we should probably include the date and respondent results filter in the name
```
anaPerezGhiglia, line 357: stale FIXME 🙂

lib/ask/survey_results.ex
```diff
@@ -353,56 +349,102 @@ defmodule Ask.SurveyResults do
...
+    prefix = file_prefix(file_type)
```
anaPerezGhiglia, line 358: 💅 nitpick: it's not a "prefix" but rather a suffix

lib/ask/survey_results.ex
```diff
@@ -353,56 +349,102 @@ defmodule Ask.SurveyResults do
...
 
   defp write_to_file(file_type, survey, rows) do
     File.mkdir_p!(@target_dir)
-    file = File.open!(file_path(survey, file_type), [:write, :utf8])
+    target_file = file_path(survey, file_type, @target_dir)
+    if File.exists?(target_file), do: File.rm!(target_file)
+
+    # Poor man's mktemp. We only want to avoid having the file living at the stable
+    # path while it's still being written to avoid partial downloads
+    temporal_file = for _ <- 1..10, into: "#{target_file}.tmp.", do: <<Enum.random(?a..?z)>>
```
anaPerezGhiglia, line 373: 💅 
~temporal_file~ -> temporal_file_name

lib/ask/survey_results.ex
```diff
@@ -353,56 +349,102 @@ defmodule Ask.SurveyResults do
...
+    # name = Regex.replace(~r/[^a-zA-Z0-9_]/, name, "_")
+    # current_time = Timex.format!(DateTime.utc_now(), "%Y-%m-%d-%H-%M-%S", :strftime)
+    # "#{target_dir}/#{name}_#{survey.state}-#{prefix}_#{current_time}.csv"
+    "#{target_dir}/survey_#{survey.id}-#{survey.state}-#{prefix}.csv"
   end
 
   defp write_to_file(file_type, survey, rows) do
     File.mkdir_p!(@target_dir)
-    file = File.open!(file_path(survey, file_type), [:write, :utf8])
+    target_file = file_path(survey, file_type, @target_dir)
```
anaPerezGhiglia, line 368: 💅 
~target_file~ -> target_file_name

lib/ask/survey_results.ex
```diff
@@ -353,56 +349,102 @@ defmodule Ask.SurveyResults do
...
-        file |> String.starts_with?(survey_filename_prefix(survey, file_prefix(:interactions)))
-      end)
-
-    !exists_file
+  defp file_prefix(:respondents_results), do: "respondents"
+  defp file_prefix({:respondents_results, filter}) do
+    if RespondentsFilter.empty?(filter) do
+      "respondents"
+    else
+      "respondents_filtered_#{filter_hash(filter)}"
```
anaPerezGhiglia, line 444: Let's add a comment here explaining why we are doing this

anaPerezGhiglia (commented):
I cannot officially approve this PR since I was the original author, but this looks good to me! ✨ 
@matiasgarciaisaia I would encourage we deploy a preview of this branch to staging so we can play with it a bit without having to set it up in each of our local environments 😬

lib/ask_web/controllers/respondent_controller.ex
```diff
@@ -740,11 +786,11 @@ defmodule AskWeb.RespondentController do
     # ?param1=value is more specific than ?q=param1:value
     filter = add_params_to_filter(filter, params)
 
-    SurveyResults.generate_respondent_result_file(survey_id, filter)
+    SurveyResults.generate_respondents_results_file(survey_id, filter)
 
-    ActivityLog.download(project, conn, survey, "survey_results") |> Repo.insert()
+    ActivityLog.generate_file(project, conn, survey, "survey_results") |> Repo.insert()
```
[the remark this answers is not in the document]
matiasgarciaisaia: I wanted to keep the files themselves backwards-compatible. Changing these names would imply running a migration to change old activity logs from the DB. Which could be OK, but these are not exactly the same strings used "internally". We should also take a look at including the filter, somehow, for the results file - here we're completely ignoring whether there's a filter or not.

anaPerezGhiglia (commented):
With the last changes the files modal behaves as expected - even changing between different surveys. ✅ 
However, what I've noticed is that once the user opens the "download files" modal, the app keeps on requesting the `/files` status - even if the user has closed the modal. This is not critical, but it's useless. Even more because if the user opens the modal again, a new request is triggered

matiasgarciaisaia:
Yes. We don't have (or I haven't found) an event for when the modal closes, so the requests keep going every 20 seconds even when we don't need them.

PR #2363, opened 2024-08-27
Fix invalid link

### LINKED ISSUE #2345: Invalid public links should return a 404 error
Surveda's public links are yielding `500 Internal Server Error` responses for inexsitent links:

```
$ curl -I https://surveda.instedd.org/link/this-does-not-exist
HTTP/1.1 500 Internal Server Error
cache-control: max-age=0, private, must-revalidate
content-length: 42
content-type: text/html; charset=utf-8
date: Thu, 30 May 2024 21:40:23 GMT
server: Cowboy
strict-transport-security: max-age=31536000
x-request-id: F9RiGqQ7LAwDyGYACMJB
```

The response's body says `There was an error. Please try again later`.

It should actually return a 404 error saying the link does not exist if it's not found in the DB.

This could help avoid automations keep retrying to fetch the same links over and over again.

ismaelbej:
## Changes

* Fix invalid short links to return 404

Closes #2345

### DIFF
--- a/lib/ask_web/controllers/short_link_controller.ex
+++ b/lib/ask_web/controllers/short_link_controller.ex
@@ -9,39 +9,43 @@ defmodule AskWeb.ShortLinkController do
       ShortLink
       |> Repo.get_by(hash: hash)
 
-    if link do
-      conn =
-        conn
-        |> assign(:skip_auth, true)
-
-      {path, query_string} =
-        case :binary.split(link.target, "?", [:global]) do
-          [path, query_string] -> {path, query_string}
-          [path] -> {path, ""}
-        end
-
-      conn = %{
-        conn
-        | request_path: path,
-          path_info: split_path(path),
-          params: %{},
-          path_params: %{},
-          private: %{},
-          req_headers: [],
-          query_string: query_string,
-          query_params: %Plug.Conn.Unfetched{aspect: :query_params}
-      }
-
-      AskWeb.Endpoint.call(conn, [])
-    else
-      conn
-      |> send_resp(:not_found, "Link not found")
-    end
+    resolve_link(conn, link)
   end
 
   # Copied from Plug.Adapters.Cowboy.Conn
   defp split_path(path) do
     segments = :binary.split(path, "/", [:global])
     for segment <- segments, segment != "", do: segment
+  end
+
+  defp resolve_link(conn, nil) do
+    conn
+    |> send_resp(:not_found, "Link not found")
+  end
+
+  defp resolve_link(conn, link) do
+    conn =
+      conn
+      |> assign(:skip_auth, true)
+
+    {path, query_string} =
+      case :binary.split(link.target, "?", [:global]) do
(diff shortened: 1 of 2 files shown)


lib/ask_web/controllers/short_link_controller.ex
```diff
@@ -9,29 +9,34 @@ defmodule AskWeb.ShortLinkController do
       ShortLink
       |> Repo.get_by(hash: hash)
 
-    conn =
-      conn
-      |> assign(:skip_auth, true)
+    if link do
```
matiasgarciaisaia, line 12: Instead of adding an `if`, the Functional Way™ is to delegate into a function with multiple definitions.

So you'd do something like `resolve_link(conn, link)`, and have one definition `defp resolve_link(conn, nil)` that returns the 404, and another one `defp resolve_link(conn, link)` that does the happy path of actually having the link in the DB.

lib/ask_web/controllers/short_link_controller.ex
```diff
@@ -9,29 +9,34 @@ defmodule AskWeb.ShortLinkController do
...
+          req_headers: [],
+          query_string: query_string,
+          query_params: %Plug.Conn.Unfetched{aspect: :query_params}
+      }
 
-    AskWeb.Endpoint.call(conn, [])
+      AskWeb.Endpoint.call(conn, [])
+    else
+      conn
+      |> send_resp(:not_found, "")
```
matiasgarciaisaia, line 38:
```suggestion
      |> send_resp(:not_found, "Link not found")
```

So we show _some_ error to the user, other than the HTTP status code.

test/ask_web/controllers/short_link_controller_test.exs
```diff
@@ -14,6 +14,14 @@ defmodule AskWeb.ShortLinkControllerTest do
     {:ok, conn: conn, user: user}
   end
 
+  test "return 404 for an invalid link", %{
+    conn: conn
+  } do
+    conn = get(conn, short_link_path(conn, :access, "invalid-link"))
+
+    assert html_response(conn, 404)
```
matiasgarciaisaia, line 22:
```suggestion
    assert response(conn, 404) == 'Link not found'
```

lib/ask_web/controllers/short_link_controller.ex
```diff
@@ -9,6 +9,21 @@ defmodule AskWeb.ShortLinkController do
...
       |> Repo.get_by(hash: hash)
 
+    resolve_link(conn, link)
+  end
+
+  # Copied from Plug.Adapters.Cowboy.Conn
+  defp split_path(path) do
+    segments = :binary.split(path, "/", [:global])
+    for segment <- segments, segment != "", do: segment
+  end
```
matiasgarciaisaia, line 19: I wouldn't even have moved this function, so the diff was even more clear :)

matiasgarciaisaia (approved):
👍 Great one!

Welcome aboard :)

:shipit:

PR #2366, opened 2024-09-03
Validate collaborators invited twice

### LINKED ISSUE #1786: Collaborators can be invited twice to the same project
If an Admin or Owner user invites a collaborator twice to the same project no error is shown, but then the project crashes and the UI doesn't work properly.

`{{%Ecto.MultipleResultsError{message: "expected at most one result but got 2 in query:\n\nfrom p in Ask.ProjectMembership,\n  where: p.project_id == ^1,\n  where: p.user_id == ^5\n"},`

ismaelbej:
## Changes

* Back:
  * Return error message when trying to add an existing collaborator to a project
* Front:
  * Remove unused function `loadCollaboratorToEdit` from CollaboratorIndex and properties `confirmationText`, `onConfirm` from InviteModel
  * Modify `changeEmail` action and reducer to accept a list of collaborators to validate the email isn't included between the collaborators
  * Show error message resulting from the email validation

Closed #1786

### DIFF
--- a/assets/js/components/collaborators/InviteModal.jsx
+++ b/assets/js/components/collaborators/InviteModal.jsx
@@ -97,7 +97,7 @@ export class InviteModal extends Component {
       </a>
     )
 
-    const emailError = this.getEmailError(guest);
+    const emailError = this.getEmailError(guest)
 
     const sendButton =
       guest.data.code && !emailError ? (
@@ -191,19 +191,19 @@ export class InviteModal extends Component {
   }
 
   getEmailError(guest) {
-    let errorCode = "invalid-email";
+    let errorCode = "invalid-email"
     if (guest.data.email) {
       errorCode = guest.errors.email
     }
 
     switch (errorCode) {
       case null:
-        return null;
+        return null
       case "existing-email":
-        return "Email already in project";
+        return "User is already a member"
       case "invalid-email":
       default:
-        return "Please enter a valid email";
+        return "Please enter a valid email"
     }
   }
 }
@@ -223,7 +223,6 @@ InviteModal.propTypes = {
   projectId: PropTypes.number,
   style: PropTypes.object,
   userLevel: PropTypes.string,
-  collaborators: PropTypes.array,
 }
 
 const mapDispatchToProps = (dispatch) => ({
--- a/assets/js/components/collaborators/InviteModal.jsx
+++ b/assets/js/components/collaborators/InviteModal.jsx
@@ -32,7 +32,7 @@ export class InviteModal extends Component {
   }
 
   emailOnChange(e) {
-    this.props.guestActions.changeEmail(e.target.value, this.props.collaborators)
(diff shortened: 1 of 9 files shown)


test/ask_web/controllers/invite_controller_test.exs
```diff
@@ -1069,4 +1069,61 @@ defmodule AskWeb.InviteControllerTest do
...
+        "email" => user2.email,
+        "project_id" => project.id
+      })
+    )
+
+    assert json_response(conn, 422)["errors"]["user_id"] == ["User already in project"]
+
+    email = wait_for_email()
+
+    assert email.subject == "#{user.name} has added you as a collaborator on #{project.name}."
```
matiasgarciaisaia, line 1121: If the user was already invited, I guess we shouldn't be sending the email - right?

assets/js/reducers/guest.js
```diff
@@ -34,15 +34,25 @@ const validEmail = (email) => {
 }
 
 const changeEmail = (state, action) => {
+  const { email, collaborators } = action;
```
matiasgarciaisaia, line 37: Aren't the `collaborators` part of the `state` already? Since it's not data generated by the user's action, I don't think we should be sending it in the action itself - but rather get it from the `state`.

assets/js/components/collaborators/InviteModal.jsx
```diff
@@ -189,6 +189,23 @@ export class InviteModal extends Component {
...
+    let errorCode = "invalid-email";
+    if (guest.data.email) {
+      errorCode = guest.errors.email
+    }
+
+    switch (errorCode) {
+      case null:
+        return null;
+      case "existing-email":
+        return "Email already in project";
```
matiasgarciaisaia, line 203:
```suggestion
        return "User is already a member"
```

test/ask_web/controllers/invite_controller_test.exs
```diff
@@ -1069,4 +1069,118 @@ defmodule AskWeb.InviteControllerTest do
...
+        "project_id" => project.id
+      })
+    )
+
+    assert json_response(conn, 422)["errors"]["user_id"] == ["User already in project"]
+
+    refute_received [:email, _email]
+  end
+
+  defp wait_for_email do
```
matiasgarciaisaia, line 1181: We could drop this function and always use `assert_received` for consistency.

matiasgarciaisaia (commented):
Looks great! 👍 

[The CI is complaining about types not being declared](https://github.com/instedd/surveda/actions/runs/10743351869/job/29797916767#step:4:223). I guess that's the only fix left before merging?

ismaelbej:
@matiasgarciaisaia It should be ready for another review. Friday commit was missing using the collaborators from the redux storage, instead of using properties. If found that other actions use `getState()` to read the redux storage.

matiasgarciaisaia (approved):
Great! :shipit:

PR #2370, opened 2024-10-04
Pause channel

ismaelbej:
* Add :paused field to channel model
* Add endpoints to pause and unpause a channel
* Update ChannelStatusServer to skip paused channels


lib/ask_web/controllers/channel_controller.ex
```diff
@@ -106,4 +106,20 @@ defmodule AskWeb.ChannelController do
 
     render(conn, "show.json", channel: channel |> Repo.preload([:projects, :user]))
   end
+
+  def pause(conn, %{"channel_id" => id}) do
+    channel_params = %{"paused" => true}
+    result = AskWeb.ChannelController.update(conn, %{"id" => id, "channel" => channel_params})
```
matiasgarciaisaia, line 112: Controller functions shouldn't call other controller functions.

This `update` endpoint may (now or later), for example, have different permissions checks than the ones needed for this `pause` endpoint - and coupling them won't help.

lib/ask_web/controllers/channel_controller.ex
```diff
@@ -106,4 +106,20 @@ defmodule AskWeb.ChannelController do
...
+    ChannelStatusServer.poll(ChannelStatusServer.server_ref())
+    ChannelStatusServer.wait(ChannelStatusServer.server_ref())
```
matiasgarciaisaia, line 114: The rationale for this is to force the status server to "learn the news" and stop the broker? I'm not sure we want to `wait` on that - do we need to block the users' interface for that?

It's probably enough if we just send the StatusServer a "recheck, please" message and let it handle it.

matiasgarciaisaia (commented):
I still have to finish reviewing the changes on the status server itself.

matiasgarciaisaia (approved):
👍 :shipit:

PR #2373, opened 2024-10-29
UI changes to pause a channel

ismaelbej:
* Rename AddButton to ActionButton
* Allow icon and color props for ActionButton
* Add ChannelPause button
* Add actions to pause a channel

Screenshot 2024-10-28 at 11 07 31 PM

Screenshot 2024-10-28 at 11 07 13 PM

Screenshot 2024-10-29 at 8 39 53 PM


assets/js/components/channels/DownChannelsStatus.jsx
```diff
@@ -12,7 +13,9 @@ class DownChannelsStatus extends PureComponent {
   render() {
     const { channels, timestamp } = this.props
     const channelNames = map(channels, (channel) => channel.name)
-    return this.downChannelsDescription(channelNames, timestamp)
+    const paused = every(channels, (channel) => channel.statusInfo && channel.statusInfo.status == "paused")
```
matiasgarciaisaia, line 16:
```suggestion
    const paused = every(channels, (channel) => channel.statusInfo?.status == "paused")
```
[Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

assets/js/components/channels/ChannelPause.jsx
```diff
@@ -0,0 +1,42 @@
...
+const ChannelPause = ({ channel, t, actions }) => {
+  if (!channel) return null
+  const { statusInfo } = channel
+  const paused = statusInfo != null && statusInfo.status == "paused"
+  const text = paused ? t("Unpause channel") : t("Pause channel")
+  const icon = paused ? "play_arrow" : "pause"
+  const color = paused ? "green" : "red"
+
+  const pauseChannel = (channel, pause) => {
+    console.log(`About to ${pause ? "pause" : "unpause"} channel ${channel.id}`)
```
matiasgarciaisaia, line 17:
```suggestion
```

matiasgarciaisaia (approved):
We'll probably have to rebase or something before merging, but the changes are OK 👍

ismaelbej:
The actions pause/unpause triggers a `fetch(channelId)` that seems to be enough to refresh the channel without waiting to the poll. I haven't found if there's an automatic mechanism to refresh UI of the channels. `ChannelBroker.on_channel_settings_change` is triggered but it seems to refresh the channel state on the backend.

matiasgarciaisaia (approved):
👍 :shipit: 

Since fetching the channel is enough for the UI to get updated (upon `CHANNEL_RECEIVE`), we could trigger a new `CHANNEL_FETCH` once per minute. I think the only part of the app that could have some feature like that is the sandbox, probably when using a real channel - maybe the simulation too?

matiasgarciaisaia:
Here's where the timer is set in the simulator's UI: https://github.com/instedd/surveda/blob/c82e6c01d53be72947484cd91e1dcc47cd0098cb/assets/js/components/surveys/SurveySimulation.jsx#L45-L49

PR #2374, opened 2024-11-13
Add active surveys to the warning message when deleting a channel

ismaelbej:
### Changes

* Add query to retrieve running surveys for a provider and baseUrl
* Change modal to show active surveys
* Add unit test for the endpoints

#### Searching for active surveys

Screenshot 2024-11-19 at 6 39 55 PM

#### No active surveys

Screenshot 2024-11-19 at 6 39 02 PM

#### List of active surveys

Screenshot 2024-11-21 at 7 48 45 PM


assets/js/components/channels/ChannelIndex.jsx
```diff
@@ -37,6 +57,31 @@ class ChannelIndex extends Component<any> {
...
+          modalIndex: index,
+        })
+      })
+      .catch(() => {
+        this.setState({
+          modalLoading: false,
+          modalSurveys: [],
+          modalProvider: provider,
+          modalIndex: index,
+        })
```
matiasgarciaisaia, line 83: This is silently dismissing an exception. Shouldn't we let it throw (so the user gets an automatic error message popup)? Or set `modalSurveys` to `null` ("error, beware") and tell it apart from an empty list ("no active channels, safe to delete provider").

assets/js/components/channels/ProviderModal.jsx
```diff
@@ -0,0 +1,60 @@
...
+      showCancel
+    >
+      <div>
+        <p>{t("Do you want to delete the channels provided by {{name}}?", { name })}</p>
+
+        <div className="provider-surveys">
+          {loading ? <span>{t("Searching active surveys...")}</span> :
+            surveys.length == 0 ? <span>{t("No active surveys")}</span> :
+              <div>
+                <span>{t("Active surveys")}</span>
```
matiasgarciaisaia, line 34:
```suggestion
                <span>{t("These surveys are active, using channels from this provider. Deleting the channels will interrupt the surveys.")}</span>
```

May be a better wording, but I'd try to explain a bit more why this is relevant to the user.

matiasgarciaisaia (changes requested):
Let's add access control to the endpoint.

assets/js/components/channels/ChannelIndex.jsx
matiasgarciaisaia: > I'm showing a generic error message an logging the error to the browser's console

Isn't that Surveda's default behaviour if we simply don't catch the exception? There's some standard component there that shows exceptions as a notification bubble, not sure if that applies here too.

ismaelbej: I've removed the catch handler and changes related.

lib/ask/survey.ex
```diff
@@ -530,10 +531,16 @@ defmodule Ask.Survey do
...
+  def with_active_channels(user_id, provider, base_url) do
     query =
       from s in Survey,
         where: s.state == :running,
+        join: pm in ProjectMembership,
+        on: pm.project_id == s.project_id and pm.user_id == ^user_id,
+        select: s
+
+    query =
+      from s in subquery(query),
```
matiasgarciaisaia, line 543:
```suggestion
```

Doesn't it work all together in a single query?

matiasgarciaisaia (approved):
It's usually a good idea to avoid DB subqueries if possible.

If the subquery is actually needed, let's merge as is!

PR #2379, opened 2025-02-26
Fix: Channel Broker crashes with nil nuntium_token

### LINKED ISSUE #2377: Survey progress halts when a contact in the Channel Broker queue has no Nuntium token
We've observed that the Channel Broker queue sometimes gets a contact queued with a `last_contacted` date set, but its `channel_state` is still `nil`.

We didn't find the cause for that yet (we suspect it may be due to a timeout from Nuntium when queueing the AOs), but when the error happens, the Channel Broker crashes due to a missing `Ask.Runtiume.NuntiumChannel.message_inactive?/2` definition with a `nil` second parameter.

This effectively stops the survey from progressing - the channel broker won't send contact attempts for that survey due to crashing.

We should:

- [ ] Make the channel broker move on and keep trying the next contact in the queue.
- [ ] Understand why are we hitting this condition
- [ ] Maybe define the function for the `nil` case

matiasgarciaisaia:
Fixes #2377

### DIFF
--- a/lib/nuntium/client.ex
+++ b/lib/nuntium/client.ex
@@ -13,6 +13,9 @@ defmodule Nuntium.Client do
     %Client{base_url: url, oauth2_client: oauth2_client}
   end
 
+  defp status_or_reason(%OAuth2.Error{reason: reason}), do: reason
+  defp status_or_reason(%{status_code: status_code}), do: status_code
+
   # @spec send_ao(t, String.t, list()) :: {:ok, %{nuntium_token: String.t}} | error
   def send_ao(client, account, messages) do
     url = "#{client.base_url}/api/ao_messages.json?#{URI.encode_query(account: account)}"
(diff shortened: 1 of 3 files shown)


mverzilli:
Trying to familiarize myself with the code. Let me parrot what I understand of the solution back at you to make sure I'm understanding:

Basically, you're making the Nuntium token returned by Guisso default to 0 to get rid of the nil causing this issue.

Now, doesn't that mean we're sending a request to Nuntium that we know is going to fail?

lib/nuntium/client.ex
```diff
@@ -13,6 +13,9 @@ defmodule Nuntium.Client do
     %Client{base_url: url, oauth2_client: oauth2_client}
   end
 
+  defp status_or_reason(%OAuth2.Error{reason: reason}), do: reason
+  defp status_or_reason(%{status_code: status_code}), do: status_code
+
```
anaPerezGhiglia, line 18: this is the same as in verboice channel. It is also the same logic that's underneath `parse_response` function (both in nuntium and verboice channel). I think parse_response could call this function for returning the error tuple instead of repeating the pattern again)

anaPerezGhiglia (approved):
LGTM 👍

matiasgarciaisaia:
CC: @mverzilli 

> Basically, you're making the Nuntium token returned by Guisso default to 0 to get rid of the nil causing this issue.

No, not really.

In `lib/nuntium/client.ex`, I'm avoiding it crashing upon timeouts.

In [`lib/ask/runtime/channel_broker.ex`](https://github.com/instedd/surveda/pull/2379/files#diff-001db06c3561f4d97e4598f22c759714c543038c8b5893df441ceb4f2bff75a1R481-R483), I'm leaving the `channel_state` unchanged upon errors (so, basically, it stays `nil` - not a 0).

And, finally, in `lib/ask/runtime/nuntium_channel.ex` I'm making it ignore (instead of crash) when the queued contact doesn't have a `nuntium_token` in that state (called from the [channel_broker_state](https://github.com/instedd/surveda/blob/7d0787bf870097707b0ef608512b86f27ef7116e/lib/ask/runtime/channel_broker_state.ex#L331)).

lib/nuntium/client.ex
matiasgarciaisaia: Yes, I took this code from the Verboice Channel.

I think `parse_response` doesn't use this because it also adds an `:ok`/`:error` "header" in the response tuple, and you need to tell them apart.

There probably is some way to improve the code, but it's not straightforward, I think.

PR #2380, opened 2025-03-05
Respect survey schedules before sending SMS

### LINKED ISSUE #2283: Respect Survey schedule when sending SMS messages
Nuntium doesn't have time schedules like Verboice has, so Surveda previously didn't have any mean to prevent messages from being sent after hours, that is after the scheduled time window ended for the day.

The channel-broker didn't fix this issue. It acts as a dam to avoid spamming and overflowing the Nuntium servers, but when the time window ends for the day, knowing nothing about the scheduled time for contacts, all messages queued during the time window will be sent, until the queue is emptied. This will happen at some point because the survey-broker noticed that the time window has ended and won't poll the survey anymore.

**Surveda now as a way to fix the issue**. `Ask.Runtime.Channel` can compute and pass the `not_before` and `not_after` attributes for the SMS modes, as it does for IVR mode, and pass them to `Ask.Runtime.ChannelBroker.ask`, allowing the channel-broker to respect the scheduled time: keep messages in queue until `not_before` and skip messages when `not_after` is reached.

Of course, this has no impact on currently active contacts (responses are sent immediately, regarless of the scheduled time, only on new contacts and retries).

mverzilli:
Fixes #2283 

The fix involves adding `not_before` and `not_after` fields to `ChannelBroker.ask` (aka, `sms`, btw why don't we rename this??), and honoring them (it was only sent with `ChannelBroker.setup` (aka, 'ivr', ditto).

The way to honor those fields involves:
- Making sure they get also saved in the queue (previously only for IVR, `nil` for SMS)
- Add expiration checks for SMS when `ChannelBroker` dequeues
- Mark contact attempts as expired accordingly (so it gets picked up the next day)

Pending work:

- [x] Make it work 
- [x] Make it nice (aka refactor)
- [x] Future proof (add some more test cases)
- [x] Update docs with new behaviors

Updated wiki: 

https://github.com/instedd/surveda/wiki/Chart:-Start-Survey-(with-Channel-Broker)/565d79ddc3ac0b23f322a6584eac6604b533793e

### DIFF
--- a/lib/ask/runtime/channel_broker.ex
+++ b/lib/ask/runtime/channel_broker.ex
@@ -412,40 +412,35 @@ defmodule Ask.Runtime.ChannelBroker do
   # Activates the next queued contact. There must be at least one contact in
   # queue. It doesn't verify if the channel capacity has been reached!
   defp activate_next_queued_contact(state) do
-    {new_state, unqueued_item} =
+    {new_broker_state, unqueued_item} =
       state
       |> refresh_runtime_channel()
       |> State.activate_next_in_queue()
 
-      case unqueued_item do
+    case unqueued_item do
       {respondent_id, token, not_before, not_after} ->
-        Respondent.with_lock(respondent_id, fn respondent ->
-          cond do
-            expired_contact_attempt?(not_after) ->
-              new_state = State.deactivate_contact(new_state, respondent.id)
-              Ask.Runtime.Survey.contact_attempt_expired(respondent)
-              new_state
-
-            true ->
-              ivr_call(new_state, respondent, token, not_before, not_after)
-          end
+        contact_or_mark_as_expired(respondent_id, not_after, new_broker_state, fn (respondent) -> 
+          ivr_call(new_broker_state, respondent, token, not_before, not_after)
         end)
       {respondent_id, token, reply, _not_before, not_after} ->
-        Respondent.with_lock(respondent_id, fn respondent ->
-          cond do
-            expired_contact_attempt?(not_after) ->
-              new_state = State.deactivate_contact(new_state, respondent.id)
-              Ask.Runtime.Survey.contact_attempt_expired(respondent)
-              new_state
-
-            # Note: we no longer need to carry not_before/not_after since
-            # ensuring schedules are respected is Surveda's responsibility
-            # for SMS mode (see expired_contact_attempt a couple of lines above)
-            true ->
-              channel_ask(new_state, respondent, token, reply)
-          end
+        contact_or_mark_as_expired(respondent_id, not_after, new_broker_state, fn (respondent) ->
+          channel_ask(new_broker_state, respondent, token, reply)
         end)
-    end
+      end
+  end
+
+  defp contact_or_mark_as_expired(respondent_id, not_after, new_broker_state, do_contact) do
+    Respondent.with_lock(respondent_id, fn respondent ->
+      cond do
+        expired_contact_attempt?(not_after) ->
+          new_state = State.deactivate_contact(new_broker_state, respondent.id)
+          Ask.Runtime.Survey.contact_attempt_expired(respondent)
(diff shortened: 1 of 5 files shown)


lib/ask/runtime/channel_broker.ex
```diff
@@ -36,8 +36,8 @@ defmodule Ask.Runtime.ChannelBroker do
     call(channel_id, {:has_delivery_confirmation?})
   end
 
-  def ask(channel_id, channel_type, respondent, token, reply) do
-    cast(channel_id, {:ask, channel_type, respondent, token, reply})
+  def ask(channel_id, channel_type, respondent, token, reply, not_before \\ nil, not_after \\ nil) do
```
mverzilli, line 39: Tempted to refactor so that nil's aren't accepted, but I wonder if there are other parts of Surveda where a schedule is really optional

test/ask/runtime/channel_broker_state_test.exs
```diff
@@ -69,9 +69,9 @@ defmodule Ask.Runtime.ChannelBrokerStateTest do
       mock_time(now)
 
       State.new(0, "sms", %{})
-      |> State.queue_contact({%{id: 2, disposition: :queued}, "secret", []}, 5)
+      |> State.queue_contact({%{id: 2, disposition: :queued}, "secret", [], nil, nil}, 5)
```
matiasgarciaisaia, line 72: Shall we set non-nil times here, so we don't add coverage for a case we don't want to support?

test/ask/runtime/channel_broker_test.exs
```diff
@@ -241,6 +241,50 @@ defmodule Ask.Runtime.ChannelBrokerTest do
...
+          |> activate_respondent("sms", Enum.at(respondents, 3), not_before, not_after)
+          |> activate_respondent("sms", Enum.at(respondents, 4), not_before, not_after)
+
+        expected_active_respondents = [
+          Enum.at(respondents, 0),
+          Enum.at(respondents, 3),
+          Enum.at(respondents, 4)
+        ]
+
+        # All respondents should be in broker queue, but only the first one contacted
```
matiasgarciaisaia, line 277: Where's Surveda guaranteeing this? We [activate respondents with a `not_before` in the following minute, too](https://github.com/instedd/surveda/blob/cb53f55d7bd973a5aa8f46c44a2b2a6e084df90f/lib/ask/runtime/channel_broker_state.ex#L149), so I think there's a good chance of contacting both respondents 3 and 4 here when the broker runs.

matiasgarciaisaia (approved):
LGTM! 👍

lib/ask/runtime/channel_broker.ex
mverzilli: Yeah, the simulator came to mind. I can do it then. Let's pray that test coverage is enough :D

lib/ask/runtime/channel_broker.ex
```diff
@@ -36,8 +36,8 @@ defmodule Ask.Runtime.ChannelBroker do
...
+    cast(channel_id, {:ask, channel_type, respondent, token, reply, not_before, not_after})
```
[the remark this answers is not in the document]
mverzilli: Would only work if your function has the node id as the first param (fair assumption, but not enforced beyond convention)

test/ask/runtime/channel_broker_test.exs
mverzilli: It's an outdated comment from a previous iteration on this test, it's actually not true even in the context of this test. Will fix

lib/ask/runtime/channel_broker.ex
mverzilli: Mmm, I just noticed `ask` is also used for mobile web mode, and in that one we don't care about schedules

Issue #2368, opened 2024-09-25
Retry attempts are done earlier than the defined delay


manumoreira:
## General Description

The delays defined in the survey settings for the re-contact attempts are not working as expected.
We were used to see delays on those attempts and that was caused by the inability of Surveda to handle contact attempts according to channel capacity. i.e. The delay was 26h, Surveda sent the attempt to the channel, but it was so busy that the attampt was done 4h later getting a 30h delay.
Since the Channel Broker (ChB) the delays are not happening that often, but we noticed another issue: Some attempts are done earlier than expected.

### Why is this happening?

The strongest hypothesis is that the Survey Broker (SB) is considering the Enqueuing call time (for IVR of course) to calculate the delay. 
When the SB was designed the difference between the Enqueueing of a contact attempt and the attempts itself was trivial (in theory), but after years of interacting with real channels and after having a ChB that handles priorities (prioritizing re-contacts over first contacts) the difference got bigger and evident.

#### A test case
In this survey we have defined a re-contact delay of 4h
![Screenshot 2024-09-25 at 13 46 24](https://github.com/user-attachments/assets/a1f87b79-853c-4174-9d17-b94364cb4d10)
The call was queued at 17:45, but it waited in the ChB queue until 19:30 when it finally went out.
At 21:45 4h after the call was queued it gets the new attempt

### What should we do?

We might want to change the behavior of the SB and start counting for the re-contact delay when the attempt is actually done. There are (at least) two options for this
1. when the ChB actually sends the call to the channel
2. when the channel sends the first callback related to that attempt

Probably the choice will be clear after checking to existing code.

## Steps to reproduce

1. Create a IVR Survey 
2. Add (at least) 1 re-contact with a delay > `n` hours and a batch-size < channel capacity by hour x `n`
3. Wait until 1 re-contact is completed for the first batch to download the interactions file and calculate the delay

Calculations were done using this notebook https://colab.research.google.com/drive/14pkYblNRNvevkBl42uvdmIIX4VZMhGrf?usp=sharing

matiasgarciaisaia:
This finding comes from analyzing the interactions files, and we have confirmed via Verboice that the calls were effectively done before they should - their `not_before` time was miscalculated.

[...]

PR #2382, opened 2025-03-14
Fix: update respondent timeout upon contact

matiasgarciaisaia:
Whenever the broker sends a contact to a channel provider, we update the respondent's timeout so the delay is from the actual contact attempt instead of when the SurveyBroker queued it in the ChannelBroker.

See #2368

### DIFF
--- a/lib/ask/runtime/survey_broker.ex
+++ b/lib/ask/runtime/survey_broker.ex
@@ -224,6 +224,8 @@ defmodule Ask.Runtime.SurveyBroker do
   end
 
   def recontact_queued_respondents(respondent_ids) do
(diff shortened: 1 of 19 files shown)


lib/ask/runtime/survey_broker.ex
```diff
@@ -224,6 +224,8 @@ defmodule Ask.Runtime.SurveyBroker do
   end
 
   def recontact_queued_respondents(respondent_ids) do
+    # TODO: it looks like this method is not used anymore since 46391a1430dbeb5b67498769080bc5a32f9cd135
```
mverzilli, line 227: why not remove then?

mverzilli (approved):

PR #2391, opened 2025-08-13
Import unused sample

### LINKED ISSUE #2389: Import unused sample from a finished survey
In #2386 we discussed offering the ability to download the unused sample of a survey, so it can be reused in a new survey. We have done that some times in the past due to technical issues with the server or the channels that spoiled the survey results.

But, after some discussion, we found out this could potentially expose some info about the sample itself (for example, deanonimize which were the only two respondents that replied in a survey by downloading the unused sample, and diff'ing it with the original one). It would also open up the possibility of a collaborator that doesn't typically handle sample to get access to it.

Instead of that, the solution we've come up with is to provide an option in the survey creation UI to add the unused sample of another survey to the current one. This could work both for new or ongoing surveys.

Superseeds #2386

matiasgarciaisaia:
Let users import into a survey all the uncontacted respondents from another finished one.

The source survey must be already terminated, and has to belong to the same project as the target one.

[surveda-import-sample.webm](https://github.com/user-attachments/assets/8401149a-e56b-40ac-8693-4ecdfc3dd8ab)

Error handling:
image

Fixes #2389

### DIFF
--- a/lib/ask_web/controllers/survey_controller.ex
+++ b/lib/ask_web/controllers/survey_controller.ex
@@ -68,6 +68,8 @@ defmodule AskWeb.SurveyController do
           left_join: r in Respondent,
           on: r.survey_id == s.id,
           where: s.project_id == ^project.id and s.state == :terminated,
+          # we could mix a `count` with a `where` clause filtering for respondent disposition
+          # instead of doing the sum+if, but that wouldn't return surveys with 0 respondents available
           select: %{survey_id: s.id, name: s.name, ended_at: s.ended_at, respondents: sum(fragment("if(?, ?, ?)", r.disposition == :registered, 1, 0))},
           group_by: [s.id]
       ) |> Enum.map(fn s -> %{


lib/ask_web/controllers/survey_controller.ex
```diff
@@ -58,6 +59,28 @@ defmodule AskWeb.SurveyController do
...
+  def list_unused(conn, %{"project_id" => project_id}) do
+    project = load_project(conn, project_id)
+
+    surveys =
+      Repo.all(
+        from s in Survey,
+          left_join: r in Respondent,
+          on: r.survey_id == s.id,
+          where: s.project_id == ^project.id and s.state == :terminated,
+          select: %{survey_id: s.id, name: s.name, ended_at: s.ended_at, respondents: sum(fragment("if(?, ?, ?)", r.disposition == :registered, 1, 0))},
```
ggiraldez, line 71: I suppose this could also be a `select ..., count(*) as respondents from ... where r.disposition = 'registered'` which may be more efficient. I'm not sure about the Ecto syntax though.

ggiraldez (approved):
Looks good as far as I can evaluate with minimal Elixir and completely rusted redux/react knowledge. Left a small suggestion on some query, but it's probably not too important.

lib/ask_web/controllers/survey_controller.ex
matiasgarciaisaia: The Ecto syntax is not a problem:

```
        from s in Survey,
          left_join: r in Respondent,
          on: r.survey_id == s.id,
          where: s.project_id == ^project.id and s.state == :terminated and r.disposition == :registered,
          select: %{survey_id: s.id, name: s.name, ended_at: s.ended_at, respondents: count(s.id)},
          group_by: [s.id],
          order_by: [desc: count(s.id)]
```

But then you don't get the surveys that have 0 available respondents listed. I'd rather explicitely show that there are no available respondents rather than ignore the survey at all.

I could, though, add a comment on the code explaining this.

There's probably another query that can be written for that, but not sure it'd be better.

ggiraldez: Good point. You can use a `left join` for that, and `count(r.id)` or any other field that is `NULL` when joining an empty set.

lib/ask_web/controllers/respondent_group_controller.ex
```diff
@@ -47,6 +47,44 @@ defmodule AskWeb.RespondentGroupController do
...
+    if !Survey.terminated?(source_survey) do
+      render_invalid_import(conn, source_survey.id, "NOT_TERMINATED", %{survey_state: source_survey.state})
+    else
+      entries = unused_respondents_from_survey(source_survey)
+
+      if !Enum.empty?(entries) do
+        sample_name = "__imported_from_survey_#{source_survey.id}.csv"
+        case RespondentGroupAction.load_entries(entries, survey) do
+          {:ok, loaded_entries} ->
+            survey |> RespondentGroupAction.disable_incentives_if_disabled_in_source!(source_survey)
```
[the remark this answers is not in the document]
ggiraldez: That's fair. I guess what threw me off was mostly that the updating function is in the respondent group file, but it updates the survey. I see that the module is `RespondentGroupAction`, but I still find it a bit confusing. Also, how are incentives related to anonymization? Is that legacy naming?

PR #2393, opened 2025-09-04
Simplify unused sample DB query

matiasgarciaisaia:
After what we discussed with @ggiraldez in #2391


ggiraldez (approved):
Looks great! :shipit:

PR #2396, opened 2025-09-11
Fix: dispositions table's last column color

matiasgarciaisaia:
When we moved the expand arrows to the left back in c4d662805bb9656420bf3af9b32a81a432f03cc4 we broke the CSS rules for coloring the table, making the last disposition's column be grey instead of its color, and leaving the percent column in black.

See #1266

### DIFF
--- a/assets/css/_card-table.scss
+++ b/assets/css/_card-table.scss
@@ -39,8 +39,8 @@
             }
           }
           &:first-child,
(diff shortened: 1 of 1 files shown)


matiasgarciaisaia:
Before:

image


Now:

image

manumoreira:
color rejected

Confirmed!

PR #2399, opened 2026-01-16
Fix #2397. Add vertical scroll on ImportSampleModal

### LINKED ISSUE #2397: Import unused respondents modal should be scrollable
If the list of surveys with sample to import is large the user cannot see it completely because the modal doesn't scroll.

Image

manumoreira:
[surveda-import-unused-sample-scroll.webm](https://github.com/user-attachments/assets/bd1fe359-b58f-4dd5-b02a-7a47204e302e)


matiasgarciaisaia (approved):
👍

PR #2400, opened 2026-03-05
Update instructions

manumoreira:


matiasgarciaisaia (changes requested):
We don't need to specify the platform for dockerdev - it's only that waj's image doesn't provide ARM builds.

Re: ngrok. We don't usually need that for development, and free ngrok URLs are not stable, so I'd leave the docs as they are for now.

I've sent #2401 to replace this one.

